PR #3, opened 2015-02-02
monad: remove S.bind and S.then in favour of R.chain compatibility

davidchambers:
Is this a step in the right direction do you think, @buzzdecafe?

### DIFF
--- a/index.js
+++ b/index.js
@@ -55,6 +55,11 @@
   }
   extend(Nothing, Maybe);
 
+  //  Nothing#chain :: (a -> m b) -> m b
+  Nothing.prototype.chain = function(f) {  // jshint ignore:line
+    return this;
+  };
+
   //  Nothing#equals :: Maybe a -> Boolean
   Nothing.prototype.equals = function(maybe) {
     return maybe instanceof Nothing;
@@ -78,6 +83,11 @@
     }
   }
   extend(Just, Maybe);
+
+  //  Just#chain :: (a -> m b) -> m b
+  Just.prototype.chain = function(f) {
+    return f(this.value);
+  };
 
   //  Just#equals :: Maybe a -> Boolean
   Just.prototype.equals = function(maybe) {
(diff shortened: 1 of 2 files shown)


buzzdecafe:
do you have any interest in fabtasy-land compat? then you would rename `bind` to chain and not worry about native `bind`

index.js
```diff
@@ -55,6 +56,11 @@
   }
   extend(Nothing, Maybe);
 
+  //  Nothing#bind :: (a -> m b) -> m b
+  Nothing.prototype.bind = function(f) {  // jshint ignore:line
+    return this;
+  };
```
buzzdecafe: this is the same as `map`, so you could use a reference to the same function if you were so inclined. not a big deal.

index.js
```diff
@@ -79,6 +85,11 @@
   }
   extend(Just, Maybe);
 
+  //  Just#bind :: (a -> m b) -> m b
+  Just.prototype.bind = function(f) {
+    return f(this.value);
+  };
+
```
buzzdecafe: this is the same as map, so you could use a reference to the same function if you were so inclined. not a big deal.

davidchambers:
> do you have any interest in fantasy-land compat? then you would rename `bind` to chain and not worry about native `bind`

Ah, so *that's* what fantasy-land's `chain` does! Great idea, @buzzdecafe. The only question is whether the exported function should be named `S.bind` or `S.chain`. I prefer the former: Haskell's `>>=` is pronounced "bind", and "chain" only makes sense when used as a method (e.g. `m.chain(f).chain(g)`). It would mean a function named "bind" delegates to a method named "chain", but I *think* I'm okay with that.

index.js
davidchambers: I don't mind a bit of repetition if it saves readers from scrolling.

benperez:
:dromedary_camel:

buzzdecafe:
for the record, `ramda.chain` also dispatches to `chain` "methods"

davidchambers:
> for the record, `ramda.chain` also dispatches to `chain` "methods"

In that case `S.bind` is an :alien:! Perhaps the best course of action is to remove `S.bind` and `S.then` and use `R.chain` instead. :)

buzzdecafe:
:+1:

PR #4, opened 2015-02-02
parse: add S.parseFloat and S.parseInt

davidchambers:
`Maybe Number` better expresses the return type than `Number` (with NaN). See purescript/purescript-globals#5 for related discussion.


buzzdecafe:
this looks like a great idea to me

PR #6, opened 2015-02-07
maybe: add S.encase

davidchambers:
> __encase__
> _verb_
> enclose or cover in a case or close-fitting surround

As I imagine it we're _encasing_ a function which may explode in an indestructible wrapper. Alternative name suggestions welcome.

### DIFF
--- a/index.js
+++ b/index.js
@@ -121,6 +121,17 @@
     return x == null ? Nothing() : Just(x);
   };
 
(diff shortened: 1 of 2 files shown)


buzzdecafe:
...and the puns you can make now... "Just encase"

davidchambers:
Very punny, @buzzdecafe.

PR #8, opened 2015-02-09
maybe: implement Apply, Applicative, Chain, and Monad

davidchambers:
Reading fantasyland/fantasy-land#85 improved my understanding of these types, and motivated me to add the methods necessary for Maybe to implement the Apply, Applicative, Chain, and Monad specifications.

### DIFF
--- a/index.js
+++ b/index.js
@@ -46,6 +46,13 @@
     throw new Error('Cannot instantiate Maybe');
   }
 
+  //  Maybe.of :: a -> m a
+  Maybe.of = function(x) {
+    return new Just(x);
+  };
+
+  Maybe.prototype.of = Maybe.of;
+
   Maybe.prototype.type = Maybe;
 
   function Nothing() {
@@ -54,6 +61,11 @@
     }
   }
   extend(Nothing, Maybe);
+
+  //  Nothing#ap :: m a -> m b
+  Nothing.prototype.ap = function(x) {  // jshint ignore:line
+    return this;
+  };
 
   //  Nothing#chain :: (a -> m b) -> m b
   Nothing.prototype.chain = function(f) {  // jshint ignore:line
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -46,6 +46,13 @@
     throw new Error('Cannot instantiate Maybe');
   }
 
+  //  Maybe.of :: a -> m a
+  Maybe.of = function(x) {
+    return new Just(x);
```
SimonRichardson: :+1:

index.js
```diff
@@ -55,6 +62,11 @@
   }
   extend(Nothing, Maybe);
```
buzzdecafe: maybe this is just me, but shouldn't `Nothing` be a singleton? maybe it doesn't matter, as long as they compare

davidchambers: I'd like to have just a single value of the Nothing type, but I haven't worked out a solution which meets all of the following requirements:

  - `new S.Nothing()` must return the value;
  - `S.Nothing()` must return the value;
  - applying `R.is(S.Nothing)` to the value must evaluate true; and
  - the `name` property of the value's constructor must be `'Nothing'`.

In fact, I don't think it's possible to meet all these requirements.

SimonRichardson: Don't know if this helps, but we solved some of those issues above for fantasy-land implementions (option, either and etc) with this... https://github.com/puffnfresh/daggy/blob/master/src/daggy.js#L103

davidchambers: Interesting! I don't fully understand that code upon first reading. I'll revisit it this weekend to work out exactly what it does. :)

davidchambers:
I haven't yet merged this as I'm hoping that fantasyland/fantasy-land#90 will be accepted, at which point I'll update the identifiers in the tests to match the updated specification.

PR #9, opened 2015-02-12
list: add S.find

davidchambers:
Similar to [`R.find`][1] but uses Maybe to encode success/failure, rather than relying on null/undefined.

[1]: http://ramdajs.com/docs/#find


buzzdecafe:
:+1:

PR #10, opened 2015-02-13
maybe: define "filter"

davidchambers:
This makes Maybe compatible with `R.filter` and, if ramda/ramda#799 is merged, with `R.reject`.

@joneshf, would you mind reviewing this to make sure I faithfully translated the first of the laws you gave me?

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -9,6 +9,12 @@ var S = require('..');
 
 var eq = assert.strictEqual;
 
+//  T :: a -> Boolean
+var T = function(x) { return true; };  // jshint ignore:line
+
+//  F :: a -> Boolean
+var F = function(x) { return false; };  // jshint ignore:line
+
 //  identity :: a -> a
 var identity = function(x) { return x; };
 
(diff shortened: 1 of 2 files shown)


test/index.js
```diff
@@ -61,6 +64,24 @@ describe('maybe', function() {
...
+      eq(nothing.filter.length, 1);
+      eq(nothing.filter(T), nothing);
+      eq(nothing.filter(F), nothing);
+
+      var m = S.Nothing();
+      var f = function(n) { return n * n; };
+      var p = function(n) { return n < 0; };
+      var q = function(n) { return n > 0; };
+
+      // FIXME: What is the name of this law?
```
joneshf: I think this is actually a free theorem by parametricity.

joneshf:
Looks like a good implementation. 

I'm not sure what your overall idea is here, but I'd like to think that you could implement `filter` once and for all if you know something is both a `Monad` and a `Monoid`:

```javascript
var filter = function(p, m) {
  return m.chain(function(x) {
    return p(x) ? m.of(x) : m.empty();
  });
}
```

Then you could put this on the prototype of whatever.

```javascript
...
Maybe.prototype.filter = function(p) {
  return filter(p, this);
}
...
```

So you don't end up having to write something similar for other types that are `Monad` and `Monoid`. If you're not planning on using the fantasyland stuff then forget about it.

davidchambers:
> I'd like to think that you could implement `filter` once and for all if you know something is both a `Monad` and a `Monoid`

Great idea! Once I merge #8 Maybe will implement the Monad specification. I'll then open a pull request to have it implement the Monoid specification. Once I merge *that* pull request I'll update this one to use your `filter` function. :)

davidchambers:
I've updated this pull request. Having Nothing and Just share their definition of `filter` is wonderful!

PR #12, opened 2015-03-10
control: add S.match

davidchambers:
This is a slight variation on the function proposed in ramda/ramda#797. I realized that a special `_` value is unnecessary unless one wishes to accept `null` and `undefined`, since `Object` matches every other value. I also decided to support types which do not implement `expose`, though I'm not convinced this is a good idea.

I'm very keen to hear what others think of this idea in general and the approach I've taken specifically.

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,6 +20,12 @@
     return toString_.call(x) === '[object Array]';
   };
 
+  var always = function(x) {
+    return function() {
+      return x;
+    };
+  };
+
   var curry = function(f) {
     var arity = f.length;
     var recurry = function(args) {
@@ -44,6 +50,10 @@
     return m.chain(function(x) {
       return pred(x) ? m.of(x) : m.empty();
     });
+  };
+
+  var identity = function(x) {
+    return x;
   };
 
   //  maybe  /////////////////////////////////////////////////////////////////
@@ -100,6 +110,11 @@
     return maybe instanceof Nothing;
   };
 
+  //  Nothing#expose :: (-> b) -> b
+  Nothing.prototype.expose = function(f) {
+    return f();
+  };
+
   //  Nothing#or :: Maybe a -> Maybe a
   Nothing.prototype.or = function(maybe) {
     return maybe;
@@ -139,6 +154,11 @@
     return maybe instanceof Just && maybe.value === this.value;
   };
 
+  //  Just#expose :: (a -> b) -> b
+  Just.prototype.expose = function(f) {
+    return f(this.value);
+  };
+
   //  Just#or :: Maybe a -> Maybe a
   Just.prototype.or = function(maybe) {  // jshint ignore:line
     return this;
@@ -151,14 +171,7 @@
 
   //  fromMaybe :: a -> Maybe a -> a
   var fromMaybe = curry(function(x, maybe) {
-    switch (true) {
-      case maybe instanceof Nothing:
-        return x;
-      case maybe instanceof Just:
-        return maybe.value;
-      default:
-        throw new TypeError('Pattern match failure');
-    }
+    return match([[Nothing, always(x)], [Just, identity]], maybe);
   });
 
   //  toMaybe :: a? -> Maybe a
@@ -182,6 +195,11 @@
   function Either() {
     throw new Error('Cannot instantiate Either');
   }
+
+  //  Either#expose :: (a -> b) -> b
+  Either.prototype.expose = function(f) {
+    return f(this.value);
+  };
 
   Either.prototype.type = Either;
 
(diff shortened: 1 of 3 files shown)


buzzdecafe:
i am thinking this over, and hope to respond in greater detail soon. My initial response is either that this goes *too far*, or that it does not go *far enough*. I'm leaning toward the latter. This seems to be tipping toward logic programming -- so go for it.

CrossEye:
I'm still not enamored of the notion of matching only on types.  I really prefer a more general pattern-matching mechanism.  I'm curious to see if this turns out to be very useful, and I understand your desire to let your types unwrap themselves in order to run these functions.  But it seems to me to be missing some of the basic features that I would want in pattern matching.  In a credit-card validation routine in Haskell, I need to double every-other number in a list.  The code is simply:

```haskell
doubleOddIndices:: [Integer] -> [Integer]
doubleOddIndices [] = []
doubleOddIndices (x:[]) = [x]
doubleOddIndices (x:(y:z)) = x : ((2 * y) : doubleOddIndices(z))
```

In each case, the pattern is matching on the same type; it's reaching deeper in to examine a bit about the structure of the data.  That strikes me as extremely useful.  The only way I see to accomplish that is to run an arbitrary function on the input, not a type-matcher.  But perhaps I already have what I need in `R.cond`.

So I guess I simply don't know the use-cases of this.  Are you viewing this mostly as a more sophisticated version of per type dynamic dispatch, or do you want it to do more?

buzzdecafe:
> I'm still not enamored of the notion of matching only on types. I really prefer a more general pattern-matching mechanism.

but as long as you are restricting yourself to that, then get some bang for your buck, and let the matching/satisfaction run in both directions. But maybe that is just my little hobby horse

davidchambers:
> I understand your desire to let your types unwrap themselves in order to run these functions.

That's the primary motivation. It's important that we're able to operate on values without accessing their internal properties. `S.fromMaybe` is one of several functions which facilitates this for maybes. For example:

```javascript
//  numberFromMaybe :: Maybe Number -> Number
var numberFromMaybe = S.fromMaybe(NaN);
```

The above is just sugar for the following:

```javascript
//  numberFromMaybe :: Maybe Number -> Number
var numberFromMaybe = S.match([
  [S.Nothing, R.always(NaN)],
  [S.Just,    R.identity],
]);
```

I'm not sure how to express the following without `S.match`:

```javascript
//  enlighten :: Either String Number -> String
var enlighten = S.match([
  [S.Left,  R.concat('Failed to compute the meaning of life: ')],
  [S.Right, R.concat('The meaning of life is ')],
]);
```

One *could* use `R.cond`, but it's then necessary to access the internal `value` property:

```javascript
R.cond(
  [R.is(S.Left),  R.pipe(R.prop('value'), R.concat('Failed to compute the meaning of life: '))],
  [R.is(S.Right), R.pipe(R.prop('value'), R.concat('The meaning of life is '))]
)
```

> But it seems to me to be missing some of the basic features that I would want in pattern matching.

I agree that this comes nowhere close to Haskell's pattern matching in terms of usefulness. I love the credit card number validation example, but I don't see how `S.match` could be extended to support unpacking of arrays, objects, and array-like objects.

> but as long as you are restricting yourself to that, then get some bang for your buck, and let the matching/satisfaction run in both directions.

What do you mean by "both directions"?

buzzdecafe:
i mean don't limit the constraint satisfaction to just the left side. But this is probably not what you are after at all --

buzzdecafe:
> One could use `R.cond`, but it's then necessary to access the internal value property:

you can do this according to spec with an `extract` or `getOrElse` function, if you are uncomfortable about directly accessing the `value` property (as you should be -- that's the whole point, after all).

CrossEye:
I don't have any objections to this function.  I'm just not sure how well I see the usefulness.  SICP talks about a somewhat different form of dynamic dispatch for types where you can register additional types as needed, even if they didn't exist when the function was defined.  I would love that sort of flexibility over `match`.  But since I don't see any straightforward ways to do that, I don't have any alternative suggestions.

I think maybe may biggest issue is that you're claiming the name `match` for something I think is a bigger idea.  But that's really a nit, and not worth bikeshedding until you prove it out.

-----

The credit-card number validation can be done with existing `cond` reasonably well, but it's just so much less concise than the Haskell version.  The main thing though is that it's not the condition testing that's the ugly part.  That part of it isn't too bad for JS.  It's the remainder that's ugly:

```javascript
var doubleOddIndices = R.cond(
  [R.isEmpty, R.always([])],
  [R.propEq('length', 1), R.identity],
  [R.T, ls => [ls[0], 2 * ls[1]].concat(doubleOddIndices(ls.slice(2)))]
);

doubleOddIndices([1, 2, 3, 4, 5, 6]); //=> [1, 4, 3, 8, 5, 12]
```

davidchambers:
> I'm not sure how to express the following without `S.match`:
>
> ```javascript
> //  enlighten :: Either String Number -> String
> var enlighten = S.match([
>   [S.Left,  R.concat('Failed to compute the meaning of life: ')],
>   [S.Right, R.concat('The meaning of life is ')],
> ]);
> ```

I forgot about [`S.either`][1]! `enlighten` can be defined as:

```javascript
//  enlighten :: Either String Number -> String
var enlighten = S.either(R.concat('Failed to compute the meaning of life: '),
                         R.concat('The meaning of life is '));
```

If Maybe has `S.fromMaybe` and Either has `S.either`, perhaps each type should define its own function(s) for applying transformations to internal values. I won't close this pull request just yet, but I'm now leaning in that direction. The one thing I will do is update this pull request to define `S.fromMaybe` and `S.either` in terms of `S.match`.


[1]: https://github.com/plaid/sanctuary/blob/v0.4.0/index.js#L196-L206

index.js
```diff
@@ -151,14 +171,7 @@
...
-    switch (true) {
-      case maybe instanceof Nothing:
-        return x;
-      case maybe instanceof Just:
-        return maybe.value;
-      default:
-        throw new TypeError('Pattern match failure');
-    }
+    return match([[Nothing, always(x)], [Just, identity]], maybe);
   });
```
davidchambers: `fromMaybe` defined in terms of `match`.

index.js
```diff
@@ -223,18 +241,25 @@
...
-    switch (true) {
-      case either instanceof Left:
-        return l(either.value);
-      case either instanceof Right:
-        return r(either.value);
-      default:
-        throw new TypeError('Pattern match failure');
-    }
+    return match([[Left, l], [Right, r]], either);
   });
```
davidchambers: `either` defined in terms of `match`.

PR #13, opened 2015-04-08
add gets function and tests

benperez:
`S.gets` allows one to do safe, nested property lookups. Nested property lookups are so common (and error prone) in JS that I think this would be a handy method to have around.
Compare the current method
```javascript
const obj = {foo: {bar: {baz: 'qux'}}};
const maybe_qux =  R.pipe(S.get('foo'),
  R.chain(S.get('bar')),
  R.chain(S.get('baz')))(obj)
```
with `S.gets`
```javascript
const maybe_qux = S.gets(['foo', 'bar', 'baz'], obj);
```
As far as implementation details, the non-reduce option is
```javascript
var gets = curry(function(keys, obj) {
  var acc = Just(obj)
  for (var idx = 0, len = keys.length; idx < len; idx += 1) {
    acc = acc.chain(get(keys[idx]));
  }
  return acc;
});
```
And `S.gets([], obj)` will return `S.Just(obj)`.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -756,6 +756,18 @@ describe('object', function() {
 
   });
 
+  describe('gets', function() {
+
+    it('returns a Maybe', function() {
+      var obj = {x: {z: 0}, y: 42};
+      assert.deepEqual(S.fromMaybe({}, S.gets([], obj)), obj);
+      assert(S.gets(['y'], obj).equals(S.Just(42)));
+      assert(S.gets(['z'], obj).equals(S.Nothing()));
+      assert(S.gets(['x', 'z'], obj).equals(S.Just(0)));
+    });
+
+  });
+
 });
 
 describe('parse', function() {
--- a/index.js
+++ b/index.js
@@ -290,6 +290,15 @@
     return hasOwnProperty_.call(obj, key) ? Just(obj[key]) : Nothing();
   });
 
+  //  gets :: [String] -> Object -> Maybe *
+  var gets = curry(function(keys, obj) {
+    var acc = Just(obj);
+    for (var idx = 0, len = keys.length; idx < len; idx += 1) {
+      acc = acc.chain(get(keys[idx]));
+    }
+    return acc;
+  });
+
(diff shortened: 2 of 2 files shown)


test/index.js
```diff
@@ -756,6 +756,17 @@ describe('object', function() {
...
   });
 
+  describe('gets', function() {
+
+    it('returns a Maybe', function() {
+      var obj = {x: {z: 0}, y: 42};
+      assert(S.gets(['x', 'z'], obj).equals(S.Just(0)));
+      assert(S.gets(['y'], obj).equals(S.Just(42)));
+      assert(S.gets(['z'], obj).equals(S.Nothing()));
+    });
```
davidchambers: You should include a test for `[]`. Also, I like to put the simpler tests before the more complex ones, so I'd move the `['x', 'z']` test after the `['y']` and `['z']` ones.

davidchambers:
I like this function. I find myself chaining gets pretty frequently. I'm coming around to the name, too. I'm interested in what others think of this function and its name.

buzzdecafe:
why "non-reduce version"? isn't the higher abstraction here that these things _fold_?

benperez:
> why "non-reduce option"? isn't the higher abstraction here that these things fold?

Yes, that's originally how I wrote it. I've never written much client-side JS so definitely correct me if I'm wrong but apparently `Array.prototype.reduce` isn't supported everywhere.

buzzdecafe:
> apparently Array.prototype.reduce isn't supported everywhere.

that's right, but that's beside the point I'm asking about. I don't know what the laws around `reduce` are, but implementing `reduce` on the type itself seems like an attractive direction to explore.

buzzdecafe:
i suppose it doesn't matter here since the first arg is guaranteed to be an array

benperez:
> implementing reduce on the type itself

What exactly would this look like?

buzzdecafe:
https://github.com/fantasyland/fantasy-land/pull/68

davidchambers:
I like it. Let's include it in the next release and see how useful it is in practice.

PR #15, opened 2015-04-15
report lengths of exported functions correctly

davidchambers:
The internal `arity` function uses the `Function` constructor to avoid the [wall of text][1] used to define `R.arity`.

[1]: https://github.com/ramda/ramda/blob/v0.13.0/src/arity.js#L32-L43

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,15 +20,29 @@
     return toString_.call(x) === '[object Array]';
   };
 
+  var arity = function(n, f) {
+    var params = Array(n + 1).join('x').replace(/x/g, function(_, n) {
+      return n === 0 ? '$0' : ', $' + n;
+    });
+    return Function(  // jshint ignore:line
+      'f',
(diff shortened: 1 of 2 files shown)


buzzdecafe:
:laughing: 

yes, it's a trade-off. not having `eval` is nice

benperez:
:dromedary_camel:

buzzdecafe:
for the record, if you ever deploy `eval`-ing code somewhere that enforces a CSP, it will be disallowed. since `arity` is fundamental, you should at least be aware of that potential pitfall.

see http://www.html5rocks.com/en/tutorials/security/content-security-policy/

davidchambers:
Thanks for the link, @buzzdecafe. Long term there's a reasonable chance we'll add Ramda as a dependency, at which point internal functions such as `arity` and `curry` will be made redundant.

PR #17, opened 2015-04-21
combinator: add S.K

davidchambers:
This is the (`R.always`) implementation suggested by @aaditmshah in https://github.com/ramda/ramda/issues/732#issuecomment-70383334. It may be too strict for Ramda, but it feels right at home here.

### DIFF
--- a/index.js
+++ b/index.js
@@ -60,6 +60,13 @@
     });
   };
 
+  //  combinator  ////////////////////////////////////////////////////////////
+
+  //  K :: a -> b -> a
+  var K = curry(function(x, y) {  // jshint ignore:line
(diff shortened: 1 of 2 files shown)


aaditmshah:
To be honest, when I read the title of this pull request I thought that you were implementing the [SKI combinator calculus](https://en.wikipedia.org/wiki/SKI_combinator_calculus).

buzzdecafe:
:+1:

CrossEye:
@aaditmshah: Ditto!

davidchambers:
That's currently over my head, @aaditmshah. :)

PR #19, opened 2015-04-21
Add Documentation for List and Object Functions

svozza:
I kept having to refer to the tests to remember the names of the list and object functions I wanted to use so I decided to add descriptions of them to the docs. I didn't add `S.gets` yet because it's not in the version on NPM.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,5 +46,104 @@ This approach is even cleaner if one uses [Ramda][1]:
 R.map(R.toUpper, S.head(words))
 ```
 
+As mentioned, Sanctuary provides several `null`-safe functions for interacting with objects and lists, they are as follows:
+
+##### Objects
+`get :: String -> Object -> Maybe *`
+
+Returns a `Just` containing the value found at the specified key or a `Nothing` if the value is not found.
+```javascript
+S.get('a', {a: 1, b: 2}) // => Just(1)
+S.get('c', {a: 1, b: 2}) // => Nothing()
+```
+
+##### Lists
+`at :: Number -> [a] -> Maybe a`
+
+Returns a `Just` containing the item at the index specified or a `Nothing` if the index is not found.
+```javascript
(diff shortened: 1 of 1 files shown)


davidchambers:
Thank you, @svozza! This is wonderful. Good timing, too. Over the past day or two I've been considering how to go about documenting this project. My thoughts follow.

  - Using the README should suffice. During this project's infancy I'd like to avoid the complexities of generating and styling HTML documents and maintaining a `gh-pages` branch.

  - Type signatures are the most important aspect of the documentation.

  - Ideally the Markdown could be derived from code comments. For example, this source code:

        //  head :: [a] -> Maybe a
        //
        //  Returns a `Just` containing the first item in the list or a `Nothing`
        //  if the list is empty.
        //
        //  ```javascript
        //  S.head([0, 1, 2, 3, 4]);  // => Just(0)
        //  S.head([]);  // => Nothing()
        //  ```
        var head = at(0);

    would produce this Markdown:

        ### `head :: [a] -> Maybe a`

        Returns a `Just` containing the first item in the list or a `Nothing`
        if the list is empty.

        ```javascript
        S.head([0, 1, 2, 3, 4]);  // => Just(0)
        S.head([]);  // => Nothing()
        ```

    Bonus points if the output were to include a link to the source code.

  - The README should document the project as it exists on the `master` branch. Once all the functions are documented, pull requests which add or alter functionality should include additions or alterations to the inline documentation, and to the README (by running `make`).

  - Documentation for published versions will be available at `https://github.com/plaid/sanctuary/blob/vX.Y.Z/README.md`.

The great news is that this pull request is compatible with the approach I've been considering. Unless you have any objection, @svozza, I'll merge this pull request as it stands and move the function descriptions to the source code once I've written a documentation generator. :)

svozza:
Sounds great. We generate nearly all our docs in Highland from code comments too,  it's a handy approach.

davidchambers:
:deciduous_tree:

Issue #20, opened 2015-04-22
S.pluck?


svozza:
This is really sheer laziness on my part but my code is littered with `R.map(S.get('key'), arr))`, it's an improvement on having to remember to do `R.reject(R.isNil, R.pluck('key', arr))` whenever I use `R.pluck` but it would be nice to have some sugar around such a common pattern.

davidchambers:
`pluck` may be sugar, but sugar can be worthwhile. `R.pluck(x)` is sugar for `R.map(R.prop(x))`; `S.pluck(x)` could similarly be sugar for `R.map(S.get(x))`. The hitch is that Sanctuary doesn't currently depend on Ramda, though it's becoming clear that it should.

I'll put together a pull request to add Ramda as a dependency. Once that has been reviewed and merged, perhaps you'd like to open a pull request to add `S.pluck`. :)

On a different note, it's worth keeping in mind this distinction:

```javascript
> R.prop('a', {})
undefined
> R.prop('a', {a: undefined})
undefined
> S.get('a', {})
Nothing()
> S.get('a', {a: undefined})
Just(undefined)
```

svozza:
Well I was planning on opening a PR that created the functionality from scratch,  or at least with no dependencies, but if you want to use Ramda I can  easily do that instead.

davidchambers:
Ramda is now a dependency, so open a pull request to add `S.pluck` if you feel so inclined, @svozza. :)

svozza:
Excellent. I'll definitely have time over the weekend.

PR #21, opened 2015-04-23
add Ramda dependency

davidchambers:
Recent discussion in #15 and #20 has made it apparent that Sanctuary should depend on Ramda, in addition to working nicely alongside it. We've so far duplicated two significant Ramda functions solely for internal use: `curry` and `arity`. As Sanctuary grows this list could easily expand to include `map`, [`toString`][1], and other functions I'd very much like to avoid duplicating.

Furthermore, every Ramda function supports the [`R.__`][2] placeholder thanks to [`R.curry`][3]. Sanctuary functions lack this useful characteristic. Sanctuary could, of course, provide its own placeholder value, but since Sanctuary is a companion to rather than a replacement for Ramda, reusing Ramda's placeholder value is a better option.

[1]: https://github.com/ramda/ramda/pull/924
[2]: http://ramdajs.com/docs/#__
[3]: http://ramdajs.com/docs/#curry

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,38 +12,17 @@
 
   'use strict';
 
-  var hasOwnProperty_ = Object.prototype.hasOwnProperty;
-  var slice_ = Array.prototype.slice;
-  var toString_ = Object.prototype.toString;
+  var R;
+  var S = {};
 
-  var isArray = function(x) {
-    return toString_.call(x) === '[object Array]';
-  };
-
-  var arity = function(n, f) {
-    var params = Array(n + 1).join('x').replace(/x/g, function(_, n) {
-      return n === 0 ? '$0' : ', $' + n;
-    });
-    return Function(  // jshint ignore:line
-      'f',
-      'return function(' + params + ') { return f.apply(this, arguments); }'
-    ).call(this, f);
-  };
-
-  var curry = function(f) {
-    var len = f.length;
-    return arity(len, function() {
-      var shortfall = len - arguments.length;
-      if (shortfall <= 0) {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -272,30 +251,26 @@
...
 
   //  last :: [a] -> Maybe a
-  var last = at(-1);
+  S.last = at(-1);
 
   //  tail :: [a] -> Maybe [a]
-  var tail = function(xs) {
-    return xs.length > 0 ? Just(slice_.call(xs, 1)) : Nothing();
-  };
+  S.tail = R.ifElse(R.isEmpty, Nothing, R.compose(Just, R.tail));
```
davidchambers: This reads nicely now: if `isEmpty` then `Nothing` else `Just` the `tail`.

index.js
```diff
@@ -307,82 +282,40 @@
...
-    var acc = Just(obj);
-    for (var idx = 0, len = keys.length; idx < len; idx += 1) {
-      acc = acc.chain(get(keys[idx]));
-    }
-    return acc;
+  S.gets = R.curry(function(keys, obj) {
+    return R.reduce(function(acc, key) {
+      return R.chain(get(key), acc);
+    }, Just(obj), keys);
   });
```
davidchambers: This gets us the fold-based implementation you suggested in #13, @benperez. :)

svozza: Nice...

benperez: :+1:

benperez:
:dromedary_camel:

PR #23, opened 2015-04-25
Add S.pluck functionality

### LINKED ISSUE #20: S.pluck?

svozza:
Resolves #20

### DIFF
--- a/index.js
+++ b/index.js
@@ -277,6 +277,20 @@
     return Nothing();
   });
 
+  //. pluck :: String -> [{String: *}] -> [Maybe *]
+  //.
+  //. Takes a list of objects and plucks the value of the specified key
+  //. for each object in the list. Returns the value wrapped in a Just
+  //. if an object has the key and a Nothing if it does not.
+  //.
+  //. ```javascript
+  //. > pluck('a', [{a: 1, b: 2}, {a: 4, b: 5}, {b: 3, c: 7}])
+  //. [Just(1), Just(4), Nothing()]
+  //. ```
+  S.pluck = R.curry(function(key, xs) {
+    return R.map(get(key), xs);
+  });
+
   //  object  ////////////////////////////////////////////////////////////////
 
   //  get :: String -> Object -> Maybe *
--- a/test/index.js
+++ b/test/index.js
@@ -835,6 +835,34 @@ describe('list', function() {
 
   });
 
+  describe('pluck', function() {
+
+    it('returns array of Justs for found keys', function() {
+      var xs = [{a: 1, b: 3}, {a: 2, b: 4}, {a: 3, b: 5}];
+      var maybeXs = [S.Just(1), S.Just(2), S.Just(3)];
+      assert.deepEqual(S.pluck('a', xs), maybeXs);
+    });
+
+    it('returns array of Nothings for keys not found', function() {
+      var xs = [{a: 1, b: 3}, {a: 2, b: 4}, {a: 3, b: 5}];
+      var maybeXs = [S.Nothing(), S.Nothing(), S.Nothing()];
+      assert.deepEqual(S.pluck('c', xs), maybeXs);
+    });
+
+    it('returns Just(undefined) for defined key with no value', function() {
+      var xs = [{a: 1, b: 3}, {a: void 0, b: 4}, {a: 3, b: 5}];
+      var maybeXs = [S.Just(1), S.Just(void 0), S.Just(3)];
+      assert.deepEqual(S.pluck('a', xs), maybeXs);
+    });
+
+    it('returns an array of Maybes for various values', function() {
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -277,6 +277,21 @@
...
+  //. if an object has the key and a Nothing if it does not.
+  //.
+  //. ```javascript
+  //. > pluck('a', [{a: 1, b: 2}, {a: 4, b: 5}, {b: 3, c: 7}])
+  //. [Just(1), Just(4), Nothing()]
+  //. ```
+
+  S.pluck = R.curry(function (key, xs) {
+    return R.map(S.get(key), xs);
+  });
```
buzzdecafe: since ramda is now a dep, I believe you could compose this as

`R.useWith(R.ap, R.map(R.prop))`

davidchambers:
Very nice! You're even using [Transcribe][1]-friendly `//.`-style comments. :)

I'll make a few minor inline comments, but overall this is excellent.


index.js
```diff
@@ -277,6 +277,21 @@
     return Nothing();
   });
 
+  //. pluck :: String -> [Object] -> [Maybe *]
```
davidchambers: It would be nice if we could use `String -> [{String: a}] -> [Maybe a]` here, but this isn't accurate since the objects may be "records" with mixed value types (e.g. `{ a :: Number, b :: Boolean }`).

index.js
```diff
@@ -277,6 +277,21 @@
...
+  //.
+  //. Takes a list of objects and plucks the value of the specified key
+  //. for each object in the list. Returns the value wrapped in a Just
+  //. if an object has the key and a Nothing if it does not.
+  //.
+  //. ```javascript
+  //. > pluck('a', [{a: 1, b: 2}, {a: 4, b: 5}, {b: 3, c: 7}])
+  //. [Just(1), Just(4), Nothing()]
+  //. ```
+
```
davidchambers: I prefer no empty line between "docstring" and function definition.

index.js
```diff
@@ -277,6 +277,21 @@
...
+  S.pluck = R.curry(function (key, xs) {
```
davidchambers: No space after `function`. :)

test/index.js
```diff
@@ -835,6 +835,33 @@ describe('list', function() {
...
+      var maybeXs = [S.Just(1), S.Just(void 0), S.Just(3)];
+      assert.deepEqual(S.pluck('a', xs), maybeXs);
+    });
+
+    it('returns an array of Maybes for various values', function() {
+      var xs = [{a: 1}, {a: void 0}, {a: 4}, {b: 1}];
+      var maybeXs = [S.Just(1), S.Just(void 0), S.Just(4), S.Nothing()];
+      assert.deepEqual(S.pluck('a', xs), maybeXs);
+    });
+  });
```
davidchambers: Please include an empty line between lines 862 and 863.

index.js
davidchambers: I imagine we'd need to replace `R.prop` with `S.get`.

davidchambers: Point-free may not be possible since `S.get` is defined further down the file.

buzzdecafe: indeed. any though about using `prop` instead for tighter compatibility?

index.js
```diff
@@ -277,6 +277,21 @@
...
+    return R.map(S.get(key), xs);
```
davidchambers: Let's use `get` rather than `S.get` here.

index.js
svozza: I know that it works but I'm having a bit of difficulty understand why. Could you explain a bit?

svozza:
Done. I'll squash it all into one commit once you're happy.

index.js
svozza: Should I change it to something like [{String: *}]?

index.js
davidchambers: > any though about using `prop` instead for tighter compatibility?

I've thought about it, certainly. I prefer "get" though as it's a verb. It's nice to be able to talk about

```javascript
R.pipe(S.get('a'),
       R.chain(S.get('b')),
       R.chain(S.get('c')))
```

as "chaining gets". #13 added `S.gets` which does multiple "gets". As a noun, "prop" does not work as well for these cases.

index.js
davidchambers: Yes, I like that idea. The `*` would provide a visual link between the type of the second argument and the return type.

davidchambers:
:deciduous_tree: Thanks, @svozza! Squash the commits and I'll merge. :)

svozza:
Done! Right, it's getting pretty late here so I'm going to hit the hay. :)

Issue #24, opened 2015-04-27
Support ES6 Map Type


benperez:
[Maps](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) are a new feature of ES6 that could replace the de facto use of objects as dictionaries in JS.

Haskell has a [Data.Map](https://downloads.haskell.org/~ghc/latest/docs/html/libraries/containers/Data-Map-Lazy.html#t:Map) module that has the following [API](https://downloads.haskell.org/~ghc/latest/docs/html/libraries/containers/Data-Map-Lazy.html#v:lookup) for getting a value for a given key.
```haskell
lookup :: Ord k => k -> Map k a -> Maybe a 
```

This feels like a pretty natural fit for sanctuary and mimics the way we're using `S.get` in a lot of places currently. How would you feel about adding some functions that explicitly operate on ES6 Maps with a nice API using the Maybe type?

davidchambers:
Currently we have:

    get :: String -> Object -> Maybe *
    gets :: [String] -> Object -> Maybe *

What do you have in mind? Perhaps:

    mapGet :: * -> Map -> Maybe *
    mapGets :: [*] -> Map -> Maybe *

benperez:
The basics would be the following:
```haskell
insert :: k -> a -> Map k a -> Map k a 
delete :: k -> Map k a -> Map k a
lookup :: k -> Map k a -> Maybe a 
```
The `Data.Map.Lazy` API is pretty huge but I think there are a few natural directions to go from there. First would be convenience functions like `toList`, `fromList`, `keys`, and `elems`. From there it could be useful to define the [map](https://downloads.haskell.org/~ghc/latest/docs/html/libraries/containers/Data-Map-Lazy.html#g:14) and [fold](https://downloads.haskell.org/~ghc/latest/docs/html/libraries/containers/Data-Map-Lazy.html#g:15) families of functions and possibly `intersection`, `union`, and `difference`.

Advantages of adding this to Sanctuary:

1. A lookup API using Sanctuary's `Maybe` type feels like a very natural fit.
2. We could replace the somewhat hacky use of JavaScript objects as Maps (we can use non-string keys and readily access the size of a map).
[...]

PR #25, opened 2015-04-29
add insert delete and lookup

benperez:
Here's a preliminary proposal for what #24 could look like. There are definitely a lot of open questions about this but I figured it would be helpful to have something to at least look at.

I've avoided creating a new type for `S.Map` for the time being and am simply using the ES6 `Map` type.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -4,7 +4,7 @@ dependencies:
 
 machine:
   node:
-    version: 0.10.26
+    version: 0.12.0
 
 test:
   override:
--- a/index.js
+++ b/index.js
@@ -328,4 +328,27 @@
     return JSON.parse(s);
   });
 
+  //  Map  ////////////////////////////////////////////////////////////////////
+
+  // insert :: k -> a -> Map k a -> Map k a
+  S.insert = R.curry(function(key, value, map) {
+    var mapCopy = new Map(map);
+    mapCopy.set(key, value);
+    return mapCopy;
+  });
+
+  // delete :: k -> Map k a -> Map k a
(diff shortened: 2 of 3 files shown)


davidchambers:
`insert` and `delete` are [`R.assoc`][1] and [`R.dissoc`][2] but for Map. I suggest we first discuss in a Ramda issue the possibility of Ramda supporting Map in its "object" functions. We could think of objects as "map-like" in the same way we think of arguments objects as "array-like". I'm not sure this'll fly with @CrossEye, but it can't hurt to find out. :)


[1]: http://ramdajs.com/docs/#assoc
[2]: http://ramdajs.com/docs/#dissoc

benperez:
Assuming something like https://github.com/ramda/ramda/pull/1056 gets merged into ramda I think that just leaves us with `S.lookup`

buzzdecafe:
> Assuming something like ramda/ramda#1056 gets merged into ramda

I think that's a safe assumption. i'd be very surprised if we just punted on ES6 features

davidchambers:
Rather than add `S.lookup` we could update `S.get` (and `S.gets`) to support Map objects.

benperez:
I like the idea of paralleling the ramda API. `S.gets` feels a little weird with nested dictionaries. When I think about objects I think about collections of key value pairs where the keys are all strings and the values are of heterogeneous type. When I think about Maps I think of `Data.Map k a` where k and a are fixed types. Given that we're using JS though this is probably not a realistic use case.

benperez:
Closing given the discussion in plaid/sanctuary/issues/24

PR #26, opened 2015-05-07
control: add S.and and S.xor

davidchambers:
> `S.and`, `S.or`, and `S.xor` are defined in terms of a private `toBoolean` function which dispatches to each operand's `toBoolean` method if available. `Just.prototype.or` and `Nothing.prototype.or` are no longer necessary so are removed in this commit.

I don't know whether `S.xor` will be useful in practice, but I know how to find out. :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -100,14 +100,14 @@
     return maybe instanceof Nothing;
   };
 
-  //  Nothing#or :: Maybe a -> Maybe a
-  Nothing.prototype.or = function(maybe) {
-    return maybe;
-  };
-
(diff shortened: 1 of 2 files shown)


benperez:
Looking very good. Are you going to update the README as well?

davidchambers:
> Are you going to update the README as well?

Yes, but not in this pull request. I'll open a separate pull request to integrate [Transcribe][1] and add "docstrings" to the source file.


benperez:
:dromedary_camel:

PR #27, opened 2015-05-09
either: implement Apply, Applicative, Chain, and Monad

davidchambers:
I *think* this is correct, but I'd greatly appreciate a review.

### DIFF
--- a/index.js
+++ b/index.js
@@ -181,6 +181,13 @@
     throw new Error('Cannot instantiate Either');
   };
 
+  //  Either.of :: b -> m a b
+  Either.of = function(x) {
+    return new Right(x);
+  };
+
+  Either.prototype.of = Either.of;
+
   Either.prototype.type = Either;
 
   var Left = S.Left = function Left(value) {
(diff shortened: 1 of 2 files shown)


buzzdecafe:
LGTM. :cow2:

2 recommendations: 
1. use the `types` test file from ramda-fantasy or something like it for legal compliance; and 
2. use jsverify or something like it for property-based testing

davidchambers:
Thank you for the recommendations, @buzzdecafe.

> use the `types` test file from ramda-fantasy or something like it for legal compliance

Good idea. This makes me wonder whether the ramda-fantasy compliance tests could live in their own package. It would be great for various libraries to share these tests.

> use jsverify or something like it for property-based testing

I watched John Hughes present the Clojure/West keynote last year: [Testing the Hard Stuff and Staying Sane][1]. I'm convinced of the value of property-based testing, but I've yet to actually use it. I should change that!


[1]: https://www.youtube.com/watch?v=zi0rHwfiX1Q

buzzdecafe:
> the ramda-fantasy compliance tests could live in their own package. 

interesting idea. might make sense to "donate" 'em to fantasyland

CrossEye:
LGTM

@buzzdecafe: How convinced are you by the rf test suite?  I remember a lot of doubts for a while about how much it was really telling us.

buzzdecafe:
that's what jsverify is for

PR #29, opened 2015-05-11
maybe, either: define "toString"

davidchambers:
This grants Maybe objects and Either objects the following property:

    eval(R.toString(x)) ≍ x

While this is primarily useful in the REPL, it also makes these values compatible with [`R.memoize`][1].

This change was made possible by ramda/ramda#924 and the publication of Ramda v0.14.0. I love it when a plan comes together. :grinning:

[1]: http://ramdajs.com/docs/#memoize

### DIFF
--- a/index.js
+++ b/index.js
@@ -110,6 +110,11 @@
     return false;
   };
 
+  //  Nothing#toString :: -> String
+  Nothing.prototype.toString = function() {
+    return 'Nothing()';
+  };
+
   var Just = S.Just = function Just(value) {
     if (!(this instanceof Just)) {
       return new Just(value);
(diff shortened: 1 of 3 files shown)


benperez:
I missed the ramda discussion. What makes this useful in the REPL? Is it calling `R.toString` on a value that exposes a toString method? I'm also guessing that `R.memoize` does dispatch now?

davidchambers:
> What makes this useful in the REPL?

The REPL's output is often ambiguous:

```node
> {}
{}
> S.Nothing()
{}
```

```node
> S.Just(42)
{ value: 42 }
> S.Left(42)
{ value: 42 }
> S.Right(42)
{ value: 42 }
```

`R.toString` can resolve these ambiguities:

```node
> R.toString({})
'{}'
> R.toString(S.Nothing())
'Nothing()'
> R.toString(S.Just(42))
'Just(42)'
> R.toString(S.Left(42))
'Left(42)'
> R.toString(S.Right(42))
'Right(42)'
```

> I'm also guessing that `R.memoize` does dispatch now?

Yes. It uses `R.toString` to produce canonical string representations, and `R.toString` dispatches to a custom `toString` method if available.

benperez:
Cool! :dromedary_camel:

PR #30, opened 2015-05-13
either: implement Semigroup

davidchambers:
@joneshf and I discussed Either on Gitter. We couldn't think of a way to implement [Monoid][1] because Either doesn't have a canonical empty value. [Semigroup][2], though, seems completely reasonable. `Left.prototype.concat` and `Right.prototype.concat` are similar to `Just.prototype.concat`.

```javascript
> R.concat(S.Left('abc'), S.Left('def'))
Left('abcdef')
> R.concat(S.Left('abc'), S.Right([1, 2, 3]))
Right([1, 2, 3])
> R.concat(S.Right([1, 2, 3]), S.Left('abc'))
Right([1, 2, 3])
> R.concat(S.Right([1, 2, 3]), S.Right([4, 5, 6]))
Right([1, 2, 3, 4, 5, 6])
```

[1]: https://github.com/fantasyland/fantasy-land#monoid
[2]: https://github.com/fantasyland/fantasy-land#semigroup

### DIFF
--- a/index.js
+++ b/index.js
@@ -218,6 +218,11 @@
     return this;
   };
 
+  //  Left#concat :: Either a b -> Either a b
+  Left.prototype.concat = function(either) {
+    return R.is(Left, either) ? Left(this.value.concat(either.value)) : either;
+  };
+
   //  Left#equals :: Either a b -> Boolean
   Left.prototype.equals = function(either) {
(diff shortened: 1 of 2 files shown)


benperez:
:dromedary_camel:

PR #31, opened 2015-05-13
lint: allow unused parameters

davidchambers:
<http://jshint.com/docs/options/#unused>:

> This can be set to `vars` to only check for variables, not function parameters, or `strict` to check all variables and parameters. The default (true) behavior is to allow unused parameters that are followed by a used parameter.

Removing a function's unreferenced parameters is not an option. Doing so would alter the function's `length`, an extremely significant property in a world of currying and partial application.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -14,7 +14,7 @@
   "nonew": true,
   "plusplus": true,
   "undef": true,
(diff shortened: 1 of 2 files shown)


charleswli:
:palm_tree:

PR #32, opened 2015-05-14
test: point-free predicates

davidchambers:
This makes assertions for functions which may throw a bit less offensive. :)


benperez:
Much cleaner. :dromedary_camel:

PR #33, opened 2015-05-14
either: define "toBoolean"

davidchambers:
This allows `S.and` and `S.or` to be applied to values of the Either type:

```javascript
> S.and(S.Left('abc'), S.Left('xyz'))
Left('abc')
> S.and(S.Left('abc'), S.Right(1))
Left('abc')
> S.and(S.Right(1), S.Left('abc'))
Left('abc')
> S.and(S.Right(1), S.Right(2))
Right(2)
> S.or(S.Left('abc'), S.Left('xyz'))
Left('xyz')
> S.or(S.Left('abc'), S.Right(1))
Right(1)
> S.or(S.Right(1), S.Left('abc'))
Right(1)
> S.or(S.Right(1), S.Right(2))
Right(1)
```

`S.xor` cannot be applied to values of the Either type since the type lacks a canonical empty value:

```javascript
> S.xor(S.Left('abc'), S.Left('xyz'))
TypeError: Left("abc") does not have an "empty" method
```

For the sake of consistency, an error is thrown even when the empty value is not needed:

```javascript
> S.xor(S.Left('abc'), S.Right(1))
TypeError: Left("abc") does not have an "empty" method
```

This pull request also adds nice error reporting for the case where the arguments to `S.and`, `S.or`, or `S.xor` lack a `toBoolean` method:

```javascript
> S.and({x: 1, y: 1}, {x: 2, y: 2})
TypeError: {"x": 1, "y": 1} does not have a "toBoolean" method
```

@michaelckelly, this pull request may be of particular interest to you in light of our discussion yesterday about how to take one Either *or* another.

### DIFF
--- a/index.js
+++ b/index.js
@@ -233,6 +233,11 @@
     return this;
   };
 
+  //  Left#toBoolean :: -> Boolean
+  Left.prototype.toBoolean = function() {
+    return false;
+  };
+
   //  Left#toString :: -> String
   Left.prototype.toString = function() {
     return 'Left(' + R.toString(this.value) + ')';
@@ -271,6 +276,11 @@
     return new Right(f(this.value));
   };
 
+  //  Right#toBoolean :: -> Boolean
+  Right.prototype.toBoolean = function() {
+    return true;
+  };
+
   //  Right#toString :: -> String
   Right.prototype.toString = function() {
     return 'Right(' + R.toString(this.value) + ')';
@@ -298,11 +308,18 @@
 
   //  toBoolean :: * -> Boolean
   var toBoolean = function(x) {
-    switch (true) {
-      case R.is(Array, x):    return x.length > 0;
-      case R.is(Boolean, x):  return x;
-      default:                return x.toBoolean();
-    }
+    if (R.is(Array, x))               return x.length > 0;
+    if (R.is(Boolean, x))             return x;
+    if (R.is(Function, x.toBoolean))  return x.toBoolean();
+    throw new TypeError(R.toString(x) + ' does not have a "toBoolean" method');
+  };
+
+  //  empty :: a -> a
+  var empty = function(x) {
+    if (R.is(Array, x))               return [];
+    if (R.is(Boolean, x))             return false;
+    if (R.is(Function, x.empty))      return x.empty();
+    throw new TypeError(R.toString(x) + ' does not have an "empty" method');
   };
 
   //  and :: a -> a -> a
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -320,12 +337,10 @@
...
-    switch (true) {
-      case toBoolean(x) !== toBoolean(y): return or(x, y);
-      case R.is(Array, x):                return [];
-      case R.is(Boolean, x):              return false;
-      default:                            return x.empty();
-    }
+    var xBool = toBoolean(x);
+    var yBool = toBoolean(y);
+    var xEmpty = empty(x);
+    return xBool !== yBool ? or(x, y) : xEmpty;
```
davidchambers: Looking at this my first reaction is to suggest this change:

```diff
-    var xEmpty = empty(x);
-    return xBool !== yBool ? or(x, y) : xEmpty;
+    return xBool !== yBool ? or(x, y) : empty(x);
```

`empty` has side effects, so this is not equivalent. If `x` does not have an `empty` method we want to throw a descriptive TypeError regardless of whether the empty value is actually needed. Otherwise, `S.xor(S.Left('abc'), S.Right(1))` will evaluate to `Right(1)` but `S.xor(S.Right(1), S.Right(2))` will throw.

michaelckelly:
This will be quite useful - thanks for adding this, @davidchambers.

I naively expected:

```javascript
S.or(S.Left('a'), S.Left('b')).equals(S.Nothing())
```
But I think that was due to my own use case clouding my judgement (and not fully considering what `or` should do!)

LGTM - but I'll defer to others here.

davidchambers:
Thanks for the review, @michaelckelly!

> I naively expected:
>
> ```javascript
> S.or(S.Left('a'), S.Left('b')).equals(S.Nothing())
> ```

As we discussed in HipChat, it's important that we respect the type signature:

    S.or :: a -> a -> a

`S.or` takes two values of type `a` and returns a value of type `a`. We must return an Either given two Either values.

An additional requirement is that `S.or` must return one of its arguments. This is consistent with JavaScript's `||` operator, which always evaluates to the value of one of its operands (`0 || 1`, for example, evaluates to `1` rather than `true`).

index.js
benperez: It may be worth leaving a comment here explaining that empty() has the intended side effecting of throwing a `TypeError`.

Also, what's special about the empty value of the first element? Should you also verify that the second element has an empty value?

benperez:
This looks good. I like the idea of using these logical functions on Sanctuary defined types. The toBoolean definitions for built in types feel a little bit like re-inventing thruthiness.

index.js
davidchambers: > It may be worth leaving a comment here explaining that empty() has the intended side effecting of throwing a `TypeError`.

We certainly could, though making the change from my previous comments would cause tests to fail, so a comment isn't necessary to prevent regression.

> Also, what's special about the empty value of the first element?

Nothing. `empty(x)` and `empty(y)` should be equivalent, since `x` and `y` are values of the same type (as asserted by `assertTypeMatch(x, y)`).

davidchambers:
> The toBoolean definitions for built in types feel a little bit like re-inventing thruthiness.

Array is a [Functor][1] and a [Semigroup][2]. It doesn't seem like a stretch to treat Array as a [Monoid][3] as well, since the type has a well-defined empty value, `[]`.

Boolean feels a bit odd alongside Array, Either, and Maybe, but it seems reasonable for `S.and` and `S.or` to support every type whose values can be divided into empty/false and non-empty/true, and for `S.xor` to support every type which satisfies this condition and has exactly one empty/false value.


[1]: https://github.com/fantasyland/fantasy-land#functor
[2]: https://github.com/fantasyland/fantasy-land#semigroup
[3]: https://github.com/fantasyland/fantasy-land#monoid

PR #35, opened 2015-05-17
Add S.match function

### LINKED ISSUE #28: match ?
Would it make sense to add a `match` function?

```
R.head(R.match(/xxx/, 'abcde')) // returns TypeError: Cannot read property '0' of null
```
http://bit.ly/1HasPS6

We'd be "safer" with something like
```
smatch = curry (re, s) ->
	res = R.match re, s
	return if res then Just(res) else Nothing()
```

Would it make sense to just add "safe" versions of all functions in Ramda that might otherwise need null checks? Is there some smart generic way to wrap Ramda functions and make them return `Maybe`s instead?

svozza:
Resolves #28 . I've actually been using R.compose with Sanctuary to create a safe version of `R.match` quite a few times recently so I thought I'd have a stab at a PR for this. The implementation is simple enough:

```javascript
S.match = R.curry(function (regex, s) {
    return R.compose(S.toMaybe, R.match(regex))(s);
  });
```

But I was having problems writing the tests. Obviously because `R.match` returns an array, if I do something like:

```javascript
assert(S.match(/abcd/, 'abcdefg').equals(S.Just(['abcd']))
```

it fails because `Just.equals` is checking by reference.

If I do

```javascript
assert.deepEqual(S.match(/abcd/, str), S.Just(['abcd']));
```

it fails because `match` gives back this mad arraylike object with the object keys `input` and `index` so I find myself having to do this in the setup of the tests to get everything to work:

```javascript
var str = 'abcdefg';
var search = /abcd/;
var result = ['abcd'];
result.index = 0;
result.input = 'abcdefg';
...
assert.deepEqual(S.match(search, str), S.Just(result));
...
//etc
```

It feels a bit weird to do so but I'm not really able to think of alternatives. Any help would be appreciated.

### DIFF
--- a/index.js
+++ b/index.js
@@ -425,4 +425,21 @@
     return JSON.parse(s);
   });
 
+  //  regexp  ////////////////////////////////////////////////////////////////
+
+  //. match :: RegExp -> String -> Maybe [Maybe String]
+  //.
+  //. Takes a regular expression and tests it against the supplied string
+  //. returning a Nothing if there are no matches or a Just containing an
+  //. array of Maybe types corresponding to the matches found.
+  //.
+  //. ```javascript
+  //. > S.match(/abcd/, 'abcdefg')
+  //. Just([Just("abcd")])
+  //.
+  //. > S.match(/(good)?bye/, 'bye')
+  //. Just([Just("bye"), Nothing()])
+  //. ```
+  S.match = R.curry(R.compose(R.map(R.map(S.toMaybe)), S.toMaybe, R.match));
+
 }.call(this));
--- a/test/index.js
+++ b/test/index.js
@@ -1338,3 +1338,31 @@ describe('parse', function() {
   });
 
 });
+
+describe('regexp', function() {
+
+  describe('match', function() {
+
+    it('returns a Just containing an array of Justs', function() {
+      eq(S.match(/abcd/, 'abcdefg').toString(), 'Just([Just("abcd")])');
+    });
+
+    it('supports global patterns', function() {
+      eq(S.match(/[a-z]a/g, 'bananas').toString(),
+        'Just([Just("ba"), Just("na"), Just("na")])');
+    });
+
+    it('supports (optional) capturing groups', function() {
+      eq(S.match(/(good)?bye/, 'goodbye').toString(),
+        'Just([Just("goodbye"), Just("good")])');
+      eq(S.match(/(good)?bye/, 'bye').toString(),
+        'Just([Just("bye"), Nothing()])');
+    });
+
+    it('returns a Nothing() if no match', function() {
+      assert(S.match(/zzz/, 'abcdefg').equals(S.Nothing()));
+    });
+
+  });
+
+});


davidchambers:
You could use `toString` to compare Maybe values:

```javascript
eq(S.Just([1, 2, 3]).toString(), S.Just([1, 2, 3]).toString());
```

index.js
```diff
@@ -425,4 +425,19 @@
     return JSON.parse(s);
   });
 
+  // Regex  //////////////////////////////////////////////////////////////////
+
+  //. match :: RegEx -> String -> Maybe [*]
```
davidchambers: The type of the first argument should be written `RegExp` to match the name of the constructor function. Also, could the `*` be replaced with `String`?

index.js
```diff
@@ -425,4 +425,19 @@
...
+  //. returning a Nothing if there is no match or a Just containing an
+  //. array of the the matched results.
+  //.
+  //. ```javascript
+  //. > S.match(/abcd/, 'abcdefg')
+  //. Just(['abcd'])
+  //. ```
+  S.match = R.curry(function(regex, s) {
+    return R.compose(S.toMaybe, R.match(regex))(s);
+  });
```
davidchambers: Could this simply be defined as `S.match = R.compose(S.toMaybe, R.match);` (or the `R.pipe` equivalent)?

index.js
svozza: Yep, makes sense.

index.js
svozza: I did try that but then the invariants test cases failed for me.

svozza:
Ah yes, I like that `toString` solution.

index.js
davidchambers: Aha! It's nice to see those tests catch something. :)

My intuition suggested that since `R.match` is a curried binary function, composing it with `S.toMaybe` would produce a curried binary function. `R.compose`, though, doesn't know that `R.match` is curried.

We could curry the resulting function ourselves:

```javascript
S.match = R.curry(R.compose(S.toMaybe, R.match));
```

This should count as one vote in favour of ramda/ramda#736.

svozza: Ah yes, that makes sense. We could also go a step further, given how common this operation is going to be and define an internal function that can be reused:

```javascript
function sanctify(f) {
  return R.curry(R.compose(S.toMaybe, f));
}

S.match = sanctify(R.match);
```

davidchambers:
Currently, we have:

```haskell
match :: RegExp -> String -> Maybe [String]
```

I've realized there's a problem with this signature: optional capturing groups can result in `undefined` appearing in the list of matches. For example:

```javascript
> S.match(/(good)?bye/, 'goodbye')
Just(['goodbye', 'good'])
> S.match(/(good)?bye/, 'bye')
Just(['bye', undefined])
```

This suggests that the type should in fact be:

```haskell
match :: RegExp -> String -> Maybe [Maybe String]
```

I've opened purescript/purescript-strings#31 for discussion, since [`match`][1] currently has the "faulty" type.


[1]: https://github.com/purescript/purescript-strings/blob/master/docs/Data.String.Regex.md#match

test/index.js
```diff
@@ -1338,3 +1338,26 @@ describe('parse', function() {
...
+
+describe('regex', function() {
+
+  describe('match', function() {
+    var str = 'abcdefg';
+    var search = /abcd/;
+    var result = ['abcd'];
+
+    it('returns a Maybe', function() {
+      eq(S.fromMaybe({}, S.match(search, str)).toString(), result.toString());
```
davidchambers: We could remove the use of `S.fromMaybe`. I'd also prefer literals in place of `search`, `str`, and `result.toString()`, so the assertion requires no context. I would write:

```javascript
eq(S.match(/abcd/, 'abcdefg').toString(), 'Just(["abcd"])');
```

index.js
davidchambers: Good idea. Also, `sanctify` is an incredible function name!

:church:

index.js
```diff
@@ -425,4 +429,17 @@
     return JSON.parse(s);
   });
 
+  // Regex  //////////////////////////////////////////////////////////////////
+
+  //. match :: RegExp -> String -> Maybe [String]
+  //. Takes a regular expression and tests it against the supplied string
+  //. returning a Nothing if there is no match or a Just containing an
+  //. array of the the matched results.
```
davidchambers: s/the the/the/

index.js
```diff
@@ -425,4 +429,17 @@
     return JSON.parse(s);
   });
 
+  // Regex  //////////////////////////////////////////////////////////////////
+
+  //. match :: RegExp -> String -> Maybe [String]
+  //. Takes a regular expression and tests it against the supplied string
```
davidchambers: Please include an empty line after the type signature:

```diff
   //. match :: RegExp -> String -> Maybe [String]
+  //.
   //. Takes a regular expression and tests it against the supplied string
```

index.js
svozza: Hehe. Yeah, I particularly like that one.

index.js
svozza: All done.

svozza:
Should I hold off on the type signature stuff until there's some replies to the Purescript issue?

index.js
```diff
@@ -425,4 +429,18 @@
     return JSON.parse(s);
   });
 
+  // Regex  //////////////////////////////////////////////////////////////////
```
svozza: Oops, I thought I'd already done that!

davidchambers:
Currently the tests do not cover capturing groups. Let's add assertions for `S.match(/(good)?bye/, 'goodbye')` and `S.match(/(good)?bye/, 'bye')`. It would also be useful to test a global pattern via `S.match(/x/g, 'xxx')` or similar.

index.js
```diff
@@ -425,4 +429,18 @@
     return JSON.parse(s);
   });
 
+  //  regexp  /////////////////////////////////////////////////////////////////
```
davidchambers: The linter is complaining that this line exceeds the maximum line length by one character. Sorry for the hassle.

I appreciate your responsiveness. I :heart: having you as a contributor, @svozza!

davidchambers:
> Should I hold off on the type signature stuff until there's some replies to the Purescript issue?

That's a good idea. Hopefully some of the bright minds of the PureScript world chime in on the issue in the coming days. :)

In the interim, let's add tests for optional capturing groups to document the current (problematic) behaviour.

index.js
svozza: Yeah, that one was a bit puzzling and then I realised I'd added another space as well as the 'p' character!

svozza:
How are we looking now?

index.js
davidchambers: Vim's Replace mode comes in handy here. :)

For the benefit of Vim users who may read this, <kbd>Shift</kbd>-<kbd>R</kbd> switches to Replace mode.

davidchambers:
We're looking good! Thanks for adding those tests.

svozza:
Excellent! :smile:

test/index.js
```diff
@@ -1338,3 +1338,22 @@ describe('parse', function() {
...
 
 });
+
+describe('regex', function() {
+
+  describe('match', function() {
+
+    it('returns a Just if there is a match', function() {
+      eq(S.match(/(good)?bye/, 'goodbye').toString(), 'Just(["goodbye", "good"])');
+      eq(S.match(/([a-z]a)/g, 'bananas').toString(), 'Just(["ba", "na", "na"])');
```
davidchambers: Nice pattern, though the capturing group is unnecessary. Capturing groups are ignored when the `g` flag is present.

It would be nice to separate these tests into different groups:

```javascript
it('returns a Just if there is a match', ...);

it('returns a Nothing() if no match', ...);

it('supports global patterns', ...);

it('supports (optional) capturing groups', ...);
```

davidchambers: Sorry to add one more comment after giving the `:thumbsup:`!

svozza: I lifted that banana one direct from the Ramda docs! Dunno why but I've always found it funny so wanted to add it in.

svozza:
One last thing, is there any chance you could do an NPM release soon? I'd really like to be able to use this function and the `gets` one as well.

davidchambers:
> One last thing, is there any chance you could do an NPM release soon? I'd really like to be able to use this function and the `gets` one as well.

Absolutely! Once we merge this and #36 I'll publish v0.5.0. I look forward to using the new features myself. :)

> Hopefully some of the bright minds of the PureScript world chime in on the issue in the coming days.

@paf31 is on board with the idea of returning `[Maybe String]` in [purescript/purescript-strings][1]. I suggest Sanctuary do the same.


[1]: https://github.com/purescript/purescript-strings

svozza:
So are we thinking something like this:

```javascript
S.match = R.curry(R.compose(S.toMaybe, R.map(S.toMaybe), R.match));
```

davidchambers:
Sorry for my slow reply, @svozza. I had my head down working on something yesterday and am only now catching up on things I missed.

> ```javascript
> S.match = R.curry(R.compose(S.toMaybe, R.map(S.toMaybe), R.match));
> ```

This isn't quite right. Since `R.match` may return null, next we must apply `S.toMaybe`. Then, since we have a `Maybe [String?]`, we need to apply `R.map(S.toMaybe)` to the `[String?]` if the match succeeded, giving:

```javascript
// S.match :: RegExp -> String -> Maybe [Maybe String]
S.match = R.curry(R.pipe(R.match, S.toMaybe, R.map(R.map(S.toMaybe))));
```

You could use `R.compose` rather than `R.pipe` if you prefer.

svozza:
Yeah, I've been a bit snowed under myself (I will get to review your doc story by the end of today, I promise!) so I'm only getting back to this now.

svozza:
Sad to see `sanctify` go though. :(

davidchambers:
Thanks for updating this pull request, @svozza!

> Sad to see `sanctify` go though. :(

So am I. It had such a fitting name. We might decide it's useful in the future, though it seems that in practice many Ramda functions will require different treatment. `R.indexOf`, for example, returns `-1` in the failure case, so we'll need to treat *that* as the empty value.

I'll make a few final in-line comments for tweaks I'd like you to make, but we'll merge this soon. :)

index.js
davidchambers: This doesn't *quite* line up:

```console
$ ack /// index.js
  //  combinator  ////////////////////////////////////////////////////////////
  //  maybe  /////////////////////////////////////////////////////////////////
  //  either  ////////////////////////////////////////////////////////////////
  //  control  ///////////////////////////////////////////////////////////////
  //  list  //////////////////////////////////////////////////////////////////
  //  object  ////////////////////////////////////////////////////////////////
  //  parse  /////////////////////////////////////////////////////////////////
  //  regexp  /////////////////////////////////////////////////////////////////
```

index.js
```diff
@@ -425,4 +425,22 @@
...
+
+  //. match :: RegExp -> String -> Maybe [Maybe String]
+  //.
+  //. Takes a regular expression and tests it against the supplied string
+  //. returning a Nothing if there are no matches or a Just containing an
+  //. array of Maybe types corresponding to the matches found.
+  //.
+  //. ```javascript
+  //. > S.match(/abcd/, 'abcdefg')
+  //. Just([Just('abcd')])
```
davidchambers: Let's use double quotes on this line, as this will be required by #36. Stick with single quotes on the input line, though.

index.js
```diff
@@ -425,4 +425,22 @@
...
+  //.
+  //. > S.match(/(good)?bye/, 'bye')
+  //. Just([Just("bye"), Nothing()])
+  //. ```
+  //
```
davidchambers: Let's remove this line.

test/index.js
```diff
@@ -1338,3 +1338,31 @@ describe('parse', function() {
...
+    it('supports global patterns', function() {
+      eq(S.match(/[a-z]a/g, 'bananas').toString(),
+        'Just([Just("ba"), Just("na"), Just("na")])');
+    });
+
+    it('supports (optional) capturing groups', function() {
+      eq(S.match(/(good)?bye/, 'bye').toString(),
+        'Just([Just("bye"), Nothing()])');
+      eq(S.match(/(good)?bye/, 'goodbye').toString(),
+        'Just([Just("goodbye"), Just("good")])');
```
davidchambers: Let's switch the order of the `'bye'` and `'goodbye'` tests above. For some reason it feels right to me to test the success case (`'goodbye'`) before the partial failure case (`'bye'`).

test/index.js
```diff
@@ -1338,3 +1338,31 @@ describe('parse', function() {
   });
 
 });
+
+describe('regex', function() {
```
davidchambers: Let's change the heading to `'regexp'` to match the section heading in the source code.

index.js
svozza: Ha. You''re nothing if not thorough David!

test/index.js
svozza: Dammit, I thought I'd gotten rid of all those. Done.

index.js
svozza: Ah yes, forgot about that.

svozza:
Yeah, I thought `sanctify` would come in quite handy but actually when I went through the Ramda function list I found that we'd pretty much done all the `null`/`undefined` returning functions.

davidchambers:
:deciduous_tree: Thank you for all your work on this pull request, @svozza.

svozza:
My pleasure. I might have a crack at `indexOf` tonight if I get the chance.

PR #36, opened 2015-05-18
add documentation and Transcribe-powered prepublish script

### LINKED ISSUE #22: generate documentation via Transcribe
<https://github.com/plaid/transcribe>

davidchambers:
Closes #22

Until @svozza contributed #19, Sanctuary's API was completely undocumented. It's currently partially documented. This pull request documents the rest of the API, and adds a section on types and how to read Sanctuary's type signatures.

Points of note:

  - The source of the documentation is __index.js__. __README.md__ is now derived from the source via [Transcribe][1].

  - I've included the changes to [__README.md__][2] for the benefit of those reviewing the pull request. I will revert these changes before merging this pull request as the file will be updated automatically with each release.

  - To generate [__README.md__][2] I ran `VERSION=0.4.0 make`. As a result, the line numbers to which the headings in the API section link are not correct. This will not be a problem when the file is updated as part of the v0.5.0 release process.

  - Several sections contain "TK" as their body. I'm working on these, and will update the pull request as I complete them.

  - The doctests are tested via a Unix pipeline which transforms

    ```javascript
    > S.head([1, 2, 3])
    Just(1)
    > S.head([])
    Nothing()
    ```

    into something like

    ```javascript
    assert.strictEqual(R.toString(S.head([1, 2, 3])), 'Just(1)');
    assert.strictEqual(R.toString(S.head([])), 'Nothing()');
    ```

    via sed, which is then piped to node for evaluation. Note that single quotes and double quotes are not interchangeable on the output lines, since each output line must exactly match the result of applying `R.toString` to the preceding expression.

Once this pull request and #35 have been merged I will publish v0.5.0, which several of us are eagerly awaiting. I'd love feedback on this pull request from @benperez, @Bpless, @buzzdecafe, @CrossEye, @michaelckelly, @svozza, and anyone else who's interested. :)

[2]: https://github.com/plaid/sanctuary/blob/dc-transcribe/README.md

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,15 +2,25 @@ ISTANBUL = node_modules/.bin/istanbul
 JSCS = node_modules/.bin/jscs
 JSHINT = node_modules/.bin/jshint
 NPM = npm
-XYZ = node_modules/.bin/xyz --repo git@github.com:plaid/sanctuary.git
+TRANSCRIBE = node_modules/.bin/transcribe
+XYZ = node_modules/.bin/xyz --repo git@github.com:plaid/sanctuary.git --script scripts/prepublish
 
-SRC = $(shell find . -name '*.js' -not -path './coverage/*' -not -path './node_modules/*')
+
+.PHONY: all
+all: README.md
+
+README.md: index.js
+	$(TRANSCRIBE) \
+	  --heading-level 4 \
+	  --url 'https://github.com/plaid/sanctuary/blob/v$(VERSION)/{filename}#L{line}' \
+	  -- $^ \
+	| sed 's/<h4 name="\(.*\)#\(.*\)">\(.*\)\1#\2/<h4 name="\1.prototype.\2">\3\1#\2/' >'$@'
 
 
 .PHONY: lint
 lint:
-	$(JSHINT) -- $(SRC)
-	$(JSCS) -- $(SRC)
+	$(JSHINT) -- index.js test/index.js
+	$(JSCS) -- index.js test/index.js
 
 
 .PHONY: release-major release-minor release-patch
@@ -27,3 +37,10 @@ setup:
 test:
 	$(ISTANBUL) cover node_modules/.bin/_mocha -- --recursive
 	$(ISTANBUL) check-coverage --branches 100
+	@<index.js \
+	    sed -n "/^[ ]*\/\/\. >/{N;p;}" \
+	  | sed -e "s:^[ ]*\/\/\. ::" \
+	        -e "s:^\([^>].*\):   '\1');:" \
+	        -e "s:^> \(.*\):eq(R.toString(\1),:" \
+	  | sed "1 s:^:var eq = require('assert').strictEqual, R = require('ramda'), S = require('./');:" \
+	  | node
--- a/index.js
+++ b/index.js
@@ -8,6 +8,59 @@
    ####     ####
       #######    */
 
+//. # Sanctuary
+//.
+//. Sanctuary is a small functional programming library inspired by Haskell
+//. and PureScript. It depends on and works nicely with [Ramda][]. Sanctuary
+//. makes it possible to write safe code without null checks.
+//.
+//. In JavaScript it's trivial to introduce a possible run-time type error:
+//.
+//.     words[0].toUpperCase()
+//.
+//. If `words` is `[]` we'll get a familiar error at run-time:
+//.
+//.     TypeError: Cannot read property 'toUpperCase' of undefined
+//.
+//. Sanctuary gives us a fighting chance of avoiding such errors. We might
+//. write:
+//.
+//.     R.map(R.toUpper, S.head(words))
+//.
+//. ## Types
+//.
+//. Sanctuary uses Haskell-like type signatures to describe the types of
+//. values, including functions. `'foo'`, for example, has type `String`;
+//. `[1, 2, 3]` has type `[Number]`. The arrow (`->`) is used to express a
+//. function's type. `Math.abs`, for example, has type `Number -> Number`.
+//. That is, it takes an argument of type `Number` and returns a value of
+//. type `Number`.
+//.
+//. [`R.map`][R.map] has type `(a -> b) -> [a] -> [b]`. That is, it takes
+//. an argument of type `a -> b` and returns a value of type `[a] -> [b]`.
+//. `a` and `b` are type variables: applying `R.map` to a value of type
+//. `String -> Number` will give a value of type `[String] -> [Number]`.
+//.
+//. Sanctuary embraces types. JavaScript doesn't support algebraic data types,
+//. but these can be simulated by providing a group of constructor functions
+//. whose prototypes provide the same set of methods. A value of the Maybe
+//. type, for example, is created via the Nothing constructor or the Just
+//. constructor.
+//.
+//. It's necessary to extend Haskell's notation to describe implicit arguments
+//. to the *methods* provided by Sanctuary's types. In `x.map(y)`, for example,
+//. the `map` method takes an implicit argument `x` in addition to the explicit
+//. argument `y`. The type of the value upon which a method is invoked appears
+//. at the beginning of the signature, separated from the arguments and return
+//. value by a squiggly arrow (`~>`). The type of the `map` method of the Maybe
+//. type is written `Maybe a ~> (a -> b) -> Maybe b`. One could read this as:
+//.
+//. _When the `map` method is invoked on a value of type `Maybe a`
+//. (for any type `a`) with an argument of type `a -> b` (for any type `b`),
+//. it returns a value of type `Maybe b`._
+//.
+//. ## API
+
 ;(function() {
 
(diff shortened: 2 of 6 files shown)


svozza:
Wow. That is some amount of work! I will have a proper look at this in the morning and give some feedback afterwards.

davidchambers:
Thank you, @svozza. I look forward to your feedback. :)

Makefile
```diff
@@ -2,15 +2,25 @@ ISTANBUL = node_modules/.bin/istanbul
...
+
+.PHONY: all
+all: README.md
+
+README.md: index.js
+	$(TRANSCRIBE) \
+	  --heading-level 4 \
+	  --url 'https://github.com/plaid/sanctuary/blob/v$(VERSION)/{filename}#L{line}' \
+	  -- $^ \
+	| sed 's/<h4 name="\(.*\)#\(.*\)">\(.*\)\1#\2/<h4 name="\1.prototype.\2">\3\1#\2/' >'$@'
```
davidchambers: This replaces something like `name="Maybe#map"` with `name="Maybe.prototype.map"`. This is useful since according to [RFC 3986][1], `#` is not a valid character in a [fragment identifier][2]. I did test `https://github.com/…/README.md#Maybe#map` in Chrome, and it did work, but it seems wise to avoid unusual URLs which may not be handled correctly by any number of applications.


[1]: http://tools.ietf.org/html/rfc3986
[2]: http://tools.ietf.org/html/rfc3986#section-3.5

README.md
```diff
@@ -1,8 +1,8 @@
 # Sanctuary
 
-Sanctuary is a small functional programming library inspired by Haskell and
-PureScript. It depends on and works nicely with [Ramda][1]. Sanctuary makes
-it possible to write safe code without null checks.
+Sanctuary is a small functional programming library inspired by Haskell
+and PureScript. It depends on and works nicely with [Ramda][]. Sanctuary
```
benperez: It looks like you lost the link url here

davidchambers: Markdown—or at least, [GitHub Flavored Markdown][1]—supports `[Foo][]` as shorthand for `[Foo][Foo]`. I like to use numeric reference links in most contexts, but for long documents such as this one updating the numbers each time a link is added or removed is irksome.


[1]: https://help.github.com/articles/github-flavored-markdown/

benperez:
I'm loving all of the new documentation! Nice work!

One thing that I initially found a little confusing is that the fat arrow (`=>`) is used to indicate that a method on a type implicitly takes itself as the first argument. I agree that some special syntax is needed to communicate this concept to the user of the API. The only problem with the fat arrow is that it's already used in Haskell to express a class constraint. It's not a huge deal but I think that it might confuse people who are used to Haskell type signatures. Maybe
```haskell
Maybe.of :: Maybe a => b -> Maybe b
```
could be expressed instead as something like
```haskell
Maybe.of :: (Maybe a ->) b -> Maybe b
```
or
```haskell
Maybe.of :: (Maybe a ->)* b -> Maybe b
```

davidchambers:
Terrific feedback, @benperez! I was aware that the `=>` has a different meaning in Haskell, and I agree that it's confusing to use it here for a different purpose. I did consider using brackets of some sort, but was deterred by the fact that `[` and `]` are already used to represent lists, `{` and `}` are already used to represent dictionaries and records, and `(` and `)` are already used for grouping. Seeing your examples made me realize that parentheses really do suggest optionality, which is close to the meaning we hope to convey. Furthermore the parentheses are clearly not included for grouping, which suggests to the reader they have some other meaning.

The other option which seems reasonable is to use a different arrow such as `~>`. Our options are:

```haskell
map :: (Maybe a ->) (a -> b) -> Maybe b
```

```haskell
map :: Maybe a ~> (a -> b) -> Maybe b
```

I believe people are more likely to infer the meaning from the first option. My one hesitation is that the parentheses make the signature a bit noisy when parentheses are also required for grouping, as above.

Which of these two options do others prefer?

davidchambers:
@benperez revealed via HipChat a fondness for the squiggly arrow (`~>`). I've updated the pull request to use this, as it's clearly better than the fat arrow (`=>`). I'm open to switching to parentheses if someone makes a compelling case.

README.md
```diff
@@ -33,118 +34,583 @@ Without Sanctuary, one might have written:
...
+```
+
+<h4 name="Either.prototype.map"><code><a href="https://github.com/plaid/sanctuary/blob/v0.4.0/index.js#L487">Either#map :: Either a b ~> (b -> c) -> Either a c</a></code></h4>
+
+Takes a function and returns `this` if `this` is a Left; otherwise it
+returns a Right whose value is the result of applying the function to
+this Right's value.
+
+```javascript
+> S.Left('Cannot divide by zero')
```
svozza: Is this example correct?

README.md
```diff
@@ -33,118 +34,583 @@ Without Sanctuary, one might have written:
...
+for each object in the list. Returns the value wrapped in a Just
+if an object has the key and a Nothing if it does not.
 
-Returns a `Just` containing a list with all but the last element in the original list or a `Nothing` that list is empty.
 ```javascript
-S.init([0, 1, 2, 3, 4]) // => Just([0, 1, 2, 3])
-S.init([]) // => Nothing()
+> S.pluck('a', [{a: 1, b: 2}, {a: 4, b: 5}, {b: 3, c: 7}])
+[Just(1), Just(4), Nothing()]
 ```
```
svozza: Should we give an example where you can end up with a `Just(undefined)` value?

README.md
```diff
@@ -33,118 +34,583 @@ Without Sanctuary, one might have written:
...
+
+Takes a list of objects and plucks the value of the specified key
+for each object in the list. Returns the value wrapped in a Just
+if an object has the key and a Nothing if it does not.
 
-Returns a `Just` containing a list with all but the last element in the original list or a `Nothing` that list is empty.
 ```javascript
-S.init([0, 1, 2, 3, 4]) // => Just([0, 1, 2, 3])
-S.init([]) // => Nothing()
+> S.pluck('a', [{a: 1, b: 2}, {a: 4, b: 5}, {b: 3, c: 7}])
```
svozza: Should we give an example where you can end up with a Just(undefined) value?

README.md
davidchambers: Good catch! Updated.

README.md
```diff
@@ -33,118 +34,583 @@ Without Sanctuary, one might have written:
...
-`parseDate :: String -> Maybe Date`
+Takes a string and returns Just the date represented by the string
+if it does in fact represent a date; Nothing otherwise.
 
-Returns a `Just` containing a Javascript date object if the date is valid or a `Nothing` otherwise.
 ```javascript
-S.parseDate('2011-01-19T17:40:00Z') // => Just('Sun Jun 19 2011 19:40:00 GMT+0100 (BST)')
-S.parseDate('2015-02-31T17:40:00Z') // => Nothing()
+> S.parseDate('2001-02-03')
+Just(new Date("2001-02-03T00:00:00.000Z"))
```
svozza: Shame the date has changed there, the previous value was the exact date and time my son was born. :smile:

README.md
davidchambers: Good idea. I've added one.

README.md
davidchambers: I didn't realize the significance of that date! I've reverted it to your son's birthday. :)

svozza: Ah I was just being sentimental, appreciate you changing it back though. :)

svozza:
So I've _finally_ had a chance to read through the docs. I'm probably not the person to be giving the sort of detailed feedback @benperez has but as someone very new to this stuff I have to say the section on types is excellent, it's really clear and concise explanation of the nomenclature.

The only thing I would say is the the intro section seems a bit disjointed. It jumps straight into the types stuff when describing `S.head` even though the explanation of type signatures comes in the next section. That bit also sort of duplicates what's in the API docs in the section below.

davidchambers:
> The only thing I would say is the the intro section seems a bit disjointed.

I agree. I'll rework the section to briefly explain the motivation for creating the library without mentioning functors and the specifics of `S.head`.

davidchambers:
I've updated the introduction. Also, all occurrences of "TK" have been replaced with documentation. :tada:

svozza:
Nice!

index.js
```diff
@@ -171,10 +442,34 @@
...
+  //. applying `f` to the same arguments: if this succeeds, `g` returns Just
+  //. the result; otherwise `g` returns Nothing.
+  //.
+  //. ```javascript
+  //. > S.encase(eval)('1 + 1')
+  //. Just(2)
+  //.
+  //. > S.encase(eval)('1 +')
+  //. Nothing()
+  //. ```
```
davidchambers: @charleswli suggested encasing a function other than `JSON.parse`, since Sanctuary exports an encased version of that function. The new examples apply `eval` to syntactically valid and syntactically invalid input.

davidchambers:
Merging!

PR #37, opened 2015-05-21
Add indexOf function

svozza:
As mentioned in #35.

### DIFF
--- a/index.js
+++ b/index.js
@@ -343,6 +343,10 @@
     return xBool !== yBool ? or(x, y) : xEmpty;
   });
 
+  var sanctifyIndexOf = function(f) {
+    return R.curry(R.compose(R.ifElse(R.gte(R.__, 0), Just, Nothing), f));
+  };
+
   //  list  //////////////////////////////////////////////////////////////////
 
   //  at :: Number -> [a] -> Maybe a
@@ -373,6 +377,30 @@
     }
     return Nothing();
   });
+
+  //. indexOf :: a -> [a] -> Maybe Number
+  //.
+  //. Returns Just the position of the first occurrence of an
+  //. element in a list, or Nothing if the element does not occur
+  //. in the list.
+  //.
+  //. ```javascript
+  //. > S.indexOf('c', ['a', 'b', 'c', 'd', 'e'])
+  //. Just(2)
+  //. ```
+  S.indexOf = sanctifyIndexOf(R.indexOf);
+
+  //. lastIndexOf :: a -> [a] -> Maybe Number
+  //.
+  //. Returns Just the position of the last occurrence of an
+  //. element in a list, or Nothing if the element does not occur
+  //. in the list.
+  //.
+  //. ```javascript
+  //. > S.lastIndexOf('c', ['a', 'b', 'b', 'c', 'c'])
+  //. Just(4)
+  //. ```
+  S.lastIndexOf = sanctifyIndexOf(R.lastIndexOf);
 
   //. pluck :: String -> [{String: *}] -> [Maybe *]
   //.
--- a/test/index.js
+++ b/test/index.js
@@ -1213,6 +1213,40 @@ describe('list', function() {
 
   });
 
+  describe('indexOf', function() {
+
+    it('returns a Nothing for an empty list', function() {
+      assert(S.indexOf(10, []).equals(S.Nothing()));
+    });
+
+    it('returns a Nothing if the element is not found', function() {
+      assert(S.indexOf(10, [1, 2, 3, 4, 5]).equals(S.Nothing()));
+    });
+
+    it('returns Just the index of the element found', function() {
+      assert(S.indexOf(3, [1, 2, 3, 4, 5]).equals(S.Just(2)));
+    });
+
+  });
+
+  describe('lastIndexOf', function() {
+
+    it('returns a Nothing for an empty list', function() {
+      assert(S.lastIndexOf('a', []).equals(S.Nothing()));
+    });
+
+    it('returns a Nothing if the element is not found', function() {
+      assert(S.lastIndexOf('d', ['a', 'b', 'b', 'c', 'c'])
+        .equals(S.Nothing()));
+    });
+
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -374,6 +374,21 @@
     return Nothing();
   });
 
+  //. indexOf :: a -> [a] - Maybe Number
```
davidchambers: Your second arrow is missing its `>`.

index.js
```diff
@@ -374,6 +374,21 @@
...
+  //.
+  //. Returns Just the position of the first occurrence of an
+  //. element in a list, or Nothing if the element does not occur
+  //. in the list.
+  //.
+  //. ```javascript
+  //. > S.indexOf(3, [1, 2, 3, 4, 5])
+  //. Just(2)
```
davidchambers: When dealing with indexes it's clearer to use a list of strings in the example.

index.js
```diff
@@ -374,6 +374,21 @@
...
+  //. ```
+  S.indexOf = R.curry(function(target, list) {
+    var idx = R.indexOf(target, list);
+    return idx === -1 ? Nothing() : Just(idx);
```
davidchambers: I prefer not to specify the arbitrarily chosen failure value:

```javascript
return idx >= 0 ? Just(idx) : Nothing();
```

davidchambers:
This is great! Have you any interest in defining `S.lastIndexOf` as well?

svozza:
Yeah, definitely. Will I add it to this PR or make a new one?

davidchambers:
> Will I add it to this PR or make a new one?

That's up to you. I'm happy with either option. :)

svozza:
Nice one. Got a better name for the `indexOfToMaybe` function I've added there? Not happy with it at all.

svozza:
Did you get my Gitter message btw?

index.js
```diff
@@ -374,6 +376,36 @@
...
+  //. in the list.
+  //.
+  //. ```javascript
+  //. > S.indexOf('c', ['a','b', 'c', 'd', 'e'])
+  //. Just(2)
+  //. ```
+  S.indexOf = R.curry(function(target, list) {
+    var idx = R.indexOf(target, list);
+    return indexOfToMaybe(idx);
+  });
```
davidchambers: If we :curry: `filter` and add `var _ = R.__;` we can write:

```javascript
S.indexOf = R.curry(R.compose(filter(R.gte(_, 0)), Just, R.indexOf));
```

This was we could avoid having to come up with a good name for `indexOfToMaybe`. :)

index.js
```diff
@@ -374,6 +376,36 @@
...
   });
 
+  //. indexOf :: a -> [a] -> Maybe Number
+  //.
+  //. Returns Just the position of the first occurrence of an
+  //. element in a list, or Nothing if the element does not occur
+  //. in the list.
+  //.
+  //. ```javascript
+  //. > S.indexOf('c', ['a','b', 'c', 'd', 'e'])
```
davidchambers: We're missing a space before `'b'`.

index.js
```diff
@@ -374,6 +376,36 @@
...
+  });
+
+  //. lastIndexOf :: a -> [a] -> Maybe Number
+  //.
+  //. Returns Just the position of the last occurrence of an
+  //. element in a list, or Nothing if the element does not occur
+  //. in the list.
+  //.
+  //. ```javascript
+  //. > S.indexOf('c', ['a','b', 'b', 'c', 'c'])
```
davidchambers: We're missing a space before `'b'` here as well.

test/index.js
```diff
@@ -1213,6 +1213,40 @@ describe('list', function() {
...
   });
 
+  describe('indexOf', function() {
+
+    it('returns a Nothing for an empty list', function() {
+      assert(S.indexOf(10, []).equals(S.Nothing()));
+    });
+
+    it('returns a Nothing if the element is not found', function() {
+      assert(S.indexOf(10, [1, 2, 3, 4, 5]).equals(S.Nothing()), true);
```
davidchambers: This should either be `assert(...)` or `eq(..., true)`. :)

index.js
svozza: Nice. But I don't think it gets us of the hook, this is a pattern that's going to be repeated in the `indexOf` methods for the strings so it'll still make sense to have a function that converts any function where a -1 indicates a `null` value and reuse it across the board.

svozza:
So I wrote a wrapper function for indexOf methods, I'll add the string ones in a separate PR though.

index.js
```diff
@@ -343,6 +343,10 @@
     return xBool !== yBool ? or(x, y) : xEmpty;
   });
 
+  var wrapIndexOf = function(f) {
+    return R.curry(R.compose(R.ifElse(R.lte(0), Just, Nothing), f));
```
benperez: This might be more clear with a placeholder:
```javascript
R.curry(R.compose(R.ifElse(R.gte(R.__, 0), Just, Nothing), f));
```

svozza: Yes, but would it be more beautiful? :smile: But yeah, I agree. Actually would there be a way to make the whole function definition point free?

davidchambers: I agree that the placeholder avoids misinterpretation of a partially applied `R.{lt,gt}{,e}`.

> Actually would there be a way to make the whole function definition point free?

Yesterday, between talks at LambdaConf, I did just that, but I didn't have a reliable internet connection so didn't post it.

```javascript
//  compose :: (b -> c) -> (a -> b) -> a -> c
var compose = R.curryN(2, R.compose);

var sanctifyIndexOf = R.pipe(compose(Just),
                             compose(filter(R.gte(_, 0))),
                             R.curry);
```

This is not necessarily as good as your version of the function, but it was an enjoyable exercise.

I do like the name `sanctifyIndexOf`, though. :)

svozza: Nice. Big +1 for `sanctifyIndexOf`. :)

davidchambers:
:deciduous_tree:

Issue #39, opened 2015-05-26
Ramda String Functions 


svozza:
Working on #38 got me thinking about the other Ramda string functions that return a negative result that's not necessarily a `null` or `undefined`. For example, if you pass an index to `substringFrom` that's greater than the length of the string being operated on you just get back an empty string. Would it be useful to have a sanctified version that returns a Nothing in these sorts of cases?

davidchambers:
Before considering string slice functions, let's decide whether a sanctified version of [`R.slice`][1] is a good idea. We could implement [`S.tail`][2] and [`S.init`][3] in terms of `S.slice`:

```javascript
S.tail = slice(1, Infinity);
S.init = slice(0, -1);
```

```javascript
> S.slice(2, 4, [])
Nothing()
> S.slice(2, 4, ['a', 'b', 'c', 'd', 'e'])
Just(['c', 'd'])
```

I'm not sure how `S.slice(2, 102, ['a', 'b', 'c', 'd', 'e'])` would evaluate.


[1]: http://ramdajs.com/docs/#slice
[2]: https://github.com/plaid/sanctuary#tail
[3]: https://github.com/plaid/sanctuary#init

svozza:
I guess there's not much difference between `slice` when `toIndex > arr.length` and `tail` so I think it's justified in only returning the remaining elements of the array. I don't think it would be that useful to return what amounts to a sparse array with 96 `Nothing`s or `Just(undefined)`s.

davidchambers:
I'm wondering how the following translates to the hypothetical `S.slice`:

```javascript
> S.tail([1, 2, 3])
Just([2, 3])
> S.tail([1, 2])
Just([2])
> S.tail([1])
Just([])
> S.tail([])
Nothing()
```

It seems `S.slice` should only return a Just if the array contains the *entire* range of indexes (otherwise it isn't well differentiated from [`R.slice`][1]):

```javascript
> S.slice(1, 1, ['a', 'b', 'c'])
Just([])
> S.slice(1, 2, ['a', 'b', 'c'])
Just(['b'])
> S.slice(1, 3, ['a', 'b', 'c'])
Just(['b', 'c'])
> S.slice(1, 4, ['a', 'b', 'c'])
Nothing()
```

This would guarantee that if `startIndex >= 0` and `endIndex >= 0` and `startIndex <= endIndex`, the resulting list will contain `endIndex - startIndex` elements.


[1]: http://ramdajs.com/docs/#slice

[...]

PR #42, opened 2015-05-29
Add Slice Function

svozza:
As discussed in #39. This actually turned out to be more difficult than anticipated, I think I've managed to test all the edge cases.

### DIFF
--- a/index.js
+++ b/index.js
@@ -864,6 +864,44 @@
 
   //. ### List
 
+  //# slice :: Number -> Number -> [a] -> Maybe [a]
+  //.
+  //. Returns Just a list containing the elements from the supplied list
+  //. from a beginning index (inclusive) to an end index (exclusive).
+  //. Returns Nothing unless the start interval is less than or equal to
+  //. the end interval, and the list contains both (half-open) intervals.
+  //. Accepts negative indices, which indicate an offset from the end of
+  //. the list.
+  //.
+  //. ```javascript
+  //. > S.slice(1, 1, ['a', 'b', 'c', 'd', 'e'])
+  //. Just([])
+  //.
+  //. > S.slice(1, 3, ['a', 'b', 'c', 'd', 'e'])
+  //. Just(["b", "c"])
+  //.
+  //. > S.slice(-2, -0, ['a', 'b', 'c', 'd', 'e'])
+  //. Just(["d", "e"])
+  //.
+  //. > S.slice(2, -0, ['a', 'b', 'c', 'd', 'e'])
+  //. Just(["c", "d", "e"])
+  //.
+  //. > S.slice(0, 2, [])
+  //. Nothing()
+  //.
+  //. > S.slice(1, 6, ['a', 'b', 'c', 'd', 'e'])
+  //. Nothing()
+  //. ```
+  var slice = S.slice = R.curry(function(start, end, xs) {
+    var len = xs.length;
+    var startIdx = R.eq(-0, start) ? len : start < 0 ? start + len : start;
+    var endIdx = R.eq(-0, end) ? len : end < 0 ? end + len : end;
+
+    return (Math.abs(start) <= len && Math.abs(end) <= len && startIdx <= endIdx) ?
+      Just(R.slice(startIdx, endIdx, xs)) :
+      Nothing();
+  });
+
   //# at :: Number -> [a] -> Maybe a
   //.
   //. Takes an index and a list and returns Just the element of the list at
@@ -881,9 +919,7 @@
   //. Just("d")
   //. ```
   var at = S.at = R.curry(function(n, xs) {
-    var len = xs.length;
-    var idx = n < 0 ? len + n : n;
-    return idx >= 0 && idx < len ? Just(xs[idx]) : Nothing();
+    return R.map(R.head, slice(n, n === -1 ? -0 : n + 1, xs));
   });
 
   //# head :: [a] -> Maybe a
@@ -927,7 +963,7 @@
   //. > S.tail([])
   //. Nothing()
   //. ```
-  S.tail = R.ifElse(R.isEmpty, Nothing, R.compose(Just, R.tail));
+  S.tail = S.slice(1, -0);
 
   //# init :: [a] -> Maybe [a]
   //.
@@ -942,7 +978,7 @@
   //. > S.init([])
   //. Nothing()
   //. ```
-  S.init = R.ifElse(R.isEmpty, Nothing, R.compose(Just, R.init));
+  S.init = S.slice(0, -1);
 
   //# find :: (a -> Boolean) -> [a] -> Maybe a
   //.
--- a/test/index.js
+++ b/test/index.js
@@ -1107,12 +1107,65 @@ describe('list', function() {
     });
 
     it('returns a Nothing if index out of bounds', function() {
+      assert(S.at(-0, ['foo', 'bar', 'baz']).equals(S.Nothing()));
       assert(S.at(3, ['foo', 'bar', 'baz']).equals(S.Nothing()));
       assert(S.at(-4, ['foo', 'bar', 'baz']).equals(S.Nothing()));
     });
 
     it('is curried', function() {
       assert(S.at(1)(['foo', 'bar', 'baz']).equals(S.Just('bar')));
+    });
+
+  });
+
+  describe('slice', function() {
+
+    it('returns a Nothing with a positive end index greater than start index', function() {
+      assert(S.slice(6, 1, [1, 2, 3, 4, 5]).equals(S.Nothing()));
+    });
+
+    it('returns a Nothing with a positive end index greater than list length', function() {
+      assert(S.slice(1, 6, [1, 2, 3, 4, 5]).equals(S.Nothing()));
+    });
+
+    it('returns a Nothing with a negative end index greater than list length', function() {
+      assert(S.slice(1, -6, [1, 2, 3, 4, 5]).equals(S.Nothing()));
+    });
+
+    it('returns a Nothing with a negative start index greater than list length', function() {
+      assert(S.slice(-6, 1, [1, 2, 3, 4, 5]).equals(S.Nothing()));
+    });
+
+    it('returns a Just with an empty array when start index equals end index', function() {
+      eq(S.slice(1, 1, [1, 2, 3, 4, 5]).toString(), 'Just([])');
+      eq(S.slice(1, -4, [1, 2, 3, 4, 5]).toString(), 'Just([])');
+      eq(S.slice(-4, 1, [1, 2, 3, 4, 5]).toString(), 'Just([])');
+      eq(S.slice(-4, -4, [1, 2, 3, 4, 5]).toString(), 'Just([])');
+      eq(S.slice(0, 0, []).toString(), 'Just([])');
+      eq(S.slice(0, -0, []).toString(), 'Just([])');
+      eq(S.slice(-0, 0, []).toString(), 'Just([])');
+      eq(S.slice(-0, -0, []).toString(), 'Just([])');
+    });
+
+    it('returns a Just with a positive start and end index', function() {
+      eq(S.slice(1, 3, [1, 2, 3, 4, 5]).toString(), 'Just([2, 3])');
+    });
+
+    it('returns a Just with a negative start index', function() {
+      eq(S.slice(-3, 5, [1, 2, 3, 4, 5]).toString(), 'Just([3, 4, 5])');
+    });
+
+    it('returns a Just with a negative end index', function() {
+      eq(S.slice(1, -2, [1, 2, 3, 4, 5]).toString(), 'Just([2, 3])');
+    });
+
+    it('accepts -0 as the position half a step beyond the last index', function() {
+      eq(S.slice(-0, 5, [1, 2, 3, 4, 5]).toString(), 'Just([])');
+      eq(S.slice(2, -0, [1, 2, 3, 4, 5]).toString(), 'Just([3, 4, 5])');
+    });
+
+    it('returns a Just with the whole list', function() {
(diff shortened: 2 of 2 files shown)


davidchambers:
Thanks! I'm away from home this weekend, and don't like to review code on my phone. I'll have a proper look on Monday. :)

svozza:
Actually, probably for the best you're not looking at it now, just realised I completely forgot to add documentation!

svozza:
Just thinking, do we want to do a check with `R.nil` as well, give that this sort of thing gives a result:

```javascript
R.slice(null, 2, [1, 2, 3, 4, 5])
// => [1, 2]

R.slice(void 0, 2, [1, 2, 3, 4, 5])
// => [1, 2]
```

davidchambers:
I'd prefer to leave the behaviour of non-number index values unspecified.

index.js
```diff
@@ -942,7 +977,7 @@
   //. > S.init([])
   //. Nothing()
   //. ```
-  S.init = R.ifElse(R.isEmpty, Nothing, R.compose(Just, R.init));
+  S.init = S.slice(0, -1);
```
davidchambers: Let's accept `-0` to indicate the position half a step beyond the last index. Currently we have:

```javascript
 [ 'a' , 'b' , 'c' , 'd' , 'e' ]

 0     1     2     3     4     5
-5    -4    -3    -2    -1
```

Instead we could have:

```javascript
 [ 'a' , 'b' , 'c' , 'd' , 'e' ]

 0     1     2     3     4     5
-5    -4    -3    -2    -1    -0
```

This would facilitate this symmetry:

```javascript
S.tail = S.slice(1, -0);
S.init = S.slice(0, -1);
```

With `R.slice` one can use `Infinity` as the end index, but that does not work when the index needs to be within the array's bounds.

The most straightforward expression to detect `-0` is `R.toString(x) === '-0'`.

index.js
```diff
@@ -886,6 +886,41 @@
     return idx >= 0 && idx < len ? Just(xs[idx]) : Nothing();
   });
 
+  //# slice :: Number -> Number -> [a] -> Maybe [a]
+  //.
+  //. Returns Just a list containing the elements from the supplied list
+  //. from a beginning index (inclusive) to an end index (exclusive).
+  //. Returns Nothing if the range specified by the two indices exceeds
+  //. length of the list. Accepts negative indices, which indicate an
```
davidchambers: It's not accurate to write _exceeds length of the list_. Perhaps we could write _Returns Nothing unless the start interval is less than or equal to the end interval, and the list contains both (half-open) intervals._

index.js
```diff
@@ -886,6 +886,41 @@
...
+  //.
+  //. > S.slice(2, -1, [1, 2, 3, 4, 5])
+  //. Just([3, 4])
+  //.
+  //. > S.slice(0, 2, [])
+  //. Nothing()
+  //.
+  //. > S.slice(1, 6, [1, 2, 3, 4, 5])
+  //. Nothing()
+  //. ```
```
davidchambers: Let's use `['a', 'b', 'c', 'd', 'e']` as the example list.

index.js
CrossEye: That's really elegant!

I've not really looked, but I haven't seen many truly useful reasons to have `-0` except that the number implementation demanded it.  This shows that there really can be such.  Nice!

svozza: This also allows us to define `S.at` in terms on `S.slice` too.

davidchambers: > This also allows us to define `S.at` in terms on `S.slice` too.

I don't see it, but I'll wait to see your implementation. :)

The important thing is that `S.at(-0, xs)` should return a Nothing for every `xs`, since in this context `-0` refers to the position *after* the last index.

svozza: The implementation is already in there. :) (Although I had forgotten to add a test for the -0 functionality in the `S.at` tests.)

svozza: Also, I started a points free implementation but gave up  as soon as I came up with this abomination:

```javascript
var at = S.at = R.curry(function(n, xs) {
  R.compose(R.map(R.head), S.slice(n, R.__, xs), R.ifElse(R.compose(R.eq(0), R.inc), R.always(-0), R.inc))(n);
});
```

:smile:

index.js
```diff
@@ -881,9 +881,47 @@
   //. Just("d")
   //. ```
   var at = S.at = R.curry(function(n, xs) {
+    var endIndex = R.eq(0, n + 1) ? -0 : n + 1;
```
davidchambers: Could `R.eq(0, n + 1)` be written as `n === -1`?

index.js
davidchambers: Yikes!

index.js
svozza: It could indeed, I was trying to be too clever there.

index.js
svozza: I guess it's not as bad in the `R.compose(R.eq(0), R.inc)` is replaced with R.eq(-1).

index.js
```diff
@@ -881,9 +881,47 @@
   //. Just("d")
   //. ```
   var at = S.at = R.curry(function(n, xs) {
+    var endIndex = n === -1 ? -0 : n + 1;
+    return R.compose(R.map(R.head), S.slice(n, endIndex))(xs);
```
davidchambers: When we use a Sanctuary function internally it's nice to use a local alias (`slice` in this case). Also, I don't see the value of `R.compose` here: since we're applying the composition immediately, we might as well nest function calls.

```javascript
return R.map(R.head, slice(n, n === -1 ? -0 : n + 1, xs));
```

test/index.js
```diff
@@ -1106,6 +1106,10 @@ describe('list', function() {
       assert(S.at(-1, ['foo', 'bar', 'baz']).equals(S.Just('baz')));
     });
 
+    it('returns Nothing when passed the index -0', function() {
+      assert(S.at(-0, ['foo', 'bar', 'baz']).equals(S.Nothing()));
+    });
```
davidchambers: I suggest we move this assertion to the following block, since `-0` is an out-of-bounds index.

index.js
svozza: Yeah, you're right the `compose` is completely unnecessary. Force of habit!

test/index.js
svozza: Makes sense.

index.js
```diff
@@ -864,6 +864,45 @@
...
+  //. ```
+  var slice = S.slice = R.curry(function(start, end, xs) {
+    var len = xs.length;
+    var startIdx = R.eq(-0, start) ? len : start < 0 ? start + len : start;
+    var endIdx = R.eq(-0, end) ? len : end < 0 ? end + len : end;
+
+    if (Math.abs(start) > len || Math.abs(end) > len || startIdx > endIdx) {
+      return Nothing();
+    }
+    return Just(R.slice(startIdx, endIdx, xs));
```
davidchambers: I like to avoid using `return` for control flow where possible. Let's write something like this instead:

```javascript
return (Math.abs(start) <= len &&
        Math.abs(end) <= len &&
        startIdx <= endIdx) ?
  Just(R.slice(startIdx, endIdx, xs)) :
  Nothing();
```

I'm sorry about the nitpicking!

test/index.js
```diff
@@ -1117,6 +1118,60 @@ describe('list', function() {
 
   });
 
+  describe('slice', function() {
+
+    it('returns a Nothing if applied to an empty list', function() {
+      assert(S.slice(0, 3, []).equals(S.Nothing()));
+    });
```
davidchambers: This is misleading since `S.slice(0, 0, [])` evaluates to `Just([])`. I think we could remove this assertion.

index.js
svozza: No bother. Is there a particular reason to avoid that pattern? Genuinely interested to know because I do it a lot.

test/index.js
```diff
@@ -1117,6 +1118,60 @@ describe('list', function() {
...
+    it('returns a Nothing with a negative end index greater than list length', function() {
+        assert(S.slice(1, -6, [1, 2, 3, 4, 5]).equals(S.Nothing()));
+      });
+
+    it('returns a Nothing with a negative start index greater than list length', function() {
+      assert(S.slice(-6, 1, [1, 2, 3, 4, 5]).equals(S.Nothing()));
+    });
+
+    it('returns a Just with an empty array when start index equals end index', function() {
+        eq(S.slice(1, 1, [1, 2, 3, 4, 5]).toString(), 'Just([])');
```
davidchambers: This could be expanded to eight assertions:

```javascript
eq(S.slice(1, 1, [1, 2, 3, 4, 5]).toString(), 'Just([])');
eq(S.slice(1, -4, [1, 2, 3, 4, 5]).toString(), 'Just([])');
eq(S.slice(-4, 1, [1, 2, 3, 4, 5]).toString(), 'Just([])');
eq(S.slice(-4, -4, [1, 2, 3, 4, 5]).toString(), 'Just([])');
eq(S.slice(0, 0, []).toString(), 'Just([])');
eq(S.slice(0, -0, []).toString(), 'Just([])');
eq(S.slice(-0, 0, []).toString(), 'Just([])');
eq(S.slice(-0, -0, []).toString(), 'Just([])');
```

test/index.js
```diff
@@ -1117,6 +1118,60 @@ describe('list', function() {
...
 
+  describe('slice', function() {
+
+    it('returns a Nothing if applied to an empty list', function() {
+      assert(S.slice(0, 3, []).equals(S.Nothing()));
+    });
+
+    it('returns a Nothing with a positive end index greater than start index', function() {
+        assert(S.slice(6, 1, [1, 2, 3, 4, 5]).equals(S.Nothing()));
+      });
```
davidchambers: :triangular_ruler: The indentation is a bit off here.

svozza: Fixed.

test/index.js
svozza: Good call!

index.js
davidchambers: Returning early isn't bad or wrong, it's simply an imperative construct. My preference is to avoid imperative constructs, though in some cases they are very useful.

In this particular case we're not returning early to avoid unnecessary work, we're simply *branching*. I believe the indentation should match the logical branching, so I advocate

```javascript
if (x) {
  return y;
} else {
  return z;
}
```

rather than

```javascript
if (x) {
  return y;
}
return z;
```

`if`/`else` is another imperative construct: it's not an expression, so it's only useful for side effects. So I like `return x ? y : z;` best of all.

test/index.js
```diff
@@ -1117,6 +1118,63 @@ describe('list', function() {
...
+    });
+
+    it('returns a Just with a negative end index', function() {
+      eq(S.slice(1, -2, [1, 2, 3, 4, 5]).toString(), 'Just([2, 3])');
+    });
+
+    it('accepts -0 for a start index as the position half a step beyond the last index',
+      function() {
+        eq(S.slice(2, -0, [1, 2, 3, 4, 5]).toString(), 'Just([3, 4, 5])');
+    });
```
davidchambers: Should the description say _end index_ rather than _start index_?

index.js
svozza: Interesting, I've not thought of if/else in those terms even though it's blindingly obviously when you explain like you have there.

test/index.js
svozza: It should indeed, I mixed the body of the two tests up there. I've actually rolled them into one test block now.

index.js
davidchambers: It was only recently that I read or heard it put this way. It was an aha moment for me too. :)

test/index.js
```diff
@@ -1117,6 +1118,59 @@ describe('list', function() {
...
+    it('returns a Just with a negative start index', function() {
+      eq(S.slice(-3, 5, [1, 2, 3, 4, 5]).toString(), 'Just([3, 4, 5])');
+    });
+
+    it('returns a Just with a negative end index', function() {
+      eq(S.slice(1, -2, [1, 2, 3, 4, 5]).toString(), 'Just([2, 3])');
+    });
+
+    it('accepts -0 as the position half a step beyond the last index',
+      function() {
```
davidchambers: Let's remove this line wrapping.

svozza: Yeah, I didn't like it either, I only had it that way because it was previously over 79 chars.

davidchambers:
:deciduous_tree: Thanks for persevering, @svozza. :)

svozza:
Well, there were 47 comments on my last PR so I'm taking this as an improvement! My next goal is to get it under 20. :)

index.js
svozza: Okay, so for a laugh I perserved with this. :smiley: 

```javascript
  var at = S.at = R.curry(R.compose(R.map(R.head), R.useWith(R.uncurryN(2, R.converge(S.slice, R.identity, R.ifElse(R.eq(-1), R.always(-0), R.inc))), R.identity, R.identity)));
```

svozza: This might be somewhat easier to read(!):

```javascript
var at = S.at = R.curry(R.pipe(
    R.useWith(
      R.uncurryN(2, R.converge(
        S.slice,
        R.identity,
        R.ifElse(R.eq(-1), R.always(-0), R.inc))
      ),
      R.identity,
      R.identity),
    R.map(R.head)))
```

davidchambers: That's quite something! Here it is reformatted:

```javascript
var at = S.at =
R.curry(R.compose(R.map(R.head),
                  R.useWith(R.uncurryN(2,
                                       R.converge(S.slice,
                                                  R.identity,
                                                  R.ifElse(R.eq(-1),
                                                           R.always(-0),
                                                           R.inc))),
                            R.identity,
                            R.identity)));
```

Sometimes a lambda is just the ticket. :)

svozza: I think what this really proves is that I have _far_ too much time on my hands.

svozza: Actually, funnily enough, the thing I wasted most time on (aside from the whole enterprise, that is) was trying to get all the tests running only to finally realise that it was because `compose` is not autocurried.

PR #43, opened 2015-06-02
test: compare string representations in eq

davidchambers:
This pull request improves our assertions in two ways:

  - By using value-based equality we can now write

    ```javascript
    eq(S.parseJson('["foo","bar"]'), S.Just(['foo', 'bar']));
    ```

    rather than

    ```javascript
    var xs = S.fromMaybe(null, S.parseJson('["foo","bar"]'));
    eq(Object.prototype.toString.call(xs), '[object Array]');
    eq(xs.length, 2);
    eq(xs[0], 'foo');
    eq(xs[1], 'bar');
    ```

  - By comparing string representations we get a useful error message when an assertion fails. For example,

    ```javascript
    eq(S.Just(['1', '2', '3']), S.Just([1, 2, 3]));
    // AssertionError: 'Just(["1", "2", "3"])' === 'Just([1, 2, 3])'
    ```

    rather than

    ```javascript
    assert(S.Just(['1', '2', '3']).equals(S.Just([1, 2, 3])));
    // AssertionError: false == true
    ```

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -9,7 +9,9 @@ var R = require('ramda');
 var S = require('..');
 
 
-var eq = assert.strictEqual;
+var eq = function(actual, expected) {
+  assert.strictEqual(R.toString(actual), R.toString(expected));
+};
 
 //  messageEq :: a -> Object -> Boolean
 var messageEq = R.propEq('message');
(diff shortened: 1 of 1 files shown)


svozza:
Nice. This removes a lot of unnecessary boilerplate from the tests.

PR #44, opened 2015-06-03
Add Safe Division Function

### LINKED ISSUE #41: safe division function
[`R.divide`][1] isn't actually unsafe, but `Infinity` is as unhelpful as `NaN` in many situations.

```haskell
div :: Number -> Number -> Maybe Number
div a 0 = Nothing
div a b = Just (a / b)
```

[1]: http://ramdajs.com/docs/#divide

svozza:
Closes #41

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -1033,6 +1033,20 @@ describe('control', function() {
 
 });
 
+describe('number', function() {
+
+  it('returns a Just when the divisor is not 0', function() {
+    eq(S.divide(20, 5), S.Just(4));
+    eq(S.divide(21, 5), S.Just(4.2));
+  });
+
+  it('returns Nothing when the divisor is 0', function() {
+    eq(S.divide(20, 0), S.Nothing());
+    eq(S.divide(20, -0), S.Nothing());
+  });
+
+});
+
 describe('list', function() {
 
   describe('at', function() {
--- a/index.js
+++ b/index.js
@@ -862,6 +862,26 @@
     return xBool !== yBool ? or(x, y) : xEmpty;
   });
 
+  //. ### Number
+  //. divide :: Number -> Number -> Maybe Number
+  //.
+  //. Divides two numbers, returns Nothing if the divisor is zero; Just
+  //. the result otherwise.
+  //.
+  //. ```javascript
+  //. > S.divide(20, 5)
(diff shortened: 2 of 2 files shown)


davidchambers:
Shall we name the function `divide` to match Ramda, or `div` to match Haskell?

svozza:
Yeah, I wasn't sure about that. Initially I was going to do something point free with `R.ifElse`and `R.divide`so that's why I named it `divide`. However, the point free version was  was complete overkill so I just went with the simple opt ion but didn't revisit the name. I'm thinking we may as well stick with the Ramda name though, it'll probably be less surprising to users.

davidchambers:
`S.divide` sounds good to me.

test/index.js
```diff
@@ -1033,6 +1033,18 @@ describe('control', function() {
...
 
+describe('number', function() {
+
+  it('returns a Just when the divisor is not 0', function() {
+    eq(S.divide(20, 5), S.Just(4));
+  });
+
+  it('returns Nothing when the divisor is 0', function() {
+    eq(S.divide(20, 0), S.Nothing());
+  });
```
davidchambers: Let's include a test case for division by the *other* zero. ;)

test/index.js
```diff
@@ -1033,6 +1033,18 @@ describe('control', function() {
 
 });
 
+describe('number', function() {
+
+  it('returns a Just when the divisor is not 0', function() {
+    eq(S.divide(20, 5), S.Just(4));
+  });
```
davidchambers: It's probably worth including a test case which demonstrates that the function does not do integer division.

svozza: Cool. I've added some examples to the docs to reflect this as well.

test/index.js
svozza: Haha. Done.

davidchambers:
This looks good. I'd like to leave this pull request open, though, until I spend some time working on #45. That way we can consider all the mathematical operators in the context of both Number and Int before committing to anything.

svozza:
Yep, sounds like a good plan.

davidchambers:
Closing for the reason given in https://github.com/plaid/sanctuary/issues/41#issuecomment-165291887.

PR #47, opened 2015-06-06
use aliases when referencing Sanctuary functions internally

davidchambers:


svozza:
Is this purely aesthetic or is there something else to it such as minimising look ups on the main Sanctuary object? You normally have good reasons for this sort of stuff and I'm genuinely interested to know what they are.

davidchambers:
Arguments have been made for avoiding internal references to the exported object so things (mostly) continue to work if a user modifies the module's properties! For example, if `S.bar` were defined in terms of `S.foo`, `S.bar` might break if a user were to assign a new value to `S.foo`. As far as I'm concerned all bets are off if a user modifies the module, so I don't find this argument compelling.

So I do consider this a matter of aesthetics. It would be equally valid to have no aliases and to use `S.` consistently. But I particularly like writing `Just` and `Nothing` (rather than `S.Just` and `S.Nothing`), so I prefer to use local aliases within the module.

CrossEye:
I've got to say that I see some irony here!

Is this the same man who managed to convince the entire community of Ramda contributors to forgo aliases?   :smile:

Not an objection, of course.   I also particularly like writing `Just` and `Nothing` (and `map` and `curry`.)

davidchambers:
The irony is not lost on me. :stuck_out_tongue_winking_eye:

PR #48, opened 2015-06-08
add drop and take functions

svozza:
As discussed in #39 . There's a few parts of this PR that need clarifying. I'm not sure if I've got the type signature right and I didn't know whether to group the string and array tests separately. Also, should there be more tests and documentation around passing negative values to the function?

### DIFF
--- a/index.js
+++ b/index.js
@@ -980,6 +980,48 @@
   //. ```
   S.init = S.slice(0, -1);
 
+  //# take :: Number -> [a] -> Maybe [a]
+  //.
+  //. Returns Just the first N elements of the given collection if N is
+  //. greater than or equal to zero and less than or equal to the length
+  //. of the collection; Nothing otherwise. Supports Array, String, and
+  //. any other collection type which provides a `slice` method.
+  //.
+  //. ```javascript
+  //. > S.take(2, ['a', 'b', 'c', 'd', 'e'])
+  //. Just(["a", "b"])
+  //.
+  //. > S.take(4, 'abcdefg')
+  //. Just("abcd")
+  //.
+  //. > S.take(4, ['a', 'b', 'c'])
+  //. Nothing()
+  //. ```
+  S.take = R.curry(function(n, xs) {
+    return n < 0 || R.eq(n, -0) ? Nothing() : slice(0, n, xs);
+  });
+
+  //# drop :: Number -> [a] -> Maybe [a]
+  //.
+  //. Returns Just all but the first N elements of the given collection
+  //. if N is greater than or equal to zero and less than or equal to the
+  //. length of the collection; Nothing otherwise. Supports Array, String,
+  //. and any other collection type which provides a `slice` method.
+  //.
+  //. ```javascript
+  //. > S.drop(2, ['a', 'b', 'c', 'd', 'e'])
+  //. Just(["c", "d", "e"])
+  //.
+  //. > S.drop(4, 'abcdefg')
+  //. Just("efg")
+  //.
+  //. > S.drop(4, 'abc')
+  //. Nothing()
+  //. ```
+  S.drop = R.curry(function(n, xs) {
+    return n < 0 || R.eq(n, -0) ? Nothing() : slice(n, -0, xs);
+  });
+
   //# find :: (a -> Boolean) -> [a] -> Maybe a
   //.
   //. Takes a predicate and a list and returns Just the leftmost element of
--- a/test/index.js
+++ b/test/index.js
@@ -1182,6 +1182,68 @@ describe('list', function() {
 
   });
 
+  describe('take', function() {
+
+    it('returns a Nothing if n is greater than collection length', function() {
+      eq(S.take(6, ['a', 'b', 'c', 'd', 'e']), S.Nothing());
+      eq(S.take(6, 'abcde'), S.Nothing());
+    });
+
+    it('returns a Nothing if n is negative', function() {
+      eq(S.take(-0, ['a', 'b', 'c', 'd', 'e']), S.Nothing());
+      eq(S.take(-1, ['a', 'b', 'c', 'd', 'e']), S.Nothing());
+      eq(S.take(-0, 'abcdefg'), S.Nothing());
+      eq(S.take(-1, 'abcde'), S.Nothing());
+    });
+
+    it('returns an empty collection if n is 0', function() {
+      eq(S.take(0, ['a', 'b', 'c', 'd', 'e']), S.Just([]));
+      eq(S.take(0, 'abcde'), S.Just(''));
+    });
+
+    it('returns Just the first two elements from the collection', function() {
+      eq(S.take(2, ['a', 'b', 'c', 'd', 'e']), S.Just(['a', 'b']));
+      eq(S.take(2, 'abcdefg'), S.Just('ab'));
+    });
+
+    it('returns Just the whole collection if n is equal to array length', function() {
+      eq(S.take(5, ['a', 'b', 'c', 'd', 'e']), S.Just(['a', 'b', 'c', 'd', 'e']));
+      eq(S.take(7, 'abcdefg'), S.Just('abcdefg'));
+    });
+
+  });
+
+  describe('drop', function() {
+
+    it('returns a Nothing if n is greater than collection length', function() {
+      eq(S.drop(6, ['a', 'b', 'c', 'd', 'e']), S.Nothing());
+      eq(S.drop(6, 'abcde'), S.Nothing());
+    });
+
+    it('returns a Nothing if n is negative', function() {
+      eq(S.drop(-3, ['a', 'b', 'c', 'd', 'e']), S.Nothing());
+      eq(S.drop(-0, ['a', 'b', 'c', 'd', 'e']), S.Nothing());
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -980,6 +980,54 @@
   //. ```
   S.init = S.slice(0, -1);
 
+  //# take :: Number -> [a] | String -> Maybe [a] | Maybe String
```
davidchambers: We could pretend that JavaScript has a Char type and that String is sugar for [Char]. Then we could write:

```javascript
//# take :: Number -> [a] -> Maybe [a]
```

If one substitutes `Char` for `a` then `String` for `[Char]` one gets `Number -> String -> Maybe String`.

index.js
```diff
@@ -980,6 +980,54 @@
...
+  //. Just("abcd")
+  //.
+  //. > S.take(2, [])
+  //. Nothing()
+  //.
+  //. > S.take(2, '')
+  //. Nothing()
+  //.
+  //. > S.take(4, ['a', 'b', 'c'])
+  //. Nothing()
```
davidchambers: This example is sufficient to demonstrate the failure case. Let's remove the `[]` and `''` examples in the interest of brevity.

index.js
svozza: Yeah, I was thinking that but wasn't sure whether the plan was to go down that route. Should we mention it anywhere in the docs that that is how we treat strings?

davidchambers: > Should we mention it anywhere in the docs that that is how we treat strings?

The thing is we don't support strings everywhere we support arrays. `S.slice` supports both types because `slice` is defined for both types. `S.take` and `S.drop` can support both since they can be defined in terms of `slice`.

The approach I favour is:

  - using a clear signature (even if it doesn't tell the full story);
  - stating in the docstring that strings are supported (as you have done); and
  - including an example of applying `S.take(n)` to a string (as you have done).

index.js
```diff
@@ -980,6 +980,40 @@
...
+  //. > S.drop(2, ['a', 'b', 'c', 'd', 'e'])
+  //. Just(["c", "d", "e"])
+  //.
+  //. > S.drop(4, 'abcdefg')
+  //. Just("efg")
+  //.
+  //. > S.drop(4, 'abc')
+  //. Nothing()
+  //. ```
+  S.drop = slice(R.__, -0);
```
davidchambers: Let's use the `_` alias here. :)

index.js
```diff
@@ -980,6 +980,42 @@
   //. ```
   S.init = S.slice(0, -1);
 
+  //# take :: Number -> [a] -> Maybe [a]
+  //.
+  //. Returns Just a list containing the first n elements of the given
+  //. list or Just the first n characters of a given string. If n exceeds
+  //. the length of the list or string, returns Nothing.
```
davidchambers: Perhaps we could use the word "collection":

_Returns Just the first N elements of the given collection if N is less than or equal to the length of the collection; Nothing otherwise. Supports Array, String, and any other collection type which provides a `slice` method._

index.js
svozza: Yup, the implication that all array functions support strings was something I was wary of; I know there was mention of it in `indexOf` PR on Ramda too.

davidchambers:
> Also, should there be more tests and documentation around passing negative values to the function?

I don't think we should document the handling of negative counts, as they're nonsensical. Some tests would be useful, though. How should we handle a negative count? Treat it as zero?

svozza:
It should probably be treated the same way we treat a count that exceeds the length of the array and return `Nothing`.

index.js
svozza: If we mention `slice` we should mention that negative indexes are not supported: even though they are nonsensical you can guarantee someone will try to use it at some point.

davidchambers:
That seems reasonable to me. Let's do that.

index.js
davidchambers: Okay. How about this?

_Returns Just the first N elements of the given collection if N is **greater than or equal to zero and** less than or equal to the length of the collection; Nothing otherwise. Supports Array, String, and any other collection type which provides a `slice` method._

svozza: Ah yes, that read nicer than my version:

_Returns Just all but the first N elements of the given collection if N is less than or equal to the length of the collection; Nothing if N is less than zero or greater than the length of the collection. Supports Array, String, and any other collection type which provides a slice method._

index.js
```diff
@@ -980,6 +980,48 @@
   //. ```
   S.init = S.slice(0, -1);
 
+  //# take :: Number -> [a] -> Maybe [a]
+  //.
+  //. Returns Just the first N elements of the given collection if N is
+  //. greater than or equal to zero and less than or equal to the length
+  //. of the collection; Nothing otherwise. Supports Array, String, and
+  //. any other collection type which provides a slice method.
```
davidchambers: It would be nice to wrap "slice" in backticks to it renders as "`slice`" rather than "slice".

svozza: Cool. What about Array and String?

index.js
```diff
@@ -980,6 +980,48 @@
...
+  S.take = R.curry(function(n, xs) {
+    return n < 0 ? Nothing() : slice(0, n, xs);
+  });
+
+  //# drop :: Number -> [a] -> Maybe [a]
+  //.
+  //. Returns Just all but the first N elements of the given collection if N is
+  //. greater than or equal to zero and less than or equal to the length of the
+  //. collection; Nothing otherwise. Supports Array, String, and any other
+  //. collection type which provides a slice method.
```
davidchambers: This is *very* minor, but we could tweak the line wrapping to increase the aesthetic appeal of this paragraph:

```javascript
  //. Returns Just all but the first N elements of the given collection
  //. if N is greater than or equal to zero and less than or equal to the
  //. length of the collection; Nothing otherwise. Supports Array, String,
  //. and any other collection type which provides a `slice` method.
```

One day I'll write a function to describe the attractiveness of paragraphs in text files. :)

svozza: Haha. It would actually be really cool if there was a way to justify text using markdown.

index.js
davidchambers: > What about Array and String?

I like them unformatted, to match Just and Nothing. I use backticks with restraint because they add noise to the raw version and can even be distracting in the rendered version if lots of things are wrapped in little grey boxes. I think the method name should stand out here, though.

test/index.js
```diff
@@ -1182,6 +1182,78 @@ describe('list', function() {
 
   });
 
+  describe('take', function() {
+
+    it('returns a Nothing if applied to empty collection', function() {
```
davidchambers: This isn't quite accurate since I assume `S.take(0, [])` evaluates to `Just([])`. I think we could remove this section since the `n > xs.length` case is tested in the following section.

test/index.js
```diff
@@ -1182,6 +1182,78 @@ describe('list', function() {
...
+
+    it('returns a Nothing if n is greater than collection length', function() {
+      eq(S.take(6, ['a', 'b', 'c', 'd', 'e']), S.Nothing());
+      eq(S.take(6, 'abcde'), S.Nothing());
+    });
+
+    it('returns a Nothing if n is less than 0', function() {
+      eq(S.take(-1, ['a', 'b', 'c', 'd', 'e']), S.Nothing());
+      eq(S.take(-1, 'abcde'), S.Nothing());
+    });
```
davidchambers: What about `-0`? I would expect this to always result in a Nothing, since we differentiate between positive zero and negative zero in `S.slice`.

test/index.js
```diff
@@ -1182,6 +1182,68 @@ describe('list', function() {
...
+      eq(S.take(0, 'abcde'), S.Just(''));
+    });
+
+    it('returns Just the first two elements from the collection', function() {
+      eq(S.take(2, ['a', 'b', 'c', 'd', 'e']), S.Just(['a', 'b']));
+      eq(S.take(2, 'abcdefg'), S.Just('ab'));
+    });
+
+    it('returns Just the whole collection if n is equal to array length', function() {
+      eq(S.take(-0, ['a', 'b', 'c', 'd', 'e']), S.Just(['a', 'b', 'c', 'd', 'e']));
```
davidchambers: This doesn't seem correct to me. Either negative zero should be treated as equivalent to positive zero, in which case this should give `S.Just([])`, or it should be treated as invalid, in which case this should give `Nothing()`. I would prefer to treat `-0` as invalid along with all other negative numbers, since we differentiate `0` and `-0` in `S.slice`.

svozza: Yeah, I wrote those tests before we decided to make negative numbers return Nothing. Fixed.

test/index.js
```diff
@@ -1182,6 +1182,68 @@ describe('list', function() {
...
   });
 
+  describe('take', function() {
+
+    it('returns a Nothing if n is greater than collection length', function() {
+      eq(S.take(6, ['a', 'b', 'c', 'd', 'e']), S.Nothing());
+      eq(S.take(6, 'abcde'), S.Nothing());
+    });
+
+    it('returns a Nothing if n is less than 0', function() {
```
davidchambers: It would be more accurate to state _if n is negative_.

davidchambers:
:deciduous_tree: Thanks, @svozza!

PR #49, opened 2015-06-11
test: compare Error objects by constructor

davidchambers:
The tests should assert exactly which constructor was used to create a particular Error object. `R.is(Error, error)` evaluates true even if `error` has a more specific type such as TypeError.


benperez:
:dromedary_camel:

PR #50, opened 2015-06-12
systematic type checking

davidchambers:
Sanctuary currently does type checking in a few places (as those of use who've seen `Pattern match failure` logged to our console can attest). This pull request addresses two problems with the status quo:

  - when a TypeError is thrown, the error message is rarely helpful; and
  - the site at which a TypeError is thrown is often one or more steps removed from the type mismatch.

See what happens when we provide arguments to a function in the wrong order:

```javascript
S.at(['foo', 'bar', 'baz'], 1);
// TypeError: at requires a value of type Number as its first argument; received ["foo", "bar", "baz"]
```

Types are checked as arguments are applied:

```javascript
S.at('1');
// TypeError: at requires a value of type Number as its first argument; received "1"
```

This approach was proposed by @paldepind in https://github.com/ramda/ramda/issues/1125#issuecomment-108999050.

It's pleasing that __index.js__ did not grow significantly, despite the fact that we now :curry: our own functions.

### DIFF
--- a/index.js
+++ b/index.js
@@ -79,6 +79,58 @@
 
   var _ = R.__;
 
+  //  placeholder :: a -> Boolean
+  var placeholder = function(x) {
+    return x != null && x['@@functional/placeholder'] === true;
+  };
+
+  /* istanbul ignore next */
+  var Any = function Any() {};
+
+  var arity = function(n, f) {
+    switch (n) {
+      case 0: return function() { return f.apply(this, arguments); };
+      case 1: return function(a) { return f.apply(this, arguments); };
+      case 2: return function(a, b) { return f.apply(this, arguments); };
+      case 3: return function(a, b, c) { return f.apply(this, arguments); };
+    }
+  };
+
+  var curry = function(name, typePairs, f) {
+    return arity(typePairs.length, function() {
+      var args = arguments;
+      var $typePairs = [];
+      for (var idx = 0; idx < args.length; idx += 1) {
+        var typePair = typePairs[idx];
+        if (placeholder(args[idx])) {
+          $typePairs.push(typePair);
+        } else if (typePair[1] !== Any && !R.is(typePair[1], args[idx])) {
+          throw new TypeError(R.join(' ', [
+            name,
+            'requires a value of type',
+            /^function (\w*)/.exec(typePair[1])[1],
+            'as its',
+            ['first', 'second', 'third'][typePair[0]],
+            'argument; received',
+            R.toString(args[idx])
+          ]));
+        }
+      }
+      $typePairs = R.concat($typePairs, typePairs.slice(args.length));
+      return $typePairs.length === 0 ?
+        f.apply(this, args) :
+        curry(name, $typePairs, function() {
+          var $args = Array.prototype.slice.call(arguments);
+          var val = function(x) { return placeholder(x) ? $args.shift() : x; };
+          return f.apply(this, R.concat(R.map(val, args), $args));
+        });
+    });
+  };
+
+  var def = function(name, types, f) {
+    return curry(name, R.zip(R.range(0, types.length), types), f);
+  };
+
   var extend = function(Child, Parent) {
     function Ctor() {
       this.constructor = Child;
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -79,6 +79,46 @@
...
+        var typePair = typePairs[idx];
+        if (args[idx] === _) {
+          $typePairs.push(typePair);
+        } else if (typePair[1] !== Any && !R.is(typePair[1], args[idx])) {
+          throw new TypeError(R.join(' ', [
+            name,
+            'requires a value of type',
+            /^function (\w*)/.exec(typePair[1])[1],
+            'as its',
+            ['first', 'second', 'third', 'fourth', 'fifth'][typePair[0]],
```
sharonrapoport: you really dont like functions that take many arguments ah... what happens if you're out of bound here?

davidchambers: This function is only used internally, so we needn't support arities greater than the maximum arity of a Sanctuary function. This is currently three, so we could get away with `['first', 'second', 'third']`. If we decide to expose this function at some point we'll need to update it to handle arbitrary arities.

sharonrapoport:
well, this code is obviously beyond my understanding, but I very much like the more detailed error logging which I hit all the time :)

davidchambers:
> I very much like the more detailed error logging which I hit all the time

I hope it will be as useful in practice as it is in theory. :)

benperez:
Looks pretty cool :dromedary_camel:

paldepind:
> I hope it will be as useful in practice as it is in theory. :)

I'm sure it will!

index.js
```diff
@@ -79,6 +79,46 @@
...
+  /* istanbul ignore next */
+  var Any = function Any() {};
+
+  var curry = function(name, typePairs, f) {
+    return R.arity(typePairs.length, function() {
+      var args = arguments;
+      var $typePairs = [];
+      for (var idx = 0; idx < args.length; idx += 1) {
+        var typePair = typePairs[idx];
+        if (args[idx] === _) {
```
davidchambers: I'd like to wait until Ramda v0.15.0 is available so that I can update this to:

```javascript
if (args[idx] != null && args[idx]['@@functional/placeholder'] === true) {
```

See ramda/ramda#1156 for details.

index.js
```diff
@@ -218,11 +266,11 @@
...
   //. Nothing()
   //. ```
-  Maybe.prototype.empty = Maybe.empty;
+  Maybe.prototype.empty = def('Maybe#empty', [], Maybe.empty);
 
-  //# Maybe#equals :: Maybe a ~> Maybe a -> Boolean
+  //# Maybe#equals :: Maybe a ~> b -> Boolean
   //.
-  //. Takes a Maybe and returns `true` if:
+  //. Takes a value of any type and returns `true` if:
```
davidchambers: I realized that the current type is too restrictive. `S.Just(42).equals([1, 2, 3])` should evaluate false rather than throw. The current implementation does not type check its argument; this simply updates the documentation to match the actual behaviour. I also added tests to ensure that the argument needn't be a Maybe.

pvrnav:
This will possibly change lives.

michaelckelly:
This is a really nice change and will make working with Sanctuary a more pleasant experience. :+1:

svozza:
Very impressive, David. Really looking forward to using this when it is merged.

index.js

index.js
```diff
@@ -79,6 +79,58 @@
...
+  var Any = function Any() {};
+
+  var arity = function(n, f) {
+    switch (n) {
+      case 0: return function() { return f.apply(this, arguments); };
+      case 1: return function(a) { return f.apply(this, arguments); };
+      case 2: return function(a, b) { return f.apply(this, arguments); };
+      case 3: return function(a, b, c) { return f.apply(this, arguments); };
+    }
+  };
```
davidchambers: `R.arity` was deprecated in v0.15.0, so we must provide this ourselves. It's not onerous since we don't export any functions with arity greater than three.

PR #52, opened 2015-06-15
list: document string support by S.indexOf, S.lastIndexOf, and S.slice

davidchambers:
[`R.indexOf`][1] and [`R.lastIndexOf`][2] now dispatch thanks to ramda/ramda#1113. [`R.slice`][3] has dispatched in its current manner since [v0.8.0][4].

[1]: http://ramdajs.com/docs/#indexOf
[2]: http://ramdajs.com/docs/#lastIndexOf
[3]: http://ramdajs.com/docs/#slice
[4]: https://github.com/ramda/ramda/blob/v0.8.0/ramda.js#L3542

### DIFF
--- a/index.js
+++ b/index.js
@@ -876,10 +876,10 @@
   //. Accepts negative indices, which indicate an offset from the end of
   //. the list.
   //.
-  //. ```javascript
-  //. > S.slice(1, 1, ['a', 'b', 'c', 'd', 'e'])
-  //. Just([])
+  //. Dispatches to its third argument's `slice` method if present. As a
+  //. result, one may replace `[a]` with `String` in the type signature.
   //.
+  //. ```javascript
   //. > S.slice(1, 3, ['a', 'b', 'c', 'd', 'e'])
   //. Just(["b", "c"])
   //.
@@ -889,11 +889,11 @@
   //. > S.slice(2, -0, ['a', 'b', 'c', 'd', 'e'])
   //. Just(["c", "d", "e"])
   //.
-  //. > S.slice(0, 2, [])
-  //. Nothing()
-  //.
   //. > S.slice(1, 6, ['a', 'b', 'c', 'd', 'e'])
   //. Nothing()
+  //.
+  //. > S.slice(2, 6, 'banana')
+  //. Just("nana")
   //. ```
   var slice = S.slice = R.curry(function(start, end, xs) {
     var len = xs.length;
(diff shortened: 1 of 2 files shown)


test/index.js
```diff
@@ -1278,11 +1285,16 @@ describe('list', function() {
     });
 
     it('returns a Nothing if the element is not found', function() {
-      eq(S.indexOf(10, [1, 2, 3, 4, 5]), S.Nothing());
+      eq(S.indexOf('x', ['b', 'a', 'n', 'a', 'n', 'a']), S.Nothing());
     });
 
     it('returns Just the index of the element found', function() {
-      eq(S.indexOf(3, [1, 2, 3, 4, 5]), S.Just(2));
+      eq(S.indexOf('a', ['b', 'a', 'n', 'a', 'n', 'a']), S.Just(1));
```
davidchambers: This is a slight improvement as it asserts that Just the index of the *first* occurrence is returned. The other similar assertions are all fine, but I updated them to match.

index.js
```diff
@@ -1057,12 +1057,22 @@
   //. of the first occurrence of the value in the list, if applicable;
   //. Nothing otherwise.
   //.
+  //. Dispatches to its second argument's `indexOf` method if present.
+  //. As a result, `String -> String -> Maybe Number` is an alternative
+  //. type signature.
+  //.
```
davidchambers: @buzzdecafe, @CrossEye, @paldepind, @raine: What do you think of this wording? Should we include a similar paragraph in the docstring of each Ramda function which dispatches? (The second sentence would not be required in all cases.)

raine: Could a tag work somehow? The obvious benefit is being able to parse those programmatically and show them in docs in some special way if so preferred.

By the way, what's the significance of dot after `//`, if any?

davidchambers: > Could a tag work somehow? The obvious benefit is being able to parse those programmatically and show them in docs in some special way if so preferred.

That could work for Ramda. Perhaps something along the lines of `@dispatchesTo reject, filter`.

> By the way, what's the significance of dot after `//`, if any?

`//.` is [Transcribe][1]'s default `--prefix` value: lines prefixed in this manner appear in the [documentation][2].


[2]: https://github.com/plaid/sanctuary#readme

charleswli:
:palm_tree:

index.js
raine: I think the wording is spot on.

PR #54, opened 2015-06-17
either: define extend

pvrnav:
Defines an `extend` method for the `Either` type as per the [fantasyland](https://github.com/fantasyland/fantasy-land#extend) spec.

### DIFF
--- a/index.js
+++ b/index.js
@@ -635,6 +635,23 @@
   //. false
   //. ```
 
+  //# Either#extend :: Either a b ~> (b -> b) -> Either a b
+  //.
+  //. Takes a function and returns `this` if `this` is a Left; otherwise it
+  //. returns a Right whose value is the result of applying the function to
+  //. this Right's value. `Either#extend` is a restricted form of
+  //. [`Either#map`](#Either.prototype.map): the function provided must
+  //. return a value of the same type as its input (`Either#map` does not
+  //. have this restriction).
+  //.
+  //. ```javascript
+  //. > S.Left('Cannot divide by zero').extend(R.inc)
+  //. Left("Cannot divide by zero")
+  //.
+  //. > S.Right(42).extend(R.inc)
+  //. Right(43)
+  //. ```
+
   //# Either#map :: Either a b ~> (b -> c) -> Either a c
   //.
   //. Takes a function and returns `this` if `this` is a Left; otherwise it
@@ -645,8 +662,8 @@
   //. > S.Left('Cannot divide by zero').map(R.inc)
   //. Left("Cannot divide by zero")
   //.
-  //. > S.Right(42).map(R.inc)
-  //. Right(43)
+  //. > S.Right([1, 2, 3]).map(R.sum)
+  //. Right(6)
   //. ```
 
   //# Either#of :: Either a b ~> b -> Either a b
@@ -724,6 +741,9 @@
     return x instanceof Left && R.eqProps('value', x, this);
   });
 
+  //  Left#extend :: Either a b ~> (b -> b) -> Either a b
+  Left.prototype.extend = def('Left#extend', [Function], self);
+
   //  Left#map :: Either a b ~> (b -> c) -> Either a c
   Left.prototype.map = def('Left#map', [Function], self);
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -724,6 +724,11 @@
     return x instanceof Left && R.eqProps('value', x, this);
   });
 
+  //  Left#extend :: Either a b ~> (b -> b) -> Either a b
+  Left.prototype.extend = def('Left#extend', [Function], function(f) {
+    return this;
+  });
```
davidchambers: There's a `self` function defined near the top of the file as `function() { return this; }`. Let's use that here. :)

pvrnav: Done.

index.js
```diff
@@ -771,6 +774,11 @@
     return x instanceof Right && R.eqProps('value', x, this);
   });
 
+  //  Right#extend :: Either a b ~> (b -> b) -> Either a b
+  Right.prototype.extend = def('Right#extend', [Function], function(f) {
+    return Right(f(this.value));
```
davidchambers: We're currently inconsistent in our use of `new`. We use it in the (identical) definition of `Right#map` below. We should open a separate pull request to remove the unnecessary uses of `new`.

index.js
```diff
@@ -771,6 +774,11 @@
     return x instanceof Right && R.eqProps('value', x, this);
   });
 
+  //  Right#extend :: Either a b ~> (b -> b) -> Either a b
```
davidchambers: Does the function need to be of type `b -> b`, or could it be of type `b -> c`?

pvrnav: The spec defines:
>  `w.extend(f)`

and then goes on to say:
> `f` must return a value of type `v`, for some variable `v` contained in `w`

so I think `b -> b` is correct.

index.js
pvrnav: Absolutely. The documentation states that:
```
Though this is a constructor function the `new` keyword needn't be used.
```
so I think we should be consistent in that regard.

index.js
davidchambers: As discussed via HipChat, your interpretation is absolutely right, @pvrnav.

In `w.extend(f)`:

> `f` must return a value of type `v`, for some variable `v` contained in `w`.

So `f` must indeed be of type `v -> v` (or `b -> b` as written above).

index.js
```diff
@@ -645,8 +661,8 @@
   //. > S.Left('Cannot divide by zero').map(R.inc)
   //. Left("Cannot divide by zero")
   //.
-  //. > S.Right(42).map(R.inc)
-  //. Right(43)
+  //. > S.Right(42).map(R.toString)
+  //. Right("42")
```
pvrnav: Updated this slightly to emphasize the difference between `extend` and `map`.

index.js
davidchambers: #55

pvrnav: :+1:

index.js
```diff
@@ -635,6 +635,22 @@
   //. false
   //. ```
 
+  //# Either#extend :: Either a b ~> (b -> b) -> Either a b
+  //.
+  //. Takes a function and returns `this` if `this` is a Left; otherwise it
+  //. returns a Right whose value is the result of applying the function to
+  //. this Right's value. Keep in mind that the difference between this and
+  //. `map` is that the type of the Right remains the same even after
+  //. applying `extend`.
```
davidchambers: It's a bit confusing to write _the difference between this and `map`_ in a paragraph with multiple references to `this`. We should write _this function_ or _`extend`_ instead.

Perhaps the second sentence should read:

_`Either#extend` is a restricted form of [`Either#map`][1]: the function provided must return a value of the same type as its input (`Either#map` does not have this restriction)._


[1]: https://github.com/plaid/sanctuary#Either.prototype.map

index.js
davidchambers: Good idea! Something like `S.Right([1, 2, 3, 4]).map(R.sum)` would work, too.

test/index.js
```diff
@@ -656,6 +656,15 @@ describe('either', function() {
       // jshint +W053
     });
 
+    it('provides an "extend" method', function() {
+      eq(S.Left(42).extend(function(x) { return x / 2; }), S.Left(42));
```
davidchambers: Let's begin by asserting arity:

```javascript
eq(S.Left('abc').extend.length, 1);
```

I'd also prefer to use `S.Left('abc')` in the `x / 2` example above, since a Left will typically contain an error message.

index.js
```diff
@@ -635,6 +635,22 @@
   //. false
   //. ```
 
+  //# Either#extend :: Either a b ~> (b -> b) -> Either a b
+  //.
+  //. Takes a function and returns `this` if `this` is a Left; otherwise it
+  //. returns a Right whose value is the result of applying the function to
+  //. this Right's value. `Either#extend` is a restricted form of
+  //. [`Either#map`][1]: the function provided must return a value of the
```
davidchambers: This should link to the appropriate section of the documentation. I suggest

    [`Either#map`](#Either.prototype.map)

since we're linking to a section in the same document.

davidchambers:
:deciduous_tree: Terrific pull request, @pvrnav!

PR #56, opened 2015-06-17
maybe: define extend

pvrnav:
Defines an `extend` method for the `Maybe` type as per the [fantasyland](https://github.com/fantasyland/fantasy-land#extend) spec.


davidchambers:
:deciduous_tree: Very nice!

PR #58, opened 2015-06-17
extend: correct the behavior and add tests to check associativity

pvrnav:
Added additional tests to check that the `extend` method satisfies the property laid out at https://github.com/fantasyland/fantasy-land#extend-method. Also fixes implementations to adhere to those specs.

### DIFF
--- a/index.js
+++ b/index.js
@@ -181,7 +181,8 @@
   //. The Maybe type represents optional values: a value of type `Maybe a` is
   //. either a Just whose value is of type `a` or a Nothing (with no value).
   //.
-  //. The Maybe type satisfies the [Monoid][] and [Monad][] specifications.
+  //. The Maybe type satisfies the [Monoid][], [Monad][], and [Extend][]
+  //. specifications.
   var Maybe = S.Maybe = function Maybe() {
     throw new Error('Cannot instantiate Maybe');
   };
@@ -309,6 +310,20 @@
   //. false
   //. ```
 
+  //# Maybe#extend :: Maybe a ~> (Maybe a -> a) -> Maybe a
+  //.
+  //. Takes a function and returns `this` if `this` is a Nothing; otherwise
+  //. it returns a Just whose value is the result of applying the function to
+  //. `this`.
+  //.
+  //. ```javascript
+  //. > S.Nothing().extend(function(x) { return x.value + 1; })
+  //. Nothing()
+  //.
+  //. > S.Just(42).extend(function(x) { return x.value + 1; })
+  //. Just(43)
+  //. ```
+
   //# Maybe#filter :: Maybe a ~> (a -> Boolean) -> Maybe a
   //.
   //. Takes a predicate and returns `this` if `this` is a Just whose value
@@ -335,8 +350,8 @@
   //. > S.Nothing().map(R.inc)
   //. Nothing()
   //.
-  //. > S.Just(42).map(R.inc)
-  //. Just(43)
+  //. > S.Just([1, 2, 3]).map(R.sum)
+  //. Just(6)
   //. ```
 
   //# Maybe#of :: Maybe a ~> b -> Maybe b
@@ -407,6 +422,9 @@
   //  Nothing#equals :: Maybe a ~> b -> Boolean
   Nothing.prototype.equals = def('Nothing#equals', [Any], R.is(Nothing));
 
+  //  Nothing#extend :: Maybe a ~> (Maybe a -> a) -> Maybe a
+  Nothing.prototype.extend = def('Nothing#extend', [Function], self);
+
   //  Nothing#map :: Maybe a ~> (a -> b) -> Maybe b
   Nothing.prototype.map = def('Nothing#map', [Function], self);
 
@@ -454,6 +472,11 @@
   //  Just#equals :: Maybe a ~> b -> Boolean
   Just.prototype.equals = def('Just#equals', [Any], function(x) {
     return x instanceof Just && R.eqProps('value', x, this);
+  });
+
+  //  Just#extend :: Maybe a ~> (Maybe a -> a) -> Maybe a
+  Just.prototype.extend = def('Just#extend', [Function], function(f) {
+    return Just(f(this));
   });
 
   //  Just#map :: Maybe a ~> (a -> b) -> Maybe b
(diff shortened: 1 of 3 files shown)


davidchambers:
It would be nice to start by defining `w`, `f`, and `g` in each case. For example:

```javascript
var w = S.Just(7);
var f = R.inc;
var g = square;
```

We can then almost copy and paste the code from the spec:

```javascript
eq(w.extend(g).extend(f),
   w.extend(function(_w) { return f(_w.extend(g)); }));
```

Testing this in the REPL revealed a problem with our implementation: the argument to the provided function must be an Extend. We should be applying the function to `this` rather than `this.value`.

We should update the docstrings, too, since `extend` is actually significantly different from `map`. We're learning as we go! :)

pvrnav:
Made changes accordingly.

index.js
```diff
@@ -309,20 +309,17 @@
...
-  //# Maybe#extend :: Maybe a ~> (a -> a) -> Maybe a
+  //# Maybe#extend :: Maybe a ~> (Maybe a -> a) -> Maybe a
   //.
   //. Takes a function and returns `this` if `this` is a Nothing; otherwise
   //. it returns a Just whose value is the result of applying the function to
-  //. this Just's value. `Maybe#extend` is a restricted form of
-  //. [`Maybe#map`](#Maybe.prototype.map): the function provided must
-  //. return a value of the same type as its input (`Maybe#map` does not
-  //. have this restriction).
+  //. this Just's value.
```
davidchambers: This should read _`this`_ rather than _this Just's value_.

test/index.js
```diff
@@ -152,7 +152,14 @@ describe('maybe', function() {
...
       eq(S.Nothing().extend.length, 1);
-      eq(S.Nothing().extend(function() { return 42; }), S.Nothing());
+      eq(S.Nothing().extend(function(x) { return x.value; }), S.Nothing());
+
+      // Extend Property Test
+      var w = S.Nothing();
+      var f = function(x) { return x.value + 1; };
+      var g = function(x) { return x.value * x.value; };
+      eq(w.extend(g).extend(f),
+         w.extend(function(_w) { return f(_w.extend(g)); }));
```
davidchambers: This is the associativity law, which can be written `extend f ∘ extend g = extend (f ∘ extend g)`.

Let's update the comment to reflect this (we use `// associativity` elsewhere).

test/index.js
```diff
@@ -359,7 +366,16 @@ describe('maybe', function() {
 
     it('provides an "extend" method', function() {
       eq(S.Just(42).extend.length, 1);
-      eq(S.Just(42).extend(function(x) { return x / 2; }), S.Just(21));
+      eq(S.Just(42).extend(function(x) {
+        return x.value / 2;
+      }), S.Just(21));
```
davidchambers: This would read better on one line.

test/index.js
pvrnav: Changed.

test/index.js
pvrnav: Changed.

index.js
pvrnav: Changed.

test/index.js
```diff
@@ -152,7 +152,14 @@ describe('maybe', function() {
 
     it('provides an "extend" method', function() {
       eq(S.Nothing().extend.length, 1);
-      eq(S.Nothing().extend(function() { return 42; }), S.Nothing());
+      eq(S.Nothing().extend(function(x) { return x.value; }), S.Nothing());
```
davidchambers: Let's make this `x.value / 2` to match the Just one.

pvrnav: Gotcha.

index.js
```diff
@@ -478,7 +476,7 @@
 
   //  Just#extend :: Maybe a ~> (a -> a) -> Maybe a
```
davidchambers: We should update this type signature.

pvrnav: Updated all the type signatures, sorry I missed those.

davidchambers: No worries. That's why we do code review. :)

index.js
```diff
@@ -660,20 +659,17 @@
...
   //. Takes a function and returns `this` if `this` is a Left; otherwise it
   //. returns a Right whose value is the result of applying the function to
-  //. this Right's value. `Either#extend` is a restricted form of
-  //. [`Either#map`](#Either.prototype.map): the function provided must
-  //. return a value of the same type as its input (`Either#map` does not
-  //. have this restriction).
+  //. `this`.
   //.
   //. ```javascript
   //. > S.Left('Cannot divide by zero').extend(R.inc)
```
davidchambers: We should replace `R.inc` with a function of the right type.

pvrnav: How about `toString`?

Defining my own function gets too long for the line.

davidchambers: I believe the 79-character maximum line length is no longer enforced by `make lint`. It seems reasonable to exceed the limit in this case.

pvrnav: Changed.

index.js
```diff
@@ -181,7 +181,8 @@
   //. The Maybe type represents optional values: a value of type `Maybe a` is
   //. either a Just whose value is of type `a` or a Nothing (with no value).
   //.
-  //. The Maybe type satisfies the [Monoid][] and [Monad][] specifications.
+  //. The Maybe type satisfies the [Extend][], [Monoid][], and [Monad][]
+  //. specifications.
```
davidchambers: I'd like to list Extend last. If one day we upgrade from Extend to Comonad (which may or may not be possible) it would be strange to list Comonad before Monad. As a bonus, Monoid, Monad, Extend matches the order the type classes appear in the spec.

pvrnav: Makes sense. I was just going for alphabetical order.

index.js
```diff
@@ -181,7 +181,8 @@
   //. The Maybe type represents optional values: a value of type `Maybe a` is
   //. either a Just whose value is of type `a` or a Nothing (with no value).
   //.
-  //. The Maybe type satisfies the [Monoid][] and [Monad][] specifications.
+  //. The Maybe type satisfies the [Monoid][], [Monad][], and [Extend][].
```
davidchambers: There's a stray full stop (`.`) at the end of this line.

davidchambers:
:deciduous_tree:

PR #59, opened 2015-06-18
extend: added support for lists

pvrnav:
Lists now support the extend method. I've chosen to have the function apply in this form to a list, say `[1,2,3]`:
```
[ f [1, 2, 3], f [2, 3], f [3] ]
```
Thoughts @davidchambers and @benperez?

### DIFF
--- a/index.js
+++ b/index.js
@@ -975,6 +975,31 @@
       Nothing();
   });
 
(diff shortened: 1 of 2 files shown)


davidchambers:
This function belongs in Ramda, if anywhere.

PR #61, opened 2015-06-24
descriptive error messages for type mismatches

davidchambers:
Before:

    > S.and([], S.Nothing())
    TypeError: Type mismatch

After:

    > S.and([], S.Nothing())
    TypeError: and requires its first and second arguments to be of the same type; [] and Nothing() are not

:memo: _Can anyone suggest improvements to the wording of this message?_

It turned out to be useful to have the implementation of the internal `curry` function further diverge from the approach used by Ramda. The recursive call now appears in this expression:

```javascript
return args.length === len ? f.apply(this, args)
                           : curry(name, types, values, f);
```

When `curry` recurses, the *same* function is provided. Only the third argument differs. In the application of `S.take` to `5`, `values` would be `[R.__, R.__]` initially and `[5, R.__]` in the recursive call.

I haven't checked whether this approach has anything in common with @megawac's metadata-based :curry: in Ramda, but it's interesting to see a solution which doesn't provide a new function each time it recurses.

### DIFF
--- a/index.js
+++ b/index.js
@@ -84,8 +84,24 @@
     return x != null && x['@@functional/placeholder'] === true;
   };
 
-  /* istanbul ignore next */
-  var Any = function Any() {};
+  var formatters = {
+    '{}': R.identity,
+    '{ord}': R.nth(_, ['first', 'second', 'third']),
+    '{repr}': R.toString,
+    '{type}': R.pipe(String, R.match(/^function (\w*)/), R.nth(1))
+  };
+
+  //  format :: (String, [*]) -> String
+  var format = function(template, values) {
+    var idx = -1;
+    return template.replace(/[{].*?[}]/g, function(match) {
+      return formatters[match](values[idx += 1]);
+    });
+  };
+
+  var a = {name: 'a'};
+  var b = {name: 'b'};
+  var c = {name: 'c'};
 
   var arity = function(n, f) {
     switch (n) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -96,39 +112,61 @@
...
+
+      var args = R.reject(placeholder, values);
+      return args.length === len ? f.apply(this, args)
+                                 : curry(name, types, values, f);
     });
   };
 
   var def = function(name, types, f) {
-    return curry(name, R.zip(R.range(0, types.length), types), f);
+    var values = R.times(function() { return _; }, types.length);
```
davidchambers: `R.repeat(_, types.length)` does not work here. :stuck_out_tongue_closed_eyes:

benperez:
The implementation is over my head but the results look cool! :dromedary_camel:

index.js
```diff
@@ -84,8 +84,24 @@
...
+  var format = function(template, values) {
+    var idx = -1;
+    return template.replace(/[{].*?[}]/g, function(match) {
+      return formatters[match](values[idx += 1]);
+    });
+  };
+
+  var a = {name: 'a'};
+  var b = {name: 'b'};
+  var c = {name: 'c'};
```
megawac: At a glance doens't look to have much overlap with metadata. What are these variables?

davidchambers: > What are these variables?

They represent type variables, which place constraints on argument combinations.

In this case the arguments can be of any combination of types:

```javascript
//# K :: a -> b -> a
S.K = def('K', [a, b], function(x, y) {
  return x;
});
```

In this case the arguments can be of any type, but must both be of the same type:

```javascript
//# and :: a -> a -> a
S.and = def('and', [a, a], function(x, y) {
  return toBoolean(x) ? y : x;
});
```

paldepind:
This looks like a really solid feature!

> When curry recurses, the same function is provided.

After my latest rewrite of it that is the case with `R.curry` as well. Actually that was one of the important things I changed. It also only uses one loop which I find favorable. I actually also implemented typechecking on top of it as well. The changed `_curry` is here just for inspiration (it's burdened by `while` loops though):

```javascript
module.exports = (function() {
  var ordinals = ['first', 'second', 'third', 'fourth', 'fifth', 'sixth'];
  return function _curryN(checkers, received, fn) {
    var length = checkers.length;
    return function() {
      var combined = [];
      var argsIdx = 0;
      var left = length;
      var combinedIdx = 0;
      while (combinedIdx < received.length || argsIdx < arguments.length) {
        var result;
        if (combinedIdx < received.length &&
            (!placeholder(received[combinedIdx]) || argsIdx >= arguments.length)) {
          result = received[combinedIdx];
        } else {
          result = arguments[argsIdx];
          argsIdx += 1;
        }
        combined[combinedIdx] = result;
        if (!placeholder(result)) {
          if (checkers[combinedIdx](result) === false) {
            throw new TypeError(
              fn.name + ' expected ' + checkers[combinedIdx].desc + ' as its ' +
              ordinals[combinedIdx] + ' argument but received ' + toString(result)
            );
          }
          left -= 1;
        }
        combinedIdx += 1;
      }
      return left <= 0 ? fn.apply(this, combined) : arity(left, _curryN(checkers, combined, fn));
    };
  };
})();
```

davidchambers:
> > When curry recurses, the same function is provided.
>
> After my latest rewrite of it that is the case with `R.curry` as well.

Ah, yes, I'd forgotten that ramda/ramda#1155 included that change. That may be where I got the idea! :)

index.js
davidchambers: `R.map(function() { return _; }, types)` is a better solution.

PR #62, opened 2015-06-28
add Int pseudotype

### LINKED ISSUE #45: integer type
This may or may not be useful:

```javascript
> S.Int(7).add(S.Int(2))
Int(9)
> S.Int(7).divideBy(S.Int(2))
Int(3)
> S.Int(7).valueOf()
7
```

An integer type would allow us to specify types more precisely. [`S.at`][1], for example, could have type `Int -> [a] -> Maybe a`.

[1]: https://github.com/plaid/sanctuary#at

davidchambers:
Closes #45

This pull request defines the following functions:

    S.Int      :: Number -> Int !
    S.Int.add  :: Int -> Int -> Int
    S.Int.sub  :: Int -> Int -> Int
    S.Int.mul  :: Int -> Int -> Int
    S.Int.quot :: Int -> Int -> Int !
    S.Int.rem  :: Int -> Int -> Int !
    S.Int.div  :: Int -> Int -> Int !
    S.Int.mod  :: Int -> Int -> Int !
    S.Int.and  :: Int -> Int -> Int
    S.Int.or   :: Int -> Int -> Int
    S.Int.xor  :: Int -> Int -> Int
    S.Int.not  :: Int -> Int
    S.Int.even :: Int -> Boolean
    S.Int.odd  :: Int -> Boolean

The `S.Int` namespace is required if we're to avoid overloading existing functions: [`S.and`][S.and], [`S.or`][S.or], and [`S.xor`][S.xor]. The use of a namespace is inconsistent with [`S.fromMaybe`][S.fromMaybe] and [`S.either`][S.either] which each apply to just one type yet are defined at the top level. @scott-christopher is grappling with exactly this issue in ramda/ramda-fantasy#56. Alternatively, we could use the `S.Int` namespace just for the three conflicting functions, though this would trade one inconsistency for another. Which approach is preferable?

@svozza, I decided against making `S.Int.{quot,rem,div,mod}` return `Maybe Int` to handle a divisor of zero (discussed in #41 and #44). Having to unwrap a Maybe when one *knows* it will always be a Just—as with `S.Int.quot(n, 2)`, for example—is a bit tedious. I opted instead to throw exceptions:

```javascript
S.Int.quot(1, 0);         // Error: Cannot divide by zero
```

This pull request also updates the types of several existing functions to use Int in place of Number. `S.take`, for example, now has type `Int -> [a] -> Maybe [a]`:

```javascript
// Before:
S.take(0.5);              // => Function
S.take(0.5, [1, 2, 3]);   // => Just([])

// After:
S.take(0.5);              // TypeError: take requires a value of type Int as its first argument; received 0.5
S.take(0.5, [1, 2, 3]);   // TypeError: take requires a value of type Int as its first argument; received 0.5
```

Since the README is updated only when we publish a new version, I've included the documentation for the Int type below.

---

### Int type

The Int pseudotype represents integers in the range [-2^31 .. 2^31). It is a pseudotype because each Int is represented by a Number value. Sanctuary's run-time type checking asserts that a valid Number value is provided wherever an Int value is required.

#### `Int :: Number -> Int !`

Throws if its argument is not a valid Int value.

```javascript
> S.Int(42)
42
```

#### `Int.add :: Int -> Int -> Int`

Returns the sum of its two arguments.

```javascript
> S.Int.add(1, 2)
3
```

#### `Int.sub :: Int -> Int -> Int`

Returns the result of subtracting its second argument from its first argument.

```javascript
> S.Int.sub(1, 2)
-1
```

#### `Int.mul :: Int -> Int -> Int`

Returns the product of its two arguments.

```javascript
> S.Int.mul(6, 7)
42
```

#### `Int.quot :: Int -> Int -> Int !`

Returns the result of dividing its first argument by its second argument, truncating towards zero.

Throws if the divisor is zero.

See also `Int.div`.

```javascript
> S.Int.quot(42, 5)
8

> S.Int.quot(42, -5)
-8

> S.Int.quot(-42, 5)
-8

> S.Int.quot(-42, -5)
8
```

#### `Int.rem :: Int -> Int -> Int !`

Integer remainder, satisfying:

    Int.quot(x, y) * y + Int.rem(x, y) === x

Throws if the divisor is zero.

See also `Int.mod`.

```javascript
> S.Int.rem(42, 5)
2

> S.Int.rem(42, -5)
2

> S.Int.rem(-42, 5)
-2

> S.Int.rem(-42, -5)
-2
```

#### `Int.div :: Int -> Int -> Int !`

Returns the result of dividing its first argument by its second argument, truncating towards negative infinity.

Throws if the divisor is zero.

See also `Int.quot`.

```javascript
> S.Int.div(42, 5)
8

> S.Int.div(42, -5)
-9

> S.Int.div(-42, 5)
-9

> S.Int.div(-42, -5)
8
```

#### `Int.mod :: Int -> Int -> Int !`

Integer modulus, satisfying:

    Int.div(x, y) * y + Int.mod(x, y) === x

Throws if the divisor is zero.

See also `Int.rem`.

```javascript
> S.Int.mod(42, 5)
2

> S.Int.mod(42, -5)
-3

> S.Int.mod(-42, 5)
3

> S.Int.mod(-42, -5)
-2
```

#### `Int.and :: Int -> Int -> Int`

[Bitwise AND][&]. Returns an Int with a one at each bit position at which both arguments have a one.

```javascript
> S.Int.and(parseInt('1100', 2), parseInt('1010', 2))
8
```

#### `Int.or :: Int -> Int -> Int`

[Bitwise OR][|]. Returns an Int with a one at each bit position at which at least one argument has a one.

```javascript
> S.Int.or(parseInt('1100', 2), parseInt('1010', 2))
14
```

#### `Int.xor :: Int -> Int -> Int`

[Bitwise XOR][^]. Returns an Int with a one at each bit position at which exactly one argument has a one.

```javascript
> S.Int.xor(parseInt('1100', 2), parseInt('1010', 2))
6
```

#### `Int.not :: Int -> Int`

[Bitwise NOT][~], satisfying:

    Int.not(x) === -(x + 1)

```javascript
> S.Int.not(42)
-43
```

#### `Int.even :: Int -> Boolean`

Returns `true` if its argument is even; `false` if it is odd.

```javascript
> S.Int.even(42)
true
```

#### `Int.odd :: Int -> Boolean`

Returns `true` if its argument is odd; `false` if it is even.

```javascript
> S.Int.odd(42)
false
```

[~]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#Bitwise_NOT
[&]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#Bitwise_AND
[|]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#Bitwise_OR
[^]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#Bitwise_XOR
[S.and]: https://github.com/plaid/sanctuary#and
[S.either]: https://github.com/plaid/sanctuary#either
[S.fromMaybe]: https://github.com/plaid/sanctuary#fromMaybe
[S.or]: https://github.com/plaid/sanctuary#or
[S.xor]: https://github.com/plaid/sanctuary#xor

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,6 +1,6 @@
 {
   "disallowEmptyBlocks": true,
-  "disallowImplicitTypeConversion": ["binary", "boolean", "numeric", "string"],
+  "disallowImplicitTypeConversion": ["boolean", "numeric", "string"],
   "disallowKeywordsOnNewLine": ["else", "catch", "finally"],
   "disallowMixedSpacesAndTabs": true,
   "disallowMultipleLineStrings": true,
--- a/.jshintrc
+++ b/.jshintrc
@@ -23,7 +23,7 @@
   "boss": false,
   "debug": false,
   "eqnull": true,
-  "esnext": true,
+  "esnext": false,
   "evil": false,
   "expr": false,
   "funcscope": false,
--- a/index.js
+++ b/index.js
@@ -31,7 +31,7 @@
 //.
 //. Sanctuary uses Haskell-like type signatures to describe the types of
 //. values, including functions. `'foo'`, for example, has type `String`;
-//. `[1, 2, 3]` has type `[Number]`. The arrow (`->`) is used to express a
+//. `[1, 2, 3]` has type `[Int]`. The arrow (`->`) is used to express a
 //. function's type. `Math.abs`, for example, has type `Number -> Number`.
 //. That is, it takes an argument of type `Number` and returns a value of
 //. type `Number`.
@@ -58,6 +58,13 @@
 //. _When the `map` method is invoked on a value of type `Maybe a`
 //. (for any type `a`) with an argument of type `a -> b` (for any type `b`),
 //. it returns a value of type `Maybe b`._
+//.
+//. An exclamation mark (`!`) at the end of a type signature indicates
+//. a partial function: a function which is not defined for all possible
+//. arguments of the specified type. [`Int.div`](#Int.div), for example,
+//. is undefined for a divisor of zero. In Sanctuary, partial functions
+//. throw when given input for which the output is indeterminate, rather
+//. than return a failure value such as `NaN` or `undefined`.
 //.
 //. ## API
 
@@ -103,6 +110,9 @@
   var b = {name: 'b'};
   var c = {name: 'c'};
 
+  //  int :: a -> Boolean
+  var int = function(x) { return R.is(Number, x) && (x | 0) === Number(x); };
+
   var arity = function(n, f) {
     switch (n) {
       case 0: return function() { return f.apply(this, arguments); };
@@ -147,7 +157,7 @@
               ));
             }
           }
-        } else if (!R.is(type, arg)) {
+        } else if (!(type === Int ? int(arg) : R.is(type, arg))) {
           throw new TypeError(format(
             '{} requires a value of type {type} as its {ord} argument; ' +
             'received {repr}',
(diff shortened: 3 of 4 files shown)


svozza:
Nice work! I agree that it makes more sense to namespace all the `Int` functions rather than just three of them.

paldepind:
Where the bitwise functions created out of an actual need? Otherwise I don't really think they make sense.

davidchambers:
> Where the bitwise functions created out of an actual need?

No. They were included to satisfy my desire for completeness. :)

davidchambers:
Superseded by plaid/sanctuary-int#1

PR #63, opened 2015-07-09
classify: add S.is

davidchambers:
ramda/ramda#296 made the behaviour of `R.is` inconsistent for primitives:

```javascript
> R.is(Number, new Number(0))
true
> R.is(Object, new Number(0))
true
> R.is(Number, 0)
true
> R.is(Object, 0)
false
```

A primitive is promoted to its object equivalent in the case of `R.is(Number)` but not in the case of `R.is(Object)`.

By removing the `R.is(Object)` special case we're able to replace many occurrences of `a` and `b` with `Object` in the type definitions. This allows us to provide more helpful error messages. For example:

```
> S.head(null)
TypeError: head requires a value of type Object as its first argument; received null
```

We needn't expose `is`, but its more "correct" behaviour may be useful in other situations as well.

The one problem with this approach is that it disallows `Object.create(null)` in contexts where such an object is valid. For example:

```
> S.get('x', Object.create(null))
TypeError: get requires a value of type Object as its second argument; received {}
```

I can think of a solution to this, but it strikes me as an ugly special case:

```javascript
var _is = function(type, x) {
  return x != null && (Object(x) instanceof type || type === Object && R.type(x) === 'Object');
};
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -87,21 +87,28 @@
   var formatters = {
     '{}': R.identity,
     '{ord}': R.nth(_, ['first', 'second', 'third']),
+    '{quote}': function(s) { return '\u2018' + s + '\u2019'; },
     '{repr}': R.toString,
     '{type}': R.pipe(String, R.match(/^function (\w*)/), R.nth(1))
   };
 
-  //  format :: (String, [*]) -> String
-  var format = function(template, values) {
+  //  format :: String -> [*] -> String
+  var format = R.curry(function(template, values) {
     var idx = -1;
     return template.replace(/[{].*?[}]/g, function(match) {
       return formatters[match](values[idx += 1]);
     });
-  };
+  });
 
+  var Accessible = /* istanbul ignore next */ function Accessible() {};
+  var Type = /* istanbul ignore next */ function Type() {};
   var a = {name: 'a'};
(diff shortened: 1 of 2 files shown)


svozza:
Regardless of how ugly that special case is, I think it's important that `Object.create(null)` is supported.

davidchambers:
The correct solution may be to have a private Accessible "type", used as follows:

```javascript
S.head = def('head', [Accessible], at(0));
```

Every value in the language except `null` and `undefined` would be members of this type.

We'd need some extra logic to handle Accessible in `curry`, but this would have the added benefit of improving our error messages:

```
> S.head(null)
TypeError: The first argument to head cannot be null or undefined
```

davidchambers:
I've updated the pull request. Let me know what you think, @svozza!

svozza:
Sorry David, I completely forgot about this! Very nice indeed, although the `curry` function is starting to look pretty hairy now. Unfortunately I think that's unavoidable.

davidchambers:
> [T]he `curry` function is starting to look pretty hairy now.

Indeed. If one sees it as JavaScript's missing run-time type system fused with a :curry: function, the amount of code seems reasonable.

PR #64, opened 2015-07-17
add type argument to S.get, S.gets, and S.pluck

davidchambers:
The fact that `S.get` returns a value of type `Maybe *` has bothered me for some time. I find myself filtering the result to ensure I have a value of the expected type. For example:

```javascript
R.pipe(...,
       S.get('x'),
       R.filter(R.is(Number)),
       ...)
```

This pull request adds a type specifier to `S.get`, `S.gets`, and `S.pluck`. The above will become:

```javascript
R.pipe(...,
       S.get(Number, 'x'),
       ...)
```

The current loose behaviour can be approximated by `S.get(Object)`. It's no longer possible to get null/undefined values. If this is important we could export an Any pseudotype, where Any contains every JavaScript value. Unless there's a use case, I'd prefer not to export and document `S.Any`.

These changes build upon the foundation laid by #63. I've included a section explaining the Accessible pseudotype, as it's now used in type signatures.

One question I have is how to link `Type` and `a` in a signature such as:

```haskell
get :: Type -> String -> Accessible -> Maybe a
```

Any suggestions?

### DIFF
--- a/index.js
+++ b/index.js
@@ -59,6 +59,30 @@
 //. (for any type `a`) with an argument of type `a -> b` (for any type `b`),
 //. it returns a value of type `Maybe b`._
 //.
+//. ### Accessible pseudotype
+//.
+//. What is the type of values which support property access? In other words,
+//. what is the type of which every value except `null` and `undefined` is a
+//. member? Object is close, but `Object.create(null)` produces a value which
+//. supports property access but which is not a member of the Object type.
+//.
+//. Sanctuary uses the Accessible pseudotype to represent the set of values
+//. which support property access.
+//.
+//. ### Type representatives
+//.
+//. What is the type of `Number`? One answer is `a -> Number`, since it's a
+//. function which takes an argument of any type and returns a Number value.
+//. When provided as the first argument to [`is`](#is), though, `Number` is
+//. really the value-level representative of the Number type.
+//.
+//. Sanctuary uses the TypeRep pseudotype to describe type representatives.
+//. For example:
+//.
+//.     Number :: TypeRep Number
+//.
+//. `Number` is the sole inhabitant of the TypeRep Number type.
+//.
 //. ## API
 
 ;(function() {
@@ -101,13 +125,13 @@
   });
 
   var Accessible = /* istanbul ignore next */ function Accessible() {};
-  var Type = /* istanbul ignore next */ function Type() {};
+  var TypeRep = /* istanbul ignore next */ function TypeRep() {};
   var a = {name: 'a'};
   var b = {name: 'b'};
   var c = {name: 'c'};
 
   var _is = function(type, x) {
-    return x != null && Object(x) instanceof type;
+    return x != null && (type === Accessible || Object(x) instanceof type);
   };
 
   var arity = function(n, f) {
@@ -119,7 +143,7 @@
     }
   };
 
-  //  curry :: (String, [Type], [*], Function) -> Function
+  //  curry :: (String, [TypeRep], [*], Function) -> Function
   var curry = function(name, types, _values, f) {
     return arity(R.filter(placeholder, _values).length, function() {
       var values = _values;  // Locally scoped variable to update.
@@ -161,7 +185,7 @@
               [paramIndex, name]
             ));
           }
-        } else if (!_is(type === Type ? Function : type, arg)) {
+        } else if (!_is(type === TypeRep ? Function : type, arg)) {
           throw new TypeError(format(
             '{quote} requires a value of type {type} as its {ord} argument; ' +
             'received {repr}',
@@ -210,11 +234,11 @@
 
   //. ### Classify
 
-  //# is :: Type -> a -> Boolean
+  //# is :: TypeRep a -> b -> Boolean
   //.
-  //. Takes a type and a value of any type and returns `true` if the given
-  //. value is of the specified type (either directly or via the prototype
-  //. chain); `false` otherwise.
+  //. Takes a [type representative](#type-representatives) and a value of
+  //. any type and returns `true` if the given value is of the specified
+  //. type (either directly or via the prototype chain); `false` otherwise.
   //.
   //. Boolean, number, string, and symbol [primitives][] are promoted to
   //. their object equivalents. `42`, for example, is considered a Number
@@ -231,7 +255,7 @@
   //. > S.is(String, 42)
   //. false
   //. ```
-  var is = S.is = def('is', [Type, a], _is);
+  var is = S.is = def('is', [TypeRep, a], _is);
 
   //. ### Combinator
 
@@ -253,7 +277,7 @@
 
   //. ### Maybe type
 
-  //# Maybe :: Type
+  //# Maybe :: TypeRep Maybe
   //.
   //. The Maybe type represents optional values: a value of type `Maybe a` is
   //. either a Just whose value is of type `a` or a Nothing (with no value).
@@ -480,7 +504,7 @@
   //. "Just([1, 2, 3])"
   //. ```
 
-  //# Maybe#type :: Type
+  //# Maybe#type :: TypeRep Maybe
   //.
   //. A reference to the Maybe type. Useful for determining whether two
   //. values such as `S.Nothing()` and `S.Just(42)` are of the same type.
@@ -648,7 +672,7 @@
 
   //. ### Either type
 
-  //# Either :: Type
+  //# Either :: TypeRep Either
   //.
   //. The Either type represents values with two possibilities: a value of type
   //. `Either a b` is either a Left whose value is of type `a` or a Right whose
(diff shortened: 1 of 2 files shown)


pvrnav:
This should be really useful @davidchambers! 

I think saying `Typeof a` is very readable, but there's possibly a better convention out there somewhere.

index.js
```diff
@@ -947,9 +957,9 @@
   //# and :: a -> a -> a
   //.
   //. Takes two values of the same type and returns the second value
-  //. if the first is "true"; the first value otherwise. An array is
-  //. considered "true" if its length is greater than zero. The Boolean
-  //. value `true` is also considered "true". Other types must provide
+  //. if the first is “true”; the first value otherwise. An array is
+  //. considered “true” if its length is greater than zero. The Boolean
+  //. value `true` is also considered “true”. Other types must provide
```
benperez: Should the fancy quotation marks be a part of this PR?

index.js
```diff
@@ -1242,59 +1252,75 @@
...
+  //.
+  //. > S.gets(Number, ['a', 'b', 'c'], {})
   //. Nothing()
   //. ```
-  S.gets = def('gets', [Accessible, Accessible], function(keys, obj) {
-    return R.reduce(function(acc, key) {
-      return R.chain(get(key), acc);
-    }, Just(obj), keys);
+  S.gets =
+  def('gets', [Type, Accessible, Accessible], function(type, keys, obj) {
```
benperez: I'm a little behind on the whole Sanctuary type checking system but is there no List type?

benperez:
> One question I have is how to link Type and a in a signature

You could use the syntax for a class constraint.
```haskell
get :: Type a => Type -> String -> Accessible -> Maybe a
```

index.js
davidchambers: I wanted to use curly quotes around “plucking”, and wanted to be consistent. I'll switch to curly quotes in a separate pull request then rebase this branch. :)

index.js
davidchambers: Correct. All we do in this case is assert that `keys` and `obj` are not null/undefined. I'd welcome a pull request to add a List pseudotype, if you have a definition in mind. :)

davidchambers:
> You could use the syntax for a class constraint.
>
> ```haskell
> get :: Type a => Type -> String -> Accessible -> Maybe a
> ```

Since `a` is already a type variable, I'm not sure what information the `Type a` constraint adds. Also, I don't see how this relates the return type to the type of the first argument.

joneshf:
I think you have to be the type checker in this case. Look at where `Type` flows through your program, and make the decision based on that.

It is sent to `is` in [`get`](https://github.com/plaid/sanctuary/pull/64/files#diff-168726dbe96b3ce427e7fedce31bb0bcR1299)

In [`is`](https://github.com/plaid/sanctuary/pull/64/files#diff-168726dbe96b3ce427e7fedce31bb0bcR120) it is compared to [`Accessible`](https://github.com/plaid/sanctuary/pull/64/files#diff-168726dbe96b3ce427e7fedce31bb0bcR112). So it should probably have a type like: `() -> Accessible`.

It also is used as the RHS of `instanceof` on the same line. So, it must be a constructor. Since what it is checked against is a value (the `a`) it should have a type like: `() -> a`.

Combine the two types and you end up with something like: `() -> (Accessible | a)`, where `|` represents a union.

So putting it all together you get something like: 

```haskell
get :: (() -> (Accessible | a)) -> String -> Accessible -> Maybe a
```

davidchambers:
@joneshf, I don't understand how you arrived at `()` in `() -> (Accessible | a)`. Are you using `()` to indicate zero arguments? Take `String` as an example of a type:

```javascript
> String(42)
'42'
> String(null)
'null'
```

This suggests we have something like `String :: * -> String`. But the "constructor" can be given no arguments (in which case it returns `''`). Then we have `RegExp` which accepts multiple arguments. So perhaps we should replace `Type` with `*... -> a`. But there are an infinite number of functions which return a value of type `a` for any given `a`, not just the `a` constructor.

How about this?

```haskell
get :: A -> String -> Accessible -> Maybe a
```

This would mean we're treating single-letter identifiers—rather than lower-case identifiers—as type variables. This is a deviation from Haskell, but the correspondence between `A` and `a` is quite nice.

tel:
One thing that might work is

    TypeRep a -> String -> Accessible -> Maybe a

This is more exact since the values like `Number` are value-level representatives of types, not types themselves (this is confusing nomenclature in Javascript, since the word "type" is so overloaded, but it is certainly accurate).

We can imagine an indexed family of types of kind `TypeRep : * -> *` where for each type `a`, `TypeRep a` has exactly one value—a singleton. This value represents the types at the value level.

davidchambers:
@tel, are you suggesting that `Number` is the sole value of the `TypeRep Number` type? I want to make sure I've understood your comment. :)

tel:
Yeah, exactly! We cannot instantiate this type directly in Haskell, but it'd look a bit like this

```haskell
data TypeRep a where
  Number :: TypeRep Number
  Array :: TypeRep a -> TypeRep (Array a)
  ...
```

The difficulty is that in Haskell we need to ultimately close the type (end it after finitely many variants have been specified) yet we're allowed to continue specifying new types afterward. Worse, there's no way to get values at `TypeRep (TypeRep X)`, e.g.

But ignoring those issues that's the exact idea.

davidchambers:
Okay. I like the idea. Do these changes seem right?

```diff
- is :: Type -> a -> Boolean
+ is :: Type a -> b -> Boolean

- Maybe :: Type
+ Maybe :: Type Maybe

- Maybe#type :: Type
+ Maybe#type :: Type Maybe

- Either :: Type
+ Either :: Type Either

- Either#type :: Type
+ Either#type :: Type Either

- pluck :: String -> [{String: *}] -> [Maybe *]
+ pluck :: Type a -> String -> [Accessible] -> [Maybe a]

- get :: String -> Object -> Maybe *
+ get :: Type a -> String -> Accessible -> Maybe a

- gets :: [String] -> Object -> Maybe *
+ gets :: Type a -> [String] -> Accessible -> Maybe a
```

Unless there's a compelling reason to use `TypeRef` rather than `Type`, I'd prefer to stick with the latter since we're already using it (`S.Just(42).type`, for example, evaluates to `S.Maybe`).

tel:
I suppose those work, but using `Type` directly feels very confusing to me. `Maybe :: Type Maybe` at the value level certainly isn't a type after all! Generally that's looking good though.

davidchambers:
Yeah, `TypeRef` *is* much better. I'll update the type signatures but leave the `type` property alone for now. We can consider renaming it `typeref` or `typeRef` in a separate pull request.

index.js
```diff
@@ -59,6 +59,30 @@
...
+//. When provided as the first argument to [`is`](#is), though, `Number` is
+//. really the value-level representative of the Number type.
+//.
+//. Sanctuary uses the TypeRep pseudotype to describe type representatives.
+//. For example:
+//.
+//.     Number :: TypeRep Number
+//.
+//. `Number` is the sole inhabitant of the TypeRep Number type.
+//.
```
davidchambers: Does this convey the idea, @tel? Are there any wording changes you think would make this section clearer?

davidchambers:
I've changed occurrences of `Type` in type signatures to `TypeRep` (not `TypeRef` as I mistakenly wrote in earlier comments).

Setting aside type signatures for a moment, do others agree with the API change in question? Do you find the rationale compelling, @svozza?

index.js
tel: That sounds great! Two minor points alone

* "... `Number` is really the..." may be better said as "`Number` acts as the..."
* `TypeRep` can be called a "type function" or "type constructor" if you like instead of invoking the term "pseudotype". Taken together, `TypeRep` might also be called a family of types or a fiber bundle... though that last term is quite rare!

:+1:

davidchambers: Thank you for your feedback. :)

> "... `Number` is really the..." may be better said as "`Number` acts as the..."

I was about to `:+1:` when I realized that if a representative is one who acts on behalf of another, _acts as the representative of X_ can be expanded to _acts as the one who acts on behalf of X_. Perhaps I'm overthinking things. :)

> `TypeRep` can be called a "type function" or "type constructor" if you like instead of invoking the term "pseudotype".

Type constructor sounds good to me. I used the term "pseudotype" in #62:

> The Int pseudotype represents integers in the range [-2^31 .. 2^31). It is a pseudotype because each Int is represented by a Number value. Sanctuary's run-time type checking asserts that a valid Number value is provided wherever an Int value is required.

I wanted a term to describe a thing which is a type in the sense of a set of values but not in the sense of `typeof` or `instanceof`. I reused the term in this pull request to describe Accessible, and later TypeRep.

Describing TypeRep as a type constructor makes its role clearer. We can use the following subheadings:

    ### Accessible pseudotype

    ### TypeRep type constructor

svozza:
I definitely like this, the main issues I used to have with `gets` was that often there'd be a chance that they'd just be wrapping `null`s, which meant I actually avoided using it. Big +1 from me.

index.js
davidchambers: I decided to use `### Type representatives` as the subheading instead.

PR #65, opened 2015-07-28
ramda@0.17.x

davidchambers:
- ramda/ramda#1277
  - ramda/ramda#1290

> :warning: `R.match` now returns `[]` rather than `null` when there are no matches ramda/ramda#1239


benperez:
:dromedary_camel:

Issue #66, opened 2015-07-30
add S.takeLast and S.dropLast


davidchambers:
We should define sanctified versions of [`R.takeLast`][1] and [`R.dropLast`][2].

If you're interested in contributing to Sanctuary for the first time, this is a well-defined, self-contained problem well-suited to a first contribution. :)

[1]: http://ramdajs.com/docs/#takeLast
[2]: http://ramdajs.com/docs/#dropLast

davidchambers:
Added in #67

PR #67, opened 2015-07-30
add takeLast and dropLast

pvrnav:
Adds functions referred to by #66 to bring Sanctuary up to date with Ramda 0.17.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1144,6 +1144,27 @@
     return n < 0 || negativeZero(n) ? Nothing() : slice(0, n, xs);
   });
 
+  //# takeLast :: Number -> [a] -> Maybe [a]
+  //.
+  //. Returns Just the last N elements of the given collection if N is
+  //. greater than or equal to zero and less than or equal to the length
+  //. of the collection; Nothing otherwise. Supports Array, String, and
+  //. any other collection type which provides a `slice` method.
+  //.
+  //. ```javascript
+  //. > S.takeLast(2, ['a', 'b', 'c', 'd', 'e'])
+  //. Just(["d", "e"])
+  //.
+  //. > S.takeLast(4, 'abcdefg')
+  //. Just("defg")
+  //.
+  //. > S.takeLast(4, ['a', 'b', 'c'])
+  //. Nothing()
+  //. ```
+  S.takeLast = def('takeLast', [Number, Accessible], function(n, xs) {
+    return n < 0 || negativeZero(n) ? Nothing() : slice(-n, -0, xs);
+  });
+
   //# drop :: Number -> [a] -> Maybe [a]
   //.
   //. Returns Just all but the first N elements of the given collection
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1144,6 +1144,27 @@
     return n < 0 || negativeZero(n) ? Nothing() : slice(0, n, xs);
   });
 
+    //# takeLast :: Number -> [a] -> Maybe [a]
```
davidchambers: :triangular_ruler:

test/index.js
```diff
@@ -1652,6 +1652,62 @@ describe('list', function() {
...
+
+    it('returns a Nothing if n is greater than collection length', function() {
+      eq(S.takeLast(6, ['a', 'b', 'c', 'd', 'e']), S.Nothing());
+      eq(S.takeLast(6, 'abcde'), S.Nothing());
+    });
+
+    it('returns a Nothing if n is negative', function() {
+      eq(S.takeLast(-0, ['a', 'b', 'c', 'd', 'e']), S.Nothing());
+      eq(S.takeLast(-1, ['a', 'b', 'c', 'd', 'e']), S.Nothing());
+      eq(S.takeLast(-0, 'abcdefg'), S.Nothing());
```
davidchambers: Let's use `'abcde'` here as well. The difference between this string literal and the one on the following line is a distraction – it's the number literals that are important.

test/index.js
```diff
@@ -1708,6 +1764,62 @@ describe('list', function() {
...
+      eq(S.dropLast(new Number(-0), ['a', 'b', 'c', 'd', 'e']), S.Nothing());
+      // jshint +W053
+    });
+
+    it('returns an empty collection if n is equal to collection length', function() {
+      eq(S.dropLast(5, ['a', 'b', 'c', 'd', 'e']), S.Just([]));
+      eq(S.dropLast(5, 'abcde'), S.Just(''));
+    });
+
+    it('returns Just the first three elements from the collection', function() {
```
davidchambers: I find this description a bit odd (though it is an apt description of `S.take(3)`). Could you reword it?

pvrnav: Do you have any suggestions? The best alternative I can think of is "Returns Just the collection less the last two elements".

davidchambers: We could coalesce this section, the preceding section, and the following section into something as these lines:

```javascript
eq(S.dropLast(0, [1, 2, 3]), S.Just([1, 2, 3]));
eq(S.dropLast(1, [1, 2, 3]), S.Just([1, 2]));
eq(S.dropLast(2, [1, 2, 3]), S.Just([1]));
eq(S.dropLast(3, [1, 2, 3]), S.Just([]));
eq(S.dropLast(4, [1, 2, 3]), S.Nothing());
eq(S.dropLast(0, 'abc'), S.Just('abc'));
eq(S.dropLast(1, 'abc'), S.Just('ab'));
eq(S.dropLast(2, 'abc'), S.Just('a'));
eq(S.dropLast(3, 'abc'), S.Just(''));
eq(S.dropLast(4, 'abc'), S.Nothing());
```

The description of this block would then be a description of the function in general terms, rather than for a specific N.

pvrnav: I've updated this, take a look.

davidchambers: The new description is good. I'd like to show every N in [0 .. xs.length + 1] as in my previous comment. Perhaps it's just the way my brain is wired, but I find the quickest way to understand a function's behaviour is to be shown a contiguous range of inputs from just beyond the lower bound to just beyond the upper bound.

pvrnav: Updated accordingly.

davidchambers: We're now testing 6, 5, 2, and 0. For a collection of length five I'd like to test 0, 1, 2, 3, 4, 5, and 6. This may be a bit tedious, so we could choose a smaller collection length. If we choose three, for example, we could then test 0, 1, 2, 3, and 4.

pvrnav: Okay, changed them to be of length 3, so we test 0,1,2,3,4.

davidchambers: Cool. Grouping the list assertions and the string assertions would make the progression clearer. At the moment it's not immediately obvious that the input lists are equal because the string assertions are interspersed.

Also, let's update the `S.takeLast` tests to match. :)

pvrnav: Changed!

davidchambers:
:deciduous_tree: Thanks!

PR #69, opened 2015-08-13
Adds Maybe#reduce.

scott-christopher:
Not sure what I'm doing wrong, but despite the tests passing when run manually, running `make test` reports:
```
sanctuary/index.js:110
    return x != null && Object(x) instanceof type;
                                             ^
TypeError: Expecting a function in instanceof check, but got undefined
    at _is (sanctuary/index.js:110:46)
    at Just.<anonymous> (sanctuary/index.js:164:21)
    at Just.map (sanctuary/index.js:116:45)
    at [stdin]:63:25
    at Object.exports.runInThisContext (vm.js:74:17)
    at Object.<anonymous> ([stdin]-wrapper:6:22)
    at Module._compile (module.js:460:26)
    at evalScript (node.js:431:25)
    at Socket.<anonymous> (node.js:164:11)
    at Socket.emit (events.js:129:20)
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -258,8 +258,8 @@
   //. The Maybe type represents optional values: a value of type `Maybe a` is
   //. either a Just whose value is of type `a` or a Nothing (with no value).
   //.
-  //. The Maybe type satisfies the [Monoid][], [Monad][], and [Extend][]
-  //. specifications.
+  //. The Maybe type satisfies the [Monoid][], [Monad][], [Foldable][], and
+  //. [Extend][] specifications.
   var Maybe = S.Maybe = function Maybe() {
     throw new Error('Cannot instantiate Maybe');
   };
@@ -441,6 +441,21 @@
   //. ```
   Maybe.prototype.of = def('Maybe#of', [a], Maybe.of);
 
+  //# Maybe#reduce :: Maybe a ~> (b -> a -> b) -> b -> b
+  //.
+  //. Takes a function and an initial value, returning:
+  //. * the initial value if `this` is a Nothing; otherwise
+  //. * the result of applying the function to the initial value and the value
+  //.   of this Just.
+  //.
+  //. ```javascript
+  //. > S.Nothing().reduce(R.add, 10)
+  //. 10
+  //.
+  //. > S.Just(5).reduce(R.add, 10)
+  //. 15
+  //. ```
+
   //# Maybe#toBoolean :: Maybe a ~> Boolean
   //.
   //. Returns `false` if `this` is a Nothing; `true` if `this` is a Just.
@@ -505,6 +520,10 @@
   //  Nothing#map :: Maybe a ~> (a -> b) -> Maybe b
   Nothing.prototype.map = def('Nothing#map', [Function], self);
 
+  //  Nothing#reduce :: Maybe a ~> (b -> a -> b) -> b -> b
+  Nothing.prototype.reduce =
+    def('Nothing#reduce', [Function, a], function(f, z) { return z; });
+
   //  Nothing#toBoolean :: Maybe a ~> Boolean
   Nothing.prototype.toBoolean = def('Nothing#toBoolean', [], R.always(false));
 
@@ -559,6 +578,11 @@
   //  Just#map :: Maybe a ~> (a -> b) -> Maybe b
   Just.prototype.map = def('Just#map', [Function], function(f) {
     return Just(f(this.value));
+  });
+
+  //  Just#reduce :: Maybe a ~> (b -> a -> b) -> b -> b
+  Just.prototype.reduce = def('Just#reduce', [Function, a], function(f, z) {
+    return f(z, this.value);
   });
 
   //  Just#toBoolean :: Maybe a ~> Boolean
--- a/test/index.js
+++ b/test/index.js
@@ -271,6 +271,16 @@ describe('maybe', function() {
                             'Function as its first argument; received null'));
     });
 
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -441,6 +441,20 @@
...
+  //.
+  //. Takes a function and an initial value, returning the result of applying
+  //. the function to the initial value and the value of the Just. If `this`
+  //. is a Nothing, then the initial value is simply returned instead.
+  //.
+  //. ```javascript
+  //. > S.Nothing().reduce(R.add, 10)
+  //. 10
+  //.
+  //. > S.Just(5).map(R.add, 10)
```
davidchambers: This should be `reduce` rather than `map`.

davidchambers:
Thanks for the pull request, @scott-christopher! One of the doctests is incorrect, which is causing the test failure. I'm sorry that the error output is not helpful. If you look in the makefile you'll see a hacky Unix pipeline. I plan to integrate [doctest][1] at some point, which will give us descriptive error messages.


index.js
```diff
@@ -441,6 +441,20 @@
   //. ```
   Maybe.prototype.of = def('Maybe#of', [a], Maybe.of);
 
+  //# Maybe#reduce :: Maybe a ~> (b -> a -> b) -> b -> b
+  //.
+  //. Takes a function and an initial value, returning the result of applying
+  //. the function to the initial value and the value of the Just. If `this`
+  //. is a Nothing, then the initial value is simply returned instead.
```
davidchambers: Perhaps it would be clearer to make each case a list item:

> Takes a function and an initial value of any type, and returns:
>
>   - the initial value if `this` is a Nothing; otherwise
>   - the result of applying the function to the initial value and this Just's value.

Do you think it's accurate to describe the value as an *initial* value?

index.js
```diff
@@ -505,6 +519,10 @@
   //  Nothing#map :: Maybe a ~> (a -> b) -> Maybe b
   Nothing.prototype.map = def('Nothing#map', [Function], self);
 
+  //  Nothing#reduce :: Maybe a ~> (b -> a -> b) -> b -> b
+  Nothing.prototype.reduce = def('Nothing#reduce', [Function, a],
+                                 function(f, z) { return z; });
```
davidchambers: I prefer this line wrapping:

```javascript
//  Nothing#reduce :: Maybe a ~> (b -> a -> b) -> b -> b
Nothing.prototype.reduce =
def('Nothing#reduce', [Function, a], function(f, z) { return z; });
```

test/index.js
```diff
@@ -271,6 +271,16 @@ describe('maybe', function() {
...
     });
 
+    it('provides a "reduce" method', function() {
+      eq(S.Nothing().reduce.length, 2);
+      eq(S.Nothing().reduce(function(a, b) { return a + b; }, 10), 10);
+
+      assert.throws(function() { S.Nothing().reduce(null); },
+        errorEq(TypeError,
+          '‘Nothing#reduce’ requires a value of type ' +
+          'Function as its first argument; received null'));
```
davidchambers: I find this nesting makes it easier to see what's being applied to what, since the arguments are aligned:

```javascript
assert.throws(function() { S.Nothing().reduce(null); },
              errorEq(TypeError,
                      '‘Nothing#reduce’ requires a value of type ' +
                      'Function as its first argument; received null'));
```

I've been referring to this as Lisp-style indentation, though I'm not sure this is accurate. :)

davidchambers:
Could you update this section to include Foldable?

> ```javascript
> //. The Maybe type satisfies the [Monoid][], [Monad][], and [Extend][]
> //. specifications.
> ```

scott-christopher:
I've pushed an update based on your feedback.

index.js
scott-christopher: I typically describe it as an *initial accumulator value*, but there isn't a whole lot of accumulating going on in the implementation for Maybe.

index.js
```diff
@@ -258,8 +258,8 @@
   //. The Maybe type represents optional values: a value of type `Maybe a` is
   //. either a Just whose value is of type `a` or a Nothing (with no value).
   //.
-  //. The Maybe type satisfies the [Monoid][], [Monad][], and [Extend][]
-  //. specifications.
+  //. The Maybe type satisfies the [Monoid][], [Monad][], [Foldable][] and
+  //. [Extend][] specifications.
```
davidchambers: Mind preserving the Oxford comma? :stuck_out_tongue_winking_eye:

scott-christopher: Done,

index.js
```diff
@@ -441,6 +441,21 @@
   //. ```
   Maybe.prototype.of = def('Maybe#of', [a], Maybe.of);
 
+  //# Maybe#reduce :: Maybe a ~> (b -> a -> b) -> b -> b
+  //.
+  //. Takes a function and an initial value, returning:
+  //. * the initial value if this is a Nothing; otherwise
+  //. * the result of applying the function to the initial value and the value
+  //.   of the Just.
```
davidchambers: Two last minor points:

  - in the first list item, "this" should be wrapped in backticks; and
  - in the second list item, do you think we should write "the value of <del>the</del> <ins>this</ins> Just"?

I'm sorry for nitpicking, but I do consider the wording of these descriptions important.

scott-christopher: > I'm sorry for nitpicking, but I do consider the wording of these descriptions important.

House rules – nitpick away :) I've updated as suggeted.

davidchambers:
:deciduous_tree: Thanks again!

PR #76, opened 2015-09-04
add maybeToEither

kedashoe:
I basically just followed the `either` tests except for the 'is curried' part, as I noticed it was quite extensive but none of the other 'is curried' tests were.

I decided against including "eitherIfElse" for now. I pretty much only created it because it seemed like a decent idea and you can use it to define `maybeToEither`. If you really like the idea of, I'm happy to submit a PR, but I figured I'd just let it hang out in my code for a while and see how much I actually use it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -972,6 +972,24 @@
     return either instanceof Left ? l(either.value) : r(either.value);
   });
 
+  //# maybeToEither :: a -> Maybe b -> Either a b
+  //.
+  //. Takes a value of any type and a Maybe, and returns an Either.
+  //. If the second argument is a Nothing, a Left containing the first
+  //. argument is returned. If the second argument is a Just, a Right
+  //. containing the Just's value is returned.
+  //.
+  //. ```javascript
+  //. > S.maybeToEither('Expecting an integer', S.parseInt(10, 'xyz'))
+  //. Left("Expecting an integer")
+  //.
+  //. > S.maybeToEither('Expecting an integer', S.parseInt(10, '42'))
+  //. Right(42)
+  //. ```
+  S.maybeToEither = def('maybeToEither', [a, Maybe], function(x, maybe) {
+    return maybe instanceof Nothing ? Left(x) : Right(maybe.value);
+  });
+
   //. ### Control
 
   var methodMissing = format('{repr} does not have {} {quote} method');
--- a/test/index.js
+++ b/test/index.js
@@ -1163,6 +1163,36 @@ describe('either', function() {
 
   });
 
+  describe('maybeToEither', function() {
+
+    it('is a binary function', function() {
+      eq(typeof S.either, 'function');
+      eq(S.maybeToEither.length, 2);
+    });
+
+    it('type checks its arguments', function() {
+      assert.throws(function() { S.maybeToEither('left', 1); },
+                    errorEq(TypeError,
+                            '‘maybeToEither’ requires a value of type Maybe ' +
+                            'as its second argument; received 1'));
+    });
+
+    it('returns a Left of its first argument when the second is Nothing', function() {
+      eq(S.maybeToEither('error msg', S.Nothing()), S.Left('error msg'));
+    });
+
+    it('returns a Right of the value contained in the Just ' +
+       'when the second argument is a Just', function() {
+      eq(S.maybeToEither('error msg', S.Just(42)), S.Right(42));
+    });
+
+    it('is curried', function() {
+      eq(S.maybeToEither(0).length, 1);
+      eq(S.maybeToEither(0)(S.Just(42)), S.Right(42));
+    });
+
+  });
+
 });
 
 describe('control', function() {


davidchambers:
Very nice!

> [I'll let "eitherIfElse"] hang out in my code for a while and see how much I actually use it.

Good idea. :)

index.js
```diff
@@ -972,6 +972,22 @@
...
+  //.
+  //. ```javascript
+  //. > S.maybeToEither('Expecting an integer', S.parseInt(10, 'xyz'))
+  //. Left("Expecting an integer")
+  //.
+  //. > S.maybeToEither('Expecting an integer', S.parseInt(10, '42'))
+  //. Right(42)
+  //. ```
+  S.maybeToEither = def('maybeToEither', [a, Maybe], function(x, maybe) {
+    return maybe instanceof Nothing ? Left(x) : maybe.chain(Right);
```
davidchambers: `maybe.chain(Right)` seems like an abuse of the `chain` method. According to the [FL spec][1]:

> `f` must return a value of the same Chain

I'd prefer to use `Right(maybe.value)` here.


[1]: https://github.com/fantasyland/fantasy-land#chain-method

kedashoe: I was wondering about that. I think I originally had `Right(maybe.chain(R.identity))`. Didn't realize I was allowed to reach into the `maybe` :)

> f must return a value of the same Chain

Interesting, since it doesn't have to return a `chain`able value at all, right (as when chaining with identity)?

kedashoe: updated

index.js
```diff
@@ -972,6 +972,22 @@
     return either instanceof Left ? l(either.value) : r(either.value);
   });
 
+  //# maybeToEither :: a -> Maybe b -> Either a b
+  //.
+  //. Converts a Maybe value to an Either value, using the supplied
+  //. parameter as the Left value if the Maybe is Nothing.
```
davidchambers: This is a good description, but the following would be more consistent with the (verbose!) style used elsewhere:

```javascript
  //. Takes a value of any type and a Maybe, and returns an Either.
  //. If the second argument is a Nothing, a Left containing the first
  //. argument is returned. If the second argument is a Just, a Right
  //. containing the Just's value is returned.
```

test/index.js
```diff
@@ -1163,6 +1163,41 @@ describe('either', function() {
...
+    it('type checks its arguments', function() {
+      assert.throws(function() { S.maybeToEither('left', 1); },
+                    errorEq(TypeError,
+                            '‘maybeToEither’ requires a value of type Maybe ' +
+                            'as its second argument; received 1'));
+
+      assert.throws(function() { S.maybeToEither(R.__, 1)('left'); },
+                    errorEq(TypeError,
+                            '‘maybeToEither’ requires a value of type Maybe ' +
+                            'as its second argument; received 1'));
```
davidchambers: Let's drop this assertion. The `S.maybeToEither('left', 1)` test should be sufficient.

index.js
```diff
@@ -972,6 +972,22 @@
...
+  //.
+  //. ```javascript
+  //. > S.maybeToEither('Expecting an integer', S.parseInt(10, 'xyz'))
```
kedashoe: Do you think the `parseInt` is too much for the example here? Seems like most of the examples in these descriptions are very simple, I can just be explicit with `Nothing` and `Just(42)`

index.js
kedashoe: I can't take credit, just copied it from Haskell's docs haha

index.js
davidchambers: > Didn't realize I was allowed to reach into the `maybe`

We do so (in Sanctuary) so others don't need to. :)

> Interesting, since it doesn't have to return a `chain`able value at all, right (as when chaining with identity)?

In this case, though, `R.identity` *does* return a value of the same Chain (Maybe):

```javascript
> R.chain(R.identity, S.Just(S.Just(42)))
Just(42)
```

index.js
davidchambers: I like your examples. They nicely demonstrate the intended purpose of the function.

kedashoe: Sounds good, updated with the other changes.

(Do I need to comment, or do you as an owner get an email when I push? I don't maintain any repos, just curious)

davidchambers: > Do I need to comment, or do you as an owner get an email when I push?

I don't get notified when you push, so it's a good idea to comment. At Plaid we `:zap:` pull request comments after addressing them. :)

davidchambers:
:deciduous_tree:

test/index.js
```diff
@@ -1163,6 +1163,36 @@ describe('either', function() {
 
   });
 
+  describe('maybeToEither', function() {
+
+    it('is a binary function', function() {
+      eq(typeof S.either, 'function');
```
davidchambers: I spotted this typo yesterday. Fixed in 72da0c9693c32774838aea6bcb68e48414ef2fb7. :)

kedashoe: ah, the copypasta got me, thanks for the fix!

PR #78, opened 2015-09-08
maybe: add S.catMaybes

### LINKED ISSUE #72: catMaybes :: [Maybe a] -> [a]
<https://hackage.haskell.org/package/base-4.8.1.0/docs/Data-Maybe.html#v:catMaybes>:

> The `catMaybes` function takes a list of `Maybe`s and returns a list of all the `Just` values.

davidchambers:
Closes #72

`maybe :: b -> (a -> b) -> Maybe a -> b` proved useful in defining this function. I'll rebase this patch once #77 has been merged.

### DIFF
--- a/index.js
+++ b/index.js
@@ -682,9 +682,20 @@
   //. > S.maybe(0, R.length, S.Nothing())
   //. 0
   //. ```
-  S.maybe = def('maybe', [a, Function, Maybe], function(x, f, maybe) {
+  var maybe = S.maybe =
+  def('maybe', [a, Function, Maybe], function(x, f, maybe) {
     return fromMaybe(x, maybe.map(f));
   });
+
+  //# catMaybes :: [Maybe a] -> [a]
+  //.
+  //. Takes a list of Maybes and returns a list containing each Just's value.
+  //.
+  //. ```javascript
+  //. > S.catMaybes([S.Just('foo'), S.Nothing(), S.Just('baz')])
+  //. ["foo", "baz"]
+  //. ```
+  S.catMaybes = def('catMaybes', [Accessible], R.chain(maybe([], R.of)));
 
   //# encase :: (* -> a) -> (* -> Maybe a)
   //.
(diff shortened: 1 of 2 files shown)


svozza:
That seems like a really elegant implementation but I don't really get why it works. Would you mind explaining it a bit?

davidchambers:
> Would you mind explaining it a bit?

Not at all. :)

Here are the relevant type signatures, for reference:

    R.chain :: (a -> [b]) -> [a] -> [b]  -- we can ignore the more general type
    R.of :: a -> [a]
    maybe :: b -> (a -> b) -> Maybe a -> b

The second argument to `maybe` has type `a -> b`. We provide `R.of`—which has type `a -> [a]`—in this position, so we replace `b` with `[a]`:

    maybe :: b -> (a -> b) -> Maybe a -> b
    maybe :: [a] -> (a -> [a]) -> Maybe a -> [a]
    maybe([], R.of) :: Maybe a -> [a]

This makes sense because `maybe([], R.of)` is a function which gives for `[]` for `Nothing()` and `[x]` for `Just(x)`.

We provide `maybe([], R.of)` as the first argument to `R.chain`. In the signature of `R.chain` we replace `a` with `Maybe a` and `b` with `a`:

    R.chain :: (a -> [b]) -> [a] -> [b]
    R.chain :: (Maybe a -> [a]) -> [Maybe a] -> [a]
    R.chain(maybe([], R.of)) :: [Maybe a] -> [a]
    S.catMaybes :: [Maybe a] -> [a]

Here's an example:

```javascript
//  maybeToList :: Maybe a -> [a]
var maybeToList = maybe([], R.of);

R.map(maybeToList, [S.Just('foo'), S.Nothing(), S.Just('baz')]);
// => [["foo"], [], ["baz"]]

R.chain(maybeToList, [S.Just('foo'), S.Nothing(), S.Just('baz')]);
// => ["foo", "baz"]
```

Thank you for asking me to explain this. Doing so has deepened my understanding.

svozza:
Thanks David, enlightening as always! :)

PR #79, opened 2015-09-08
maybe: add S.mapMaybe

### LINKED ISSUE #73: mapMaybe :: (a -> Maybe b) -> [a] -> [b]
<https://hackage.haskell.org/package/base-4.8.1.0/docs/Data-Maybe.html#v:mapMaybe>:

> The `mapMaybe` function is a version of `map` which can throw out elements. In particular, the functional argument returns something of type `Maybe b`. If this is `Nothing`, no element is added on to the result list. If it is `Just b`, then `b` is included in the result list.

davidchambers:
Closes #73

`catMaybes :: [Maybe a] -> [a]` proved useful in defining this function. I'll rebase this patch once #78 has been merged.

### DIFF
--- a/index.js
+++ b/index.js
@@ -695,7 +695,25 @@
   //. > S.catMaybes([S.Just('foo'), S.Nothing(), S.Just('baz')])
   //. ["foo", "baz"]
   //. ```
-  S.catMaybes = def('catMaybes', [Accessible], R.chain(maybe([], R.of)));
+  var catMaybes = S.catMaybes =
+  def('catMaybes', [Accessible], R.chain(maybe([], R.of)));
+
+  //# mapMaybe :: (a -> Maybe b) -> [a] -> [b]
+  //.
+  //. Takes a function and a list, applies the function to each element of
+  //. the list, and returns a list of "successful" results. If the result of
+  //. applying the function to an element of the list is a Nothing, the result
+  //. is discarded; if the result is a Just, the Just's value is included in
+  //. the output list.
+  //.
+  //. In general terms, `mapMaybe` filters a list while mapping over it.
+  //.
+  //. ```javascript
+  //. > S.mapMaybe(S.head, [[], [1, 2, 3], [], [4, 5, 6], []])
+  //. [1, 4]
+  //. ```
+  S.mapMaybe =
+  def('mapMaybe', [Function, Accessible], R.compose(catMaybes, R.map));
 
   //# encase :: (* -> a) -> (* -> Maybe a)
   //.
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -682,10 +682,29 @@
...
 
+  //  catMaybes :: [Maybe a] -> [a]
+  var catMaybes = def('catMaybes', [Accessible], R.chain(maybe([], R.of)));
+
+  //# mapMaybe :: (a -> Maybe b) -> [a] -> [b]
+  //.
+  //. Takes a function and a list, and returns a list containing the value
+  //. of each Just resulting from the application of the function to each
+  //. element of the input list.
+  //.
```
kedashoe: This description was a little confusing to me at first.. maybe mention that `Nothing`s are rejected in the description?

davidchambers: Good idea. I'll reword it tomorrow. :)

davidchambers: How does this sound, @kedashoe?

```javascript
//. Takes a function and a list, applies the function to each element of
//. the list, and returns a list of "successful" results. If the result of
//. applying the function to an element of the list is a Nothing, the result
//. is discarded; if the result is a Just, the Just's value is included in
//. the output list.
```

It's 11:56pm, so I (just) kept my word. :)

kedashoe: :+1:

benperez: I know I'm chiming in a little late here but I like the Haskell description:
> The mapMaybe function is a version of map which can throw out elements. In particular, the functional argument returns something of type Maybe b. If this is Nothing, no element is added on to the result list. If it is Just b, then b is included in the result list.

benperez:
I'm fine with either description so :dromedary_camel:. I actually needed to use this function in an extractor yesterday!

index.js
davidchambers: I, too, like the description from the Haskell library. The first sentence sets the scene nicely. The main problem I have with using this definition verbatim is that it's not consistent with the form used to describe other Sanctuary functions: _Takes a foo and a bar, and returns a baz._ I intended the second paragraph to give the high-level description:

```javascript
//. In general terms, `mapMaybe` filters a list while mapping over it.
```

I have one other problem with the description from the Haskell library (a minor one).

> If this is `Nothing`, no element is added on to the result list.

I don't like the use of "added on to", as it describes an implementation detail not relevant to the caller. It's nice to ignore the intermediate states between input and output. Functions are applied atomically in mathematics, and we can think of the application of a pure JavaScript functions in this way, even though we know it to be untrue.

benperez: Those are fair points. I'm on board with your revised description.

PR #80, opened 2015-09-09
add List pseudotype

davidchambers:
This is another minor improvement to the descriptiveness of our type errors.

My only concern is the word used to describe values such as `[1, 2, 3]` and `'foo'`. Referring to these as lists is fine, but the way in which we detect such values is by consulting the `length` property. As @CrossEye is fond of saying, an actual List type would be a linked list. The length of a linked list is usually determined via a recursive function, so if one day we were to provide an actual List type, a value of that type may not be considered a member of the List pseudotype!

Perhaps "ArrayLike" is a better name for the pseudotype. What do you think?

### DIFF
--- a/index.js
+++ b/index.js
@@ -76,6 +76,11 @@
 //. Sanctuary's run-time type checking asserts that a valid Number value is
 //. provided wherever an Integer value is required.
 //.
+//. ### List pseudotype
+//.
+//. The List pseudotype represents non-Function values with numeric `length`
+//. properties greater than or equal to zero, such as `[1, 2, 3]` and `'foo'`.
+//.
 //. ### Type representatives
 //.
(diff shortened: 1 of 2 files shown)


benperez:
I think that differentiating between linked lists and arrays here is unnecessary. To me the difference between linked lists and arrays is that arrays have O(1) random access while linked lists are better at inserting elements in the list at positions other than the end. There is nothing stopping a linked list implementation from manually keeping track of a length property and exposing it.

I think that it's fine to call this a List.

benperez:
:dromedary_camel: from me, this seems like a helpful improvement!

davidchambers:
@benperez has convinced me to stick with "List" rather than switch to "ArrayLike".

PR #85, opened 2015-09-14
circle: run tests on Node v0.12.x in addition to Node v0.10.x

davidchambers:
I'm pleased to have learnt how to do this (from one of the members of the CircleCI support team).


benperez:
:dromedary_camel:

PR #86, opened 2015-09-16
either: add S.encaseEither

davidchambers:
I've been meaning to add this function for some time. @lucamartinetti asked me today whether such a function exists, prompting this pull request.

My only question is which of the following is the best combination of names:

| `(* -> a) -> (* -> Maybe a)`  | `(* -> a) -> (* -> Either Error a)` |
| ----------------------------- | ----------------------------------- |
| `encase`                      | `encaseEither`                      |
| `encaseMaybe`                 | `encase`                            |
| `encaseMaybe`                 | `encaseEither`                      |

I don't have a strong inclination in any direction, so I'm inclined to leave `S.encase` alone. Does anyone have a strong preference for one of the other options?

### DIFF
--- a/index.js
+++ b/index.js
@@ -729,6 +729,8 @@
   //. applying `f` to the same arguments: if this succeeds, `g` returns Just
   //. the result; otherwise `g` returns Nothing.
   //.
+  //. See also [`encaseEither`](#encaseEither).
+  //.
   //. ```javascript
   //. > S.encase(eval)('1 + 1')
   //. Just(2)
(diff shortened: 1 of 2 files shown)


davidchambers:
Thinking about this further as I rode home this evening I reached the conclusion that `(* -> a) -> (* -> Either Error a)` is not sufficiently flexible. Consider Kleisli composition:

```haskell
f :: a -> Either x b
g :: b -> Either x c
h :: a -> Either x c
h = f >=> g
```

In order to compose `f` and `g`, both must return values of type `Either x _` for some type `x`. If instead `g` were to return values of type `Either y c`, say, `h` would not be well typed, since we would not know whether to expect a Left resulting from the application of `h` to contain an `x` or a `y`.

The function as it currently stands could be used in conjunction with functions which return values of type `Either Error _`, but not with functions which return values of type `Either String _`.

Instead, we could provide `encaseEither :: (Error -> a) -> (* -> b) -> (* -> Either a b)`. One could then use:

  - `S.encaseEither(R.identity)` to produce results of type `Either Error _`; or
  - `S.encaseEither(R.prop('message'))` to produce results of type `Either String _`.

davidchambers:
I've updated the pull request to provide the more general function.

benperez:
After some discussion with @davidchambers IRL I agree that we don't want to lock ourselves into always having to chain together functions that return `Either Error a`. I'm on board with the `R.I` approach until we figure out what the really common use cases for this are.

:church:

PR #87, opened 2015-09-18
string: add S.words, S.unwords, S.lines, and S.unlines

davidchambers:
Supersedes ramda/ramda#1382

See [Data.String][1] from the Haskell Prelude.

[1]: https://hackage.haskell.org/package/base-4.8.1.0/docs/Data-String.html#g:1

### DIFF
--- a/index.js
+++ b/index.js
@@ -1643,6 +1643,65 @@
     return R.map(R.map(toMaybe), toMaybe(s.match(pattern)));
   });
 
+  //. ### String
+
+  //# words :: String -> [String]
+  //.
+  //. Takes a string and returns the list of words the string contains
+  //. (words are delimited by whitespace characters).
+  //.
+  //. See also [`unwords`](#unwords).
+  //.
+  //. ```javascript
+  //. > S.words(' foo bar baz ')
+  //. ["foo", "bar", "baz"]
+  //. ```
+  S.words =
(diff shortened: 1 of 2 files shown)


benperez:
I see the utility of these methods but what's the rationale for having them live in Sanctuary?

davidchambers:
> I see the utility of these methods but what's the rationale for having them live in Sanctuary?

@CrossEye made a good point in https://github.com/ramda/ramda/pull/1382#issuecomment-140272971:

> I simply don't want anyone to think -- ourselves included -- that we're trying to port over all the Haskell functions.

Making Haskell concepts and functions available to those programming in JavaScript is very much the goal of Sanctuary, as suggested by the first line of the README:

> Sanctuary is a small functional programming library inspired by Haskell and PureScript.

In the aforementioned Ramda issue, @paldepind noted that there exist two valid behaviours for a "join lines" function: it could map `['foo', 'bar', 'baz']` to `'foo\nbar\nbaz'` or to `'foo\nbar\nbaz\n'`. In the context of Ramda this is a legitimate conundrum. In the context of Sanctuary the answer is clear: name the function `unlines` and have it behave like its Haskell namesake.

These functions could one day be defined in Ramda as well, of course.

benperez:
:church:

PR #90, opened 2015-09-18
combinator: add S.I

davidchambers:
Ramda provides [`R.identity`][1], of course, but `S.I` is much shorter:

```javascript
S.encaseEither(R.identity, JSON.parse)
```

```javascript
S.encaseEither(S.I, JSON.parse)
```

Since we already provide `S.K`, it seems more fitting to name this function `I` rather than `id`.

[1]: http://ramdajs.com/docs/#identity

### DIFF
--- a/index.js
+++ b/index.js
@@ -289,6 +289,17 @@
 
   //. ### Combinator
 
+  //# I :: a -> a
+  //.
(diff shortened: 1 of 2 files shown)


benperez:
:dromedary_camel:

PR #92, opened 2015-09-19
make exported functions throw if applied to too many arguments

### LINKED ISSUE #81: check arguments.length in def function
Current behaviour:

```javascript
> S.or(true, true, true)
TypeError: Expecting a function in instanceof check, but got undefined
```

Desired behaviour:

```javascript
> S.or(true, true, true)
TypeError: ‘or’ requires two arguments; received three arguments
```

`S.or(true)(true, true)` should behave the same way.

davidchambers:
Closes #81

### DIFF
--- a/index.js
+++ b/index.js
@@ -125,6 +125,10 @@
 
   var formatters = {
     '{}': R.identity,
+    '{card}': R.ifElse(R.lte(_, 10),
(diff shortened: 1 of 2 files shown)


benperez:
:dromedary_camel:

PR #93, opened 2015-09-19
maybe, either: descriptive errors for missing "concat" methods

### LINKED ISSUE #84: descriptive "method missing" errors
For example:

```javascript
> S.Just(2).concat(S.Just(2))
TypeError: undefined is not a function
```

It would be nice to see something this instead:

```javascript
> S.Just(2).concat(S.Just(2))
TypeError: ‘Maybe#concat’ requires a semigroup as its first argument; received Just(2)
```

davidchambers:
Closes #84

I spent a surprisingly long time working this patch into its current form. I'm pleased with the current approach based on `assertMethodExists`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -262,6 +262,21 @@
     });
   };
 
+  var assertMethodExists = function(method, x) {
+    if (x == null || typeof x[method] !== 'function') {
+      throw new TypeError(format(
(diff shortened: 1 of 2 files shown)


benperez:
Very nice errors! :dromedary_camel:

PR #94, opened 2015-09-20
composition: add S.compose, S.pipe, and S.meld

davidchambers:
I proposed these functions in https://github.com/ramda/ramda/issues/1318#issuecomment-132005376. It's unlikely that Ramda will provide these exact functions but I happen to contribute to another functional programming library as well. ;)

The implementation of `compose` is wonderfully straightforward.

The implementation of `pipe` is elegant: a right fold over the list of functions, with `compose` as the reducing function and `identity` as the identity element.

I'm pleased with `meld` as the name of the third function. Is it appealing to others as well?

I'm interested to know whether this patch gets the @joneshf seal of approval; `compose` and `pipe` are intended to be easier to reason about than their Ramda counterparts.

I'd also like to hear from @CrossEye, @paldepind, and other Ramda contributors as some of these ideas should filter back into Ramda. We could, for example, revert `R.compose` and `R.pipe` to the pre-v0.17.0 behaviour and add `R.meld` (which would be a variadic to match `R.pipe`).

### DIFF
--- a/index.js
+++ b/index.js
@@ -348,6 +348,83 @@
     return x;
   });
 
+  //. ### Composition
+
+  //# compose :: (b -> c) -> (a -> b) -> a -> c
+  //.
+  //. Takes two functions assumed to be unary and a value of any type,
+  //. and returns the result of applying the first function to the result
+  //. of applying the second function to the given value.
+  //.
+  //. In general terms, `compose` performs right-to-left composition of two
+  //. unary functions.
+  //.
+  //. See also [`pipe`](#pipe).
+  //.
+  //. ```javascript
+  //. > S.compose(Math.sqrt, R.inc)(99)
+  //. 10
+  //. ```
+  var compose = S.compose =
+  def('compose', [Function, Function, a], function(f, g, x) {
+    return f(g(x));
+  });
+
+  //# pipe :: [(a -> b), (b -> c), ..., (m -> n)] -> a -> n
+  //.
+  //. Takes a list of functions assumed to be unary and a value of any type,
+  //. and returns the result of applying the sequence of transformations to
+  //. the initial value.
+  //.
+  //. In general terms, `pipe` performs left-to-right composition of a list
+  //. of functions. `pipe([f, g, h], x)` is equivalent to `h(g(f(x)))`.
+  //.
+  //. See also [`meld`](#meld).
+  //.
+  //. ```javascript
+  //. > S.pipe([R.inc, Math.sqrt, R.dec])(99)
+  //. 9
+  //. ```
+  S.pipe = def('pipe', [List, a], function(fs, x) {
+    return R.reduceRight(compose, I, fs)(x);
+  });
+
+  //# meld :: [** -> *] -> (* -> * -> ... -> *)
+  //.
+  //. Takes a list of non-nullary functions and returns a curried function
+  //. whose arity is one greater than the sum of the arities of the given
+  //. functions less the number of functions.
+  //.
+  //. The behaviour of `meld` is best conveyed diagrammatically. The following
+  //. diagram depicts the "melding" of binary functions `f` and `g`:
+  //.
+  //.               +-------+
+  //.     --- a --->|       |
+  //.               |   f   |                +-------+
+  //.     --- b --->|       |--- f(a, b) --->|       |
+  //.               +-------+                |   g   |
+  //.     --- c ---------------------------->|       |--- g(f(a, b), c) --->
+  //.                                        +-------+
+  //.
+  //. See also [`pipe`](#pipe).
+  //.
+  //. ```javascript
+  //. > S.meld([Math.pow, R.subtract])(3, 4, 5)
+  //. 76
+  //.
+  //. > S.meld([Math.pow, R.subtract])(3)(4)(5)
+  //. 76
+  //. ```
+  S.meld = def('meld', [List], function(fs) {
+    return R.curryN(1 + R.sum(R.map(R.length, fs)) - fs.length, function() {
+      var args = Array.prototype.slice.call(arguments);
+      for (var idx = 0; idx < fs.length; idx += 1) {
+        args.unshift(fs[idx].apply(this, args.splice(0, fs[idx].length)));
+      }
+      return args[0];
+    });
+  });
+
   //. ### Maybe type
 
   //# Maybe :: TypeRep Maybe
@@ -1745,7 +1822,7 @@
   //. ["foo", "bar", "baz"]
   //. ```
   S.words =
-  def('words', [String], R.compose(R.reject(R.isEmpty), R.split(/\s+/)));
+  def('words', [String], compose(R.reject(R.isEmpty), R.split(/\s+/)));
 
   //# unwords :: [String] -> String
   //.
@@ -1774,7 +1851,7 @@
   //. ```
   S.lines =
   def('lines', [String],
-      R.compose(R.match(/^(?=[\s\S]).*/gm), R.replace(/\r\n?/g, '\n')));
+      compose(R.match(/^(?=[\s\S]).*/gm), R.replace(/\r\n?/g, '\n')));
 
   //# unlines :: [String] -> String
   //.
@@ -1788,7 +1865,7 @@
   //. "foo\nbar\nbaz\n"
   //. ```
   S.unlines =
-  def('unlines', [List], R.compose(R.join(''), R.map(R.concat(_, '\n'))));
+  def('unlines', [List], compose(R.join(''), R.map(R.concat(_, '\n'))));
 
   return S;
 
--- a/test/index.js
+++ b/test/index.js
@@ -205,6 +205,107 @@ describe('combinator', function() {
 
 });
 
+describe('composition', function() {
+
+  describe('compose', function() {
+
+    it('is a ternary function', function() {
+      eq(typeof S.compose, 'function');
+      eq(S.compose.length, 3);
+    });
+
+    it('composes two functions assumed to be unary', function() {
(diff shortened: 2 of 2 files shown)


kedashoe:
There is something very pleasing about defining compose `return f(g(x));`.. but in the bigger javascript picture, it just feels like an unnecessary special case to me. If I am running some data through some functions, should I use `compose` if there are only two of them and `pipe` if there are more? And if I start with two and then realize I need more, I need to change from `compose` to `pipe`, reverse the order, and add brackets. For certain things I'm sure we are quite safe with a binary compose (`lines` is unlikely to all of a sudden need a 3rd function), but when I just have any old random data from some 3rd party API and I'm doing whatever to it, I find it's not that uncommon to be changing how many functions I'm using. Do we want `compose3`, 4, etc? I feel like I would probably just always use `pipe`, in which case the only thing `compose` is doing for me is slowing down `pipe`.

:+1:  for `meld`, excited to see how that can be used!

davidchambers:
Thank you for your thoughtful feedback, @kedashoe.

I'm not certain that a curried ternary `compose` is the best option. I have, on a small number of occasions, wanted a curried `compose` and have defined it via `R.curryN(2, R.compose)` (doctest's source contains one [example][1]). Composing _n_ functions at once is far more common, though, and is the more important use case.

> I feel like I would probably just always use `pipe`

`pipe` can be used to compose _n_ functions (for _n_ = 2, even). Perhaps, as you say, we're well advised to use `pipe` exclusively since pipelines may grow beyond the strict length limit imposed by `compose`. This is true, though this is arguably already the case with the Ramda counterparts if one considers line length. Even those who prefer `compose` over `pipe` for single-line compositions will concede that `pipe` is the better choice when composing many functions over many lines. A function initially defined via

```javascript
R.compose(..., ..., ...)
```

might become

```javascript
R.pipe(...,
       ...,
       ...,
       ...,
       ...)
```

as the pipeline grows in response to changing requirements.

I quite *like* the fact that `compose` and `pipe` are not symmetrical, as it makes it easier to decide which to use: `compose` in the rare cases in which one wishes to partially apply `compose` itself; `pipe` otherwise.

> […] in which case the only thing `compose` is doing for me is slowing down `pipe`.

Do you say this because `compose` is used in the implementation of `pipe`? If so, we could certainly make the implementation more efficient (in a subsequent pull request, perhaps). I'm most interested in discussing the types at this stage.

> :+1: for `meld`, excited to see how that can be used!

I am too. I really like the idea, but forcing it into `R.compose` and `R.pipe` made these functions behave in surprising ways. When choosing `meld` over `pipe` one should realize that the arity of each function in the pipeline is significant: after all, one would choose `meld` because one *wants* at least one of the functions in the composition to receive more than one argument.


[1]: https://github.com/davidchambers/doctest/blob/0.9.0/lib/doctest.js#L192

index.js
```diff
@@ -316,6 +316,83 @@
...
+  //.               |   f   |                +-------+
+  //.     --- b --->|       |--- f(a, b) --->|       |
+  //.               +-------+                |   g   |
+  //.     --- c ---------------------------->|       |--- g(f(a, b), c) --->
+  //.                                        +-------+
+  //.
+  //. See also [`pipe`](#pipe).
+  //.
+  //. ```javascript
+  //. > S.meld([Math.pow, R.multiply, R.subtract])(3, 4, 5, 6)
```
raine: I think there's too much going on in this example for someone to quickly understand the function based on the example alone. But maybe that's just me. Can the idea be conveyed with two functions?

davidchambers: Using two binary functions in the examples to match the diagram is a great idea! Do you think this is suitable?

```javascript
//. > S.meld([Math.pow, R.subtract])(3, 4, 5)
//. 76
```

I'll welcome suggestions for better examples. :)

paldepind:
I think these looks neat. Except `meld`. It's far from being as elegant and simple as the other functions.

I think I would prefer:

```js
// compose2 :: (c -> d) -> (a -> b -> c) -> a -> b -> d
compose2 = compose(compose, compose);
```

It is not as general. But in my opinion code is probably not very readable if you're using the more complex versions anyway.

I find this to be quite hard to understand:
```js
S.meld([Math.pow, R.multiply, R.subtract])(3, 4, 5, 6)
```

But this is understandable to me:
```js
S.compose2(Math.pow, R.multiply)(3, 4);
```

I don't like how reverse-composing is given preference when composing more than two functions. I prefer the normal composition order. I.e. I'd personally want `pipe` to be reversed. I think such an option should be available. Then the library is not opinionated about composition order.

CrossEye:
> I find this to be quite hard to understand:
>
> ```js
> S.meld([Math.pow, R.multiply, R.subtract])(3, 4, 5, 6)
> ```
> 
> But this is understandable to me:
> 
> ```js
> S.compose2(Math.pow, R.multiply)(3, 4);
> ```

I don't actually make any sense of that latter one.  It looks to me like 

```js
Math.pow(R.multiply(3, 4)); //=>
Math.pow(12); //=> NaN
```

Is that what you're looking for?  But in this proposal

```js
S.meld([Math.pow, R.multiply])(3, 4, 5); //=>
R.multiply(Math.pow(3, 4), 5); //=>
R.multiply(81, 5); //=>
405
```

seems pretty straightforward, **if** you know the arities of the functions involved.  My biggest concern with this function is that unlike a normal `pipe`/`compose`, I really have to understand the component functions, or at least their arities to make any sense of it.

    S.meld([f, g, h])(3, 4, 5, 6) //=>
    h(g(f(3), 4), 5, 6)); // or
    h(g(f(3), 4, 5), 6)); // or
    h(g(f(3, 4)), 5, 6)); // or
    h(g(f(3, 4), 5, 6))); // or
    h(g(f(3, 4, 5)), 6)); // or
    h(g(f(3, 4, 5), 6))); // or
    h(g(f(3, 4, 5, 6))); // and possibly others.

While I'm generally in favor of things that help with building complex function expressively in source code, even if they make less sense when viewed as more general-purpose functions operating on data, that list gives me some pause.

paldepind:
> I don't actually make any sense of that latter one. It looks to me like 

Sorry. That was a mistake. It should have been something like this:

```js
S.compose2(Math.sqrt, R.multiply)(3, 4);
```

index.js
davidchambers: I've updated the doctests.

benperez:
Composition is a relatively straightforward concept in Haskell thanks to curried functions and a wealth of infix operators. Clearly there are some difficulties in making composition fit 100% with JavaScript. This PR introduces three concepts, 1 of which I support and 2 of which I am neutral on.

For me `S.pipe` is the big addition from this PR. Just looking at the type signature I think it's a lot easier to reason about than the standard variadic ramda pipe. I think it's an improvement in terms of simplification.

`S.compose` and `S.meld` are both well defined and somewhat straightforward but I don't think either of us is overly convinced of their usefulness. Having said that, I'm fine with letting them live in Sanctuary while we figure out if they turn out to be nice to have in practice.

One last point is that I actually prefer the semantics of `R.compose` over `R.pipe` which made me a little hesitant to accept sidelining `S.compose` to a binary function in favor of `S.pipe` but I don't feel strongly enough about this to hold up a PR. Ramda's `R.compose` will also still be an option of course, which I'm fine with.

:dromedary_camel:

davidchambers:
Thank you for your feedback, folks. I'm very happy with `pipe`; I'm not so sure about the other two.

`compose` isn't particularly useful, but it will be nice in the rare cases when one would otherwise write `R.partial(R.compose, f)` or `R.curryN(2, R.compose)`. In all other cases one can simply use `pipe`<sup>†</sup>.

The primary use for `meld` will likely be to create a curried binary function by composing one binary function and one or more unary functions. When `f` is a binary function and `g` and `h` are unary functions, the following are equivalent:

```javascript
S.meld([f, g, h])

R.pipe(f, g, h)           // ramda@0.17.x

R.curry(R.pipe(f, g, h))  // ramda@0.18.x
```

`meld` feels like [`pipeK`][1] in the sense that it may or may not prove useful. We can speculate, of course, but the best way to find out is to make the function available and see whether we find it useful in practice.

---

† Those who prefer right-to-left composition can stick with `R.compose` or push for `S.composeAll`.


[1]: https://github.com/ramda/ramda/pull/1212

CrossEye:
I think I'd like to borrow and adapt that `pipe` explanation for Ramda.  Sanctuary:

> In general terms, `pipe` performs left-to-right composition of a list of functions. `pipe([f, g, h], x)` is equivalent to `h(g(f(x)))`.

For Ramda:

> In general terms, `pipe` performs left-to-right composition of a list of functions. `pipe(f, g, h)(...args)` is equivalent to `h(g(f(...args)))`.

Granted the type signatures are not as simple, but there is something additional going on.

davidchambers:
Please do, @CrossEye. :)

davidchambers:
I learnt a few things while updating a reasonably large project to run against this branch:

  - Immediately applying pipelines is common. Previously I would write:

        R.pipe(
          ...,
          ...,
          ...,
          ...
        )(x)

    I can now write:

        S.pipe([...,
                ...,
                ...,
                ...],
               x)

  - Partially applying `S.compose` is useful when defining nested lenses. To define `.owner.name` and `.owner.email` lenses, for example, one could write:

    ```javascript
    // owner :: { name :: Lens, email :: Lens }
    const owner = R.mapObj(S.compose(S.compose(R.lensProp('owner')), R.lensProp),
                           R.converge(R.zipObj, S.I, S.I)(['name', 'email']));

    R.view(owner.name, {owner: {name: 'alice'}});  // => 'alice'
    ```

  - `S.meld` is *occasionally* useful. I'm currently using it in two places:

    ```javascript
    S.meld([R.map, S.fromMaybe(false)])
    ```

    ```javascript
    S.meld([R.lift(R.lt), S.fromMaybe(false)])
    ```

PR #95, opened 2015-09-20
AMD compatibility

davidchambers:
This pull request makes Sanctuary compatible with AMD module loaders such as [RequireJS][1], as requested by @ChetHarrison.

@CrisFeo, this patch ended up being far less convoluted than we initially feared it would need to be. :)

[1]: http://requirejs.org/

### DIFF
--- a/index.js
+++ b/index.js
@@ -97,21 +97,26 @@
 //.
 //. ## API
 
-;(function() {
+/* global define, self */
+
+;(function(f) {
 
   'use strict';
 
-  var R;
-  var S = {};
-
   /* istanbul ignore else */
   if (typeof module !== 'undefined') {
-    R = require('ramda');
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -97,22 +97,27 @@
...
   if (typeof module !== 'undefined') {
-    R = require('ramda');
-    module.exports = S;
+    module.exports = f(require('ramda'));
+  } else if (typeof define === 'function' && define.amd != null) {
+    define(['ramda'], f);
   } else {
-    R = this.R;
-    this.sanctuary = S;
+    self.sanctuary = f(self.R);
```
davidchambers: [`window.self`][1]


[1]: https://developer.mozilla.org/en-US/docs/Web/API/Window/self

PR #99, opened 2015-09-23
unfoldr

benperez:
Closes https://github.com/plaid/sanctuary/issues/89


index.js
```diff
@@ -1790,6 +1790,30 @@
   S.unlines =
   def('unlines', [List], R.compose(R.join(''), R.map(R.concat(_, '\n'))));
 
+
```

index.js
```diff
@@ -1790,6 +1790,30 @@
...
+  //# unfoldr :: (b -> Maybe (a, b)) -> b -> [a]
+  //.
+  //. The unfoldr function is a `dual' to foldr: while foldr reduces a list to
+  //. a summary value, unfoldr builds a list from a seed value. The function
+  //. takes the element and returns Nothing if it is done producing the list or
+  //. returns Just (a,b), in which case, a is a prepended to the list and b is
+  //. used as the next element in a recursive call.
+  //.
+  //. ```Javascript
```
davidchambers: s/Javascript/javascript/

index.js
```diff
@@ -1790,6 +1790,30 @@
...
+  //. > S.unfoldr(function(n){return n >= 5 ? S.Nothing() : S.Just([n, n + 1]);}, 1)
+  //. [1, 2, 3, 4]
+  //. ```
+  S.unfoldr = def('unfoldr', [Function, a], function(f, init) {
```
davidchambers: Let's name the second parameter `x` to avoid confusion with `S.init`.

index.js
```diff
@@ -1790,6 +1790,30 @@
...
+    var resultList = [];
+    var maybeTuple = f(init);
+    while (is(Just, maybeTuple)) {
+      var tuple = fromMaybe(null, maybeTuple);
+      resultList.push(tuple[0]);
+      maybeTuple = f(tuple[1]);
```
davidchambers: Let's use `maybeTuple.value` rather than `fromMaybe(null, maybeTuple)`. While others shouldn't assume the existence of a `value` property on a Just, it's acceptable for Sanctuary itself to do so (and it *must*, in places).

```javascript
resultList.push(maybeTuple.value[0]);
maybeTuple = f(maybeTuple.value[1]);
```

It'd be nice to rename `maybeTuple` something short such as `m`.

index.js
```diff
@@ -1790,6 +1790,30 @@
...
   def('unlines', [List], R.compose(R.join(''), R.map(R.concat(_, '\n'))));
 
+
+  //# unfoldr :: (b -> Maybe (a, b)) -> b -> [a]
+  //.
+  //. The unfoldr function is a `dual' to foldr: while foldr reduces a list to
+  //. a summary value, unfoldr builds a list from a seed value. The function
+  //. takes the element and returns Nothing if it is done producing the list or
+  //. returns Just (a,b), in which case, a is a prepended to the list and b is
+  //. used as the next element in a recursive call.
```
davidchambers: It's strange to mention `foldr`, a name that doesn't appear here or in Ramda. Could you have a go at rewriting this to reference `R.reduce`?

test/index.js
```diff
@@ -2764,4 +2764,30 @@ describe('string', function() {
 
   });
 
+  describe('unfoldr', function() {
+
+    var f = function(n) {
+      return n >= 5 ? S.Nothing() : S.Just([n, n + 1]);
+    };
```
davidchambers: Let's move this to the `it` block in which it's referenced.

index.js
benperez: :+1:

index.js
benperez: I rewrote this to more closely match the names and JavaScript types we use.

index.js
```diff
@@ -1790,6 +1790,29 @@
...
+  //.
+  //. The unfoldr function is a `dual' to reduce: while reduce "reduces" a list
+  //. to a summary value, unfoldr builds a list from a seed value. The function
+  //. takes the element and returns an instance of Nothing if it is done
+  //. producing the list or returns an instance of Just [a, b], in which case,
+  //. a is a prepended to the list and b is used as the next element in a
+  //. recursive call.
+  //.
+  //. ```javascript
+  //. > S.unfoldr(function(n){return n >= 5 ? S.Nothing() : S.Just([n, n + 1]);}, 1)
```
davidchambers: This function looks a bit squashed. Let's give it some breathing room:

```diff
-  //. > S.unfoldr(function(n){return n >= 5 ? S.Nothing() : S.Just([n, n + 1]);}, 1)
+  //. > S.unfoldr(function(n) { return n >= 5 ? S.Nothing() : S.Just([n, n + 1]); }, 1)
```

The conditional would seem more natural to me written `n < 5 ? S.Just(...) : S.Nothing()`. Did you consider this?

benperez: When I write recursive functions I usually write the base case first, then the recursive case next.

davidchambers: You can update this to use an arrow function now that #104 has been merged. :)

benperez: :zap:

davidchambers:
This function belongs in the List section rather than the String section.

benperez:
:zap:

index.js
```diff
@@ -1616,6 +1616,29 @@
...
   });
 
+  //# unfoldr :: (b -> Maybe (a, b)) -> b -> [a]
+  //.
+  //. The unfoldr function is a `dual' to reduce: while reduce "reduces" a list
+  //. to a summary value, unfoldr builds a list from a seed value. The function
+  //. takes the element and returns an instance of Nothing if it is done
+  //. producing the list or returns an instance of Just [a, b], in which case,
+  //. a is a prepended to the list and b is used as the next element in a
+  //. recursive call.
```
davidchambers: It took well over an hour across several sittings during a two-week period, but I've succeeded in writing a description of this complex function in the style used elsewhere:

```javascript
  //. Takes a function and a seed value, and returns a list generated by
  //. applying the function repeatedly. The list is initially empty. The
  //. function is initially applied to the seed value. Each application
  //. of the function should result in either:
  //.
  //.   - a Nothing, in which case the list is returned; or
  //.
  //.   - Just a pair, in which case the first element is appended to
  //.     the list and the function is applied to the second element.
```

benperez: :zap:

davidchambers:
:deciduous_tree:

PR #101, opened 2015-09-23
use uncurried compose internally for S.pipe

benperez:
Using a curried compose function at every step of `S.pipe` incurs a pretty big performance hit. This leaves the external API unchanged, since `S.compose` and `S.pipe` are still curried.

### DIFF
--- a/index.js
+++ b/index.js
@@ -252,6 +252,12 @@
     return curry(name, types, R.map(function() { return _; }, types), f);
   };
 
+  var compose2 = function(f, g) {
+    return function(x) {
+      return f(g(x));
+    };
+  };
+
   var extend = function(Child, Parent) {
     function Ctor() {
       this.constructor = Child;
@@ -386,7 +392,7 @@
   //. 9
   //. ```
   S.pipe = def('pipe', [List, a], function(fs, x) {
-    return R.reduceRight(compose, I, fs)(x);
+    return R.reduceRight(compose2, I, fs)(x);
   });
 
   //# meld :: [** -> *] -> (* -> * -> ... -> *)


davidchambers:
Thank you, @benperez!

index.js
```diff
@@ -385,8 +385,13 @@
...
   //. ```
+  var uncurriedCompose = function(f, g) {
+    return function(x) {
+      return f(g(x));
+    };
+  };
   S.pipe = def('pipe', [List, a], function(fs, x) {
-    return R.reduceRight(compose, I, fs)(x);
+    return R.reduceRight(uncurriedCompose, I, fs)(x);
   });
```
davidchambers: Let's define the function in situ. Also, could we switch the application order and use `R.reduce` rather than `R.reduceRight`?

kedashoe:
:+1: , but if we're going to improve performance, why not just use a for loop? `meld` and `find` use them, so it seems to be allowed :)

benperez:
@kedashoe here are the two implementations I've come up with:
```javascript
// Reduce
S.pipe = def('pipe', [List, a], function(fs, x) {
  return R.reduce(function(f, g) {
                    return function(x) {
                      return g(f(x));
                    };
                  }, I, fs)(x);
});
```

```javascript
// for loop immediate application
S.pipe = def('pipe', [List, a], function(fs, x) {
  var acc = x;
  for (var idx = 0, len = fs.length; idx += 1) {
    acc = fs[idx](acc);
  }
  return acc;
});
```

In our admittedly unscientific benchmark of an internal Plaid project that makes heavy use of `S.pipe` the reduce version ran in 25.272s while the for loop version took 34.377s. For comparison the original version with a curried compose took 44.053s.

davidchambers:
:deciduous_tree:

PR #104, opened 2015-09-29
use arrow functions in doctests

davidchambers:
This change was motivated by a doctest in #99; ramda/ramda#1398 likely gave me the idea.


benperez:
:arrow_forward:

PR #105, opened 2015-09-30
do not rely on constructor identity

### LINKED ISSUE #100: remove instanceof checks
Sanctuary's run-time type checking currently relies on `instanceof` checks. This guarantees that if a value passes type checking it will support all the expected operations.

If a function requires an argument of type `Maybe a`, for example, a Maybe created by a *different* version of Sanctuary will fail type checking (with a rather confusing error message). For example:

    TypeError: ‘fromMaybe’ requires a value of type Maybe as its second argument; received Just(42)

The problem with this approach is that it rejects *valid* values too. Requiring the *same* version of Sanctuary from two different paths results in two incompatible Sanctuary modules existing in memory:

```javascript
const S1 = require('./path/to/sanctuary-0.7.0');
const S2 = require('./another/path/to/sanctuary-0.7.0');

S1.fromMaybe(NaN, S2.Just(42));
// ! TypeError: ‘fromMaybe’ requires a value of type Maybe as its second argument; received Just(42)
```

When might this occur?

  - In the browser, when dealing with iframes.

  - In Node, when using the [vm][1] module. @raine and I both discovered this while building REPLs.

  - In Node, when `package-a` depends on both `sanctuary` and `package-b`, and `package-b` also depends on `sanctuary`. If `package-a` applies a `package-b` function which returns a Maybe, the Maybe will not be recognized by the `sanctuary` module required by `package-a`.

    Note that npm performs dependency deduplication, so if `package-a` and `package-b` depend on the same version of `sanctuary`, both will use __package-a/node\_modules/sanctuary__. *Relying* on deduplication is undesirable: it's an optimization which should not effect a program's behaviour.

__The first and second scenarios also affect built-in types such as Array.__ [`R.is`][2] is also affected by this issue, as discussed in ramda/ramda#1140.

Clearly we must avoid `instanceof` checks. What might we do instead? For built-in types we have `Object.prototype.toString` and [`R.type`][3]:

```javascript
> Object.prototype.toString.call([])
'[object Array]'
> R.type([])
'Array'
```

For Sanctuary types (and other custom types) we need a different approach.

We can't use `.constructor.name` because a type may have multiple constructors: Maybe has `Nothing` and `Just`, for example. Furthermore, there's no guarantee that two constructors with the same name are equivalent.

I propose a special property named `'@@type'` or similar, with a [reverse-DNS][4]-ish value such as `'com.github.plaid.sanctuary/Maybe'`. The implementation would be straightforward:

```javascript
Maybe.prototype['@@type'] = 'com.github.plaid.sanctuary/Maybe';
```

We could then write a function to determine the unique identifier for a value's type:

```javascript
//  typeName :: a -> String
var typeName = function(x) {
  return x != null && R.type(x['@@type']) === 'String' ?
    x['@@type'] :
    'org.ecma-international.ecma-262/' + R.type(x);
};

typeName(Just(42));
// => 'com.github.plaid.sanctuary/Maybe'

typeName([1, 2, 3]);
// => 'org.ecma-international.ecma-262/Array'
```

Hat tip to @jethrolarson who has been giving this matter a lot of thought. I got the idea for a `'@@type'` property from his [xface][5] project (though xface uses the property for a slightly different purpose).

Solving the problems associated with `instanceof` checks is the most important issue I'd like to address in v0.8.0. I'd love to know what others think of this proposal. Counterproposals welcome!

[1]: https://nodejs.org/api/vm.html
[2]: http://ramdajs.com/docs/#is
[3]: http://ramdajs.com/docs/#type
[4]: https://en.wikipedia.org/wiki/Reverse_domain_name_notation
[5]: https://gist.github.com/jethrolarson/5eaad6d7b22b231741c9

davidchambers:
Closes #100

> `S.is`, `S.get`, `S.gets`, and `S.pluck` now ignore constructor identity, allowing them to operate on objects from other environments, such as iframes and sandboxes. Additionally, packages which export functions which take and/or return values of Sanctuary types no longer rely on npm package deduplication to satisfy the requirements of Sanctuary's run-time type checking.
>
> Since `instanceof` checks are no longer acceptable, `S.is` no longer respects subtyping.
>
> This [pull request] also introduces the following functions:
>
> ```haskell
> type      :: a -> String
> isNothing :: Maybe a -> Boolean
> isJust    :: Maybe a -> Boolean
> isLeft    :: Either a b -> Boolean
> isRight   :: Either a b -> Boolean
> ```

### DIFF
--- a/index.js
+++ b/index.js
@@ -141,6 +141,9 @@
     return x != null && x['@@functional/placeholder'] === true;
   };
 
+  //  functionName :: TypeRep a -> String
+  var functionName = R.compose(R.nth(1), R.match(/^function (\w*)/), String);
+
   var formatters = {
     '{}': R.identity,
     '{card}': R.ifElse(R.lte(_, 10),
@@ -150,7 +153,7 @@
     '{ord}': R.nth(_, ['first', 'second', 'third']),
     '{quote}': function(s) { return '\u2018' + s + '\u2019'; },
     '{repr}': R.toString,
-    '{type}': R.pipe(String, R.match(/^function (\w*)/), R.nth(1))
+    '{type}': functionName
   };
 
   //  format :: String -> [*] -> String
@@ -170,23 +173,37 @@
   var b = {name: 'b'};
   var c = {name: 'c'};
 
+  //  _type :: a -> String
+  var _type = function(x) {
+    return x != null && R.type(x['@@type']) === 'String' ? x['@@type']
+                                                         : R.type(x);
+  };
+
+  //  _is :: (TypeRep a, b) -> Boolean
   var _is = function(type, x) {
-    if (x == null) return false;
-    switch (type) {
-      case Accessible:
-        return true;
-      case Integer:
-        return _is(Number, x) && Math.floor(x) === Number(x) &&
-               x >= MIN_SAFE_INTEGER && x <= MAX_SAFE_INTEGER;
-      case List:
-        return !_is(Function, x) && _is(Number, x.length) && x.length >= 0;
-      case RegexFlags:
-        return R.test(/^g?i?m?$/, x);
-      case TypeRep:
-        return _is(Function, x);
-      default:
-        return Object(x) instanceof type;
-    }
+    return (
+      x == null ?
+        false :
+      type === Accessible ?
+        true :
+      type === Integer ?
+        R.type(x) === 'Number' &&
+        Math.floor(x) === Number(x) &&
+        x >= MIN_SAFE_INTEGER &&
+        x <= MAX_SAFE_INTEGER :
+      type === List ?
+        R.type(x) !== 'Function' &&
+        R.type(x.length) === 'Number' &&
+        x.length >= 0 :
+      type === RegexFlags ?
+        R.test(/^g?i?m?$/, x) :
+      type === TypeRep ?
+        R.type(x) === 'Function' :
+      R.type(type.prototype['@@type']) === 'String' ?
+        x['@@type'] === type.prototype['@@type'] :
+      // else
+        R.type(x) === functionName(type)
+    );
   };
 
   var arity = function(n, f) {
@@ -198,7 +215,7 @@
     }
   };
 
-  //  curry :: (String, [TypeRep], [*], Function) -> Function
+  //  curry :: (String, [TypeRep *], [*], Function) -> Function
   var curry = function(name, types, _values, f) {
     return arity(R.filter(placeholder, _values).length, function() {
       var values = _values;  // Locally scoped variable to update.
@@ -232,7 +249,7 @@
           for (idx = 0; idx < values.length; idx += 1) {
             var val = values[idx];
             if (types[idx] === type && !placeholder(val) &&
-                !(R.type(val) === R.type(arg) && val.type === arg.type)) {
+                _type(val) !== _type(arg)) {
               throw new TypeError(format(
                 '{quote} requires its {ord} and {ord} arguments ' +
                 'to be of the same type; {repr} and {repr} are not',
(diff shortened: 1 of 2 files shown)


albrow:
> Subtyping is not respected.

This is unfortunate. I've found subtyping to be incredibly useful for expressing error types, for example.  But I understand the reasoning. I also realized that if someone wants to consider subtypes they can always use ramda or some other solution.

I wonder if for name-spacing `@@type` it might be better to use the npm package name? It would be shorter, and IMO easier to understand where the type came from. Or is the idea that not every javascript library will be an npm package?

kedashoe:
We could also have a companion function, something like

```js
S.isA = function isA(type, x) {
  do {
    if (S.is(type, x)) {
      return true;
    }
  } while (x = Object.getPrototypeOf(x));
  return false;
}

var x = new TypeError('bar');
console.log(S.isA(TypeError, x)); //=> true
console.log(S.isA(RangeError, x)); //=> false
console.log(S.isA(Error, x)); //=> true

var y = S.Just(10);
console.log(S.isA(S.Just, y)); //=> true
console.log(S.isA(S.Nothing, y)); //=> false
console.log(S.isA(S.Maybe, y)); //=> true
```
(this blows up on `null` of course but you get the idea..)

kedashoe:
> I wonder if for name-spacing @@type it might be better to use the npm package name?

How about for native objects such as `Array` there is no "namespace", but for anything else we use the package name?

```js
S.type([1, 2, 3]); //=> "Array"
S.type(S.Just(42)); //=> "sanctuary/Maybe"
```

davidchambers:
> I wonder if for name-spacing `@@type` it might be better to use the npm package name?

I like it! I'd been thinking that built-in types need to live in a namespace too, but as @kedashoe points out it's fine for `S.type([1, 2, 3])` to evaluate to `'Array'`. I'll update the patch. :)

index.js
```diff
@@ -170,23 +173,36 @@
   var b = {name: 'b'};
   var c = {name: 'c'};
 
+  //  _type :: a -> String
+  var _type = function(x) {
+    return x != null && R.type(x['@@type']) === 'String' ? x['@@type']
+                                                         : R.type(x);
+  };
+
   var _is = function(type, x) {
```
soroushjp: Do we want a type signature here? --

    //  _type :: (TypeRep a, b) -> Boolean

CrossEye:
I'm wondering if this should use `Symbol.for('type')` in environments that support it before `@@type`.  That seems to be the direction of the language.

davidchambers:
Could you outline the advantages of doing so, @CrossEye? I'm not yet familiar with the Symbol type.

CrossEye:
The main reason I would consider doing this is that it is becoming the standard way to attach metadata to an object.  For this use, the chief advantages would be fairly small, but we would be working with the way the language is proceeding.  Those advantages:

  * `Symbol` keys are non-enumerable.  You can get at them through the `Reflect` API or `Object.getOwnPropertySymbols(obj)`, but they won't be cluttering up normal property iteration of an object.

  * `Symbol`s are clearly metadata.  There is simply no chance of mistaking someone's strangely named and unrelated `@@type` property for our metadata reference.

  * Describing an object's type seems a worthy parallel to many of the [well-known symbols][wk] built into the language.

One usual advantage we **don't** get in this suggestion is clear-cut uniqueness.  This has to be a shared symbol, available across code realms, and available to the library and its users independently.  By using `Symbol.for`, we guarantee that we always are using the same symbol, but there is, of course, a chance that we'd conflict with other code that uses the same construct for different purposes.

But I do think that the benefits are real, even if small.

The best article I know on Symbols is the one by [Axel Rauschmayer ][ar], but the ones by [Keith Cirkel][kc] and [Jason Orendorff][jo] are also worthwhile, as, of course, is the [MDN article][md].

  [ar]: http://www.2ality.com/2014/12/es6-symbols.html
  [jo]: https://hacks.mozilla.org/2015/06/es6-in-depth-symbols/
  [kc]: http://blog.keithcirkel.co.uk/metaprogramming-in-es6-symbols/
  [md]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol 
  [wk]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#Well-known_

davidchambers:
> By using `Symbol.for`, we guarantee that we always are using the same symbol, but there is, of course, a chance that we'd conflict with other code that uses the same construct for different purposes.

This is as I'd feared. `Symbol.for('type')` and `'@@type'` both rely on convention. The question is whether the minor benefits of using a Symbol in certain environments outweigh the costs of having different behaviour in different environments. At this stage, I'm unconvinced. :)

Thanks for the excellent summary, Scott!

PR #106, opened 2015-10-01
maybe, either: simplify S.encase and S.encaseEither

### LINKED ISSUE #103: simplify S.encase and S.encaseEither
Currently, [`encase`][1] and [`encaseEither`][2] have misleading type signatures:

```haskell
encase :: (* -> a) -> (* -> Maybe a)

encaseEither :: (Error -> a) -> (* -> b) -> (* -> Either a b)
```

They are misleading because Hindley–Milner notation cannot represent a function which returns a curried function of unknown arity.

In addition to the documentation problem, there's a usability problem. Functions with unwanted optional arguments must be wrapped in lambdas. For example:

```javascript
S.encaseEither(S.I, x => JSON.parse(x))
```

A solution to both problems is to have one function for each arity up to some arbitrary limit:

```haskell
encase  :: (a -> b) -> a -> Maybe b
encase2 :: (a -> b -> c) -> a -> b -> Maybe c
encase3 :: (a -> b -> c -> d) -> a -> b -> c -> Maybe d

encaseEither  :: (Error -> l) -> (a -> r) -> a -> Either l r
encaseEither2 :: (Error -> l) -> (a -> b -> r) -> a -> b -> Either l r
encaseEither3 :: (Error -> l) -> (a -> b -> c -> r) -> a -> b -> c -> Either l r
```

An incidental benefit of this approach is that arguments to the wrapped function may be provided up front. For example:

```javascript
> S.encaseEither(R.prop('message'), JSON.parse, '[1,2,3]')
Right([1, 2, 3])
> S.encaseEither(R.prop('message'), JSON.parse, 'XXX')
Left("Unexpected token X")
```

[1]: https://github.com/plaid/sanctuary#encase
[2]: https://github.com/plaid/sanctuary#encaseEither

davidchambers:
Closes #103

### DIFF
--- a/index.js
+++ b/index.js
@@ -150,7 +150,7 @@
                        R.nth(_, ['zero', 'one', 'two', 'three', 'four', 'five',
                                  'six', 'seven', 'eight', 'nine', 'ten']),
                        String),
-    '{ord}': R.nth(_, ['first', 'second', 'third', 'fourth']),
+    '{ord}': R.nth(_, ['first', 'second', 'third', 'fourth', 'fifth']),
     '{quote}': function(s) { return '\u2018' + s + '\u2019'; },
     '{repr}': R.toString,
     '{type}': functionName
@@ -213,6 +213,7 @@
       case 2: return function(a, b) { return f.apply(this, arguments); };
       case 3: return function(a, b, c) { return f.apply(this, arguments); };
       case 4: return function(a, b, c, d) { return f.apply(this, arguments); };
+      case 5: return function(a, b, c, d, e) { return f.apply(this, arguments); };
     }
   };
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -2016,9 +2058,7 @@
   //. > S.parseJson('[')
   //. Nothing()
   //. ```
-  S.parseJson = def('parseJson', [String], encase(function(s) {
-    return JSON.parse(s);
-  }));
+  S.parseJson = def('parseJson', [String], encase(JSON.parse));
```
CrisFeo: Nice!

CrisFeo:
Funny how sometimes simplifying requires adding 200 more lines ;)

Good work on this @davidchambers! :moyai:

index.js
davidchambers: It'll be nice to no longer be dependent on the `length` of the provided function. :)

davidchambers:
Are you in favour of these changes, @svozza?

svozza:
Initially I was bit hesitant about but having seen how fixing arity simplifies the implementations of functions (not just in this PR) I'm fully supportive of this direction now.

PR #107, opened 2015-10-01
regexp: add S.regex

### LINKED ISSUE #97: function for creating RegExp values
Similar to [`regex`][1] from PureScript's Data.String.Regex module.

[1]: https://github.com/purescript/purescript-strings/blob/v0.7.0/docs/Data/String/Regex.md#regex-1

davidchambers:
Closes #97

I'm not thrilled about naming this function `regex` when the type's name is RegEx<strong>p</strong>, but `regexp` is inconsistent in its capitalization and `regExp` is unbearably ugly. `regex` seems the best of the three options.

Having dropped the "p" elsewhere, I'm happy to name the pseudotype RegexFlags (as PureScript does).

### DIFF
--- a/index.js
+++ b/index.js
@@ -81,6 +81,19 @@
 //. The List pseudotype represents non-Function values with numeric `length`
 //. properties greater than or equal to zero, such as `[1, 2, 3]` and `'foo'`.
 //.
+//. ### RegexFlags pseudotype
+//.
+//. The RegexFlags pseudotype represents the set of canonical [RegExp][] flags:
+//.
+//.   - `''`
+//.   - `'g'`
+//.   - `'i'`
+//.   - `'m'`
+//.   - `'gi'`
+//.   - `'gm'`
+//.   - `'im'`
+//.   - `'gim'`
+//.
 //. ### Type representatives
 //.
 //. What is the type of `Number`? One answer is `a -> Number`, since it's a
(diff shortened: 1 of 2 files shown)


lucamartinetti:
@davidchambers I will need a little context here to review. Will ask about this later :-)

davidchambers:
`S.regex` is a function for creating RegExp values. JavaScript provides a constructor for this, but using `new` is inconvenient when writing functional pipelines. The following function, for example, could be used to grep a log file:

```javascript
// matchPrefixedLines :: String -> String -> [String]
//
// Takes a prefix pattern and a string and returns a list of lines matched
// by the given prefix pattern.
const matchPrefixedLines =
S.pipe([R.concat('^'),
        R.concat(_, '.*$'),
        s => new RegExp(s, 'gm'),
        R.match]);
```

Having to write

```javascript
s => new RegExp(s, 'gm')
```

here is unfortunate, and in environments which don't support arrow functions one must write

```javascript
function(s) { return new RegExp(s, 'gm'); }
```

which is quite wordy.

This pull request will enable:

```javascript
S.regex('gm')
```

The other advantages of `S.regex` over the RegExp constructor are that it validates that the arguments are of the appropriate types, and ensures that the "flags" argument is one of the eight permitted values.

lucamartinetti:
LGTM. What about using some descriptive constants for the flags like `S.REGEX_MATCH_CASE_INSENSITIVE ` ?

davidchambers:
> What about using some descriptive constants for the flags like `S.REGEX_MATCH_CASE_INSENSITIVE ` ?

Given that we're validating the "flags" argument, this would not provide additional safety. It is arguably more descriptive, but because `''`, `'g'`, `'i'`, `'m'`, `'gi'`, `'gm'`, `'im'`, and `'gim'` are used when providing flags to the RegExp constructor, I find

```javascript
S.regex('gm')
```

and

```javascript
S.regex([S.REGEX_MATCH_GLOBAL, S.REGEX_MATCH_MULTILINE])
```

equally clear. All else being equal, the shorter expression gets my vote.

I appreciate your feedback, @lucamartinetti. This pull request will be a more useful source of information for current and future Sanctuary users as a result of this discussion.

PR #109, opened 2015-10-25
regexp: add S.test

davidchambers:
Like [`R.test`][1], but with the type checking we know and love. See ramda/ramda#1413 for a compelling example of the problems that arise due to Ramda's lack of type checking.

/cc @albrow

[1]: http://ramdajs.com/docs/#test

### DIFF
--- a/index.js
+++ b/index.js
@@ -2015,6 +2015,22 @@
     return new RegExp(source, flags);
   });
 
+  //# test :: RegExp -> String -> Boolean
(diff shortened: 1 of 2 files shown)


svozza:
LGTM. I've actually been caught by a similar bug to 1413 before so I think this is great.

albrow:
@davidchambers thanks for this! Looks good to me.

Issue #102, opened 2015-09-24
port higher-order Ramda functions which accept variadic functions


davidchambers:
In Ramda we're slowly replacing variadic functions with fixed-arity functions. I'm very much in favour of this. I'd like to go a step further in Sanctuary by fixing the arities of function *arguments*. [`R.both`][1], for example, has type `(** -> Boolean) -> (** -> Boolean) -> (** -> Boolean)`, where `**` represents zero or more arguments. I'd like to provide a `both` function in Sanctuary which is less flexible but which handles the common case with a simpler type, `(a -> Boolean) -> (a -> Boolean) -> a -> Boolean`.

The following is a list of such functions I'd like Sanctuary to provide (which I'll update it if turns out not to be exhaustive).

  - `both :: (a -> Boolean) -> (a -> Boolean) -> a -> Boolean`
  - `either :: (a -> Boolean) -> (a -> Boolean) -> a -> Boolean`
  - `anyPass :: [a -> Boolean] -> a -> Boolean`
  - `allPass :: [a -> Boolean] -> a -> Boolean`
  - `complement :: (a -> Boolean) -> a -> Boolean`
  - `ifElse :: (a -> Boolean) -> (a -> b) -> (a -> b) -> a -> b`
  - `converge :: (b -> c -> d) -> (a -> b) -> (a -> c) -> a -> d`
  - `flip :: (a -> b -> c) -> b -> a -> c`

This approach was briefly discussed in ramda/ramda#1395.

[1]: http://ramdajs.com/docs/#both

svozza:
If no one is working on this I can grab `both`, `either` and `complement`?

davidchambers:
Please do, @svozza. :)

svozza:
Quick question, do we want to do the non-short circuiting version using `lift`?

davidchambers:
I think we should short-circuit. It's likely Ramda will do the same.

svozza:
Should I just wrap the relevant Ramda functions because we're still on v0.17 or will I reimplement them from scratch?

davidchambers:
It would be clearer to implement them from scratch. This will also simplify the process of upgrading to Ramda v0.18.x.

svozza:
What will we do about `either` given that it shares a name with a function currently in Sanctuary.

davidchambers:
Yikes! That's a terrific question. `and` and `or` already exist as well. We should not merge #110 until we resolve this conundrum.

[...]

PR #110, opened 2015-11-04
Add S.both function

svozza:
I've chosen to only implement the `both` part of #102 for the moment just to get a feel for what's expected. The examples I've given in the docs are pretty contrived because there's really only `S.test` that returns a boolean. I could have used something like `R.has` but I wasn't sure what the policy of including Ramda functions in the docs was or even defining a custom function at the head of the `S.both` docs. I'll squash everything down to one commit at the end of code review.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1497,6 +1497,27 @@
     return xBool !== yBool ? or(x, y) : xEmpty;
   });
 
+  //# both :: (a -> Boolean) -> (a -> Boolean) -> a -> Boolean
+  //.
+  //. Takes two unary functions that return a boolean and applies them
+  //. both to the supplied value, returning true iff both functions
+  //. evaluate to true.
+  //.
+  //. ```javascript
+  //.
+  //. > S.both(S.test(/^a/), S.test(/.*f$/), 'abcdef')
+  //. true
+  //.
+  //. > S.both(S.test(/^a/), S.test(/.*e$/), 'abcdef')
+  //. false
+  //.
+  //. > S.both(S.test(/^b/), S.test(/.*e$/), 'abcdef')
+  //. false
+  //. ```
+  S.both = def('both', [Function, Function, a], function(f, g, x) {
+    return f(x) && g(x);
+  });
+
   //. ### List
 
   //# slice :: Integer -> Integer -> [a] -> Maybe [a]
--- a/index.js
+++ b/index.js
@@ -1500,7 +1500,7 @@
   //# both :: (a -> Boolean) -> (a -> Boolean) -> a -> Boolean
   //.
   //. Takes two unary predicates and a value of any type, and returns
-  //. true if the value satisfies both predicates; false otherwise.
+  //. `true` if the value satisfies both predicates; `false` otherwise.
   //.
   //. ```javascript
   //. > S.both(S.test(/a/), S.test(/b/), 'banana')
--- a/index.js
+++ b/index.js
@@ -1504,7 +1504,6 @@
   //. evaluate to true.
   //.
   //. ```javascript
-  //.
   //. > S.both(S.test(/^a/), S.test(/.*f$/), 'abcdef')
   //. true
   //.
--- a/index.js
+++ b/index.js
@@ -1499,18 +1499,17 @@
 
   //# both :: (a -> Boolean) -> (a -> Boolean) -> a -> Boolean
   //.
-  //. Takes two unary functions that return a boolean and applies them
-  //. both to the supplied value, returning true iff both functions
-  //. evaluate to true.
+  //. Takes two unary predicates and a value of any type, and returns
+  //. true if the value satisfies both predicates; false otherwise.
   //.
   //. ```javascript
-  //. > S.both(S.test(/^a/), S.test(/.*f$/), 'abcdef')
+  //. > S.both(S.test(/a/), S.test(/b/), 'banana')
   //. true
   //.
-  //. > S.both(S.test(/^a/), S.test(/.*e$/), 'abcdef')
+  //. > S.both(S.test(/a/), S.test(/b/), 'insouciant')
   //. false
   //.
-  //. > S.both(S.test(/^b/), S.test(/.*e$/), 'abcdef')
+  //. > S.both(S.test(/a/), S.test(/b/), 'serendipity')
   //. false
   //. ```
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1497,6 +1497,26 @@
     return xBool !== yBool ? or(x, y) : xEmpty;
   });
 
+  //# both :: (a -> Boolean) -> (a -> Boolean) -> a -> Boolean
+  //.
+  //. Takes two unary functions that return a boolean and applies them
+  //. both to the supplied value, returning true iff both functions
+  //. evaluate to true.
```
davidchambers: - Let's write _unary predicates_ rather than _unary functions that return a boolean_.
  - It's not accurate to state _applies them both_ (due to short-circuiting).

How about the following?

_Takes two unary predicates and a value of any type, and returns `true` if the value satisfies both predicates; `false` otherwise._

index.js
```diff
@@ -1497,6 +1497,26 @@
...
+  //. ```javascript
+  //. > S.both(S.test(/^a/), S.test(/.*f$/), 'abcdef')
+  //. true
+  //.
+  //. > S.both(S.test(/^a/), S.test(/.*e$/), 'abcdef')
+  //. false
+  //.
+  //. > S.both(S.test(/^b/), S.test(/.*e$/), 'abcdef')
+  //. false
+  //. ```
```
davidchambers: As you intimated, `S.both(S.test(/^a/), S.test(/.*f$/))` can be written `S.test(/^a.*f$/)`. It would be nice to use predicates which cannot be combined in this way. How about these?

```javascript
> S.both(S.test(/a/), S.test(/b/), 'banana')
true

> S.both(S.test(/a/), S.test(/b/), 'orange')
false
```

index.js
```diff
@@ -1497,6 +1497,26 @@
...
+  S.both = def('both', [Function, Function, a], function(f, g, x) {
+    return f(x) && g(x);
+  });
```
davidchambers: I :heart: this implementation.

test/index.js
```diff
@@ -1914,6 +1914,53 @@ describe('control', function() {
...
+    it('returns false when one predicate is satisfied', function() {
+      eq(S.both(S.test(/^a/), S.test(/.*f$/), 'abcde'), false);
+      eq(S.both(S.test(/^b/), S.test(/.*f$/), 'abcdef'), false);
+    });
+
+    it('returns false when no predicates are satisfied', function() {
+      eq(S.both(S.test(/^b/), S.test(/.*e$/), 'abcdef'), false);
+    });
+
+    it('short circuits if the first predicate is not satisfied', function() {
```
davidchambers: I think "short-circuits" should be hyphenated.

test/index.js
```diff
@@ -1914,6 +1914,53 @@ describe('control', function() {
...
+      var evaluated = false;
+      var evaluate = function() { evaluated = true; };
+      eq(S.both(S.test(/^b/), evaluate, 'abcdef'), false);
+      eq(evaluated, false);
+    });
```
davidchambers: This is a nice way to test the short-circuiting behaviour. :)

test/index.js
```diff
@@ -1914,6 +1914,53 @@ describe('control', function() {
...
+
+    it('throws if not supplied a function', function() {
```
davidchambers: Let's make this the second `it` block, as it is for other functions. For consistency, the description should be `'type checks its arguments'`.

test/index.js
```diff
@@ -1914,6 +1914,53 @@ describe('control', function() {
...
+      assert.throws(function() { S.both('string', S.test(/^a/), 'abcdef'); },
+          errorEq(TypeError,
+              '‘both’ requires a value of type Function as its ' +
+              'first argument; received "string"'));
```
davidchambers: I like to horizontally align the arguments to a function, when wrapped. In this case:

```javascript
assert.throws(function() { S.both('string', S.test(/^a/), 'abcdef'); },
              errorEq(TypeError,
                      '‘both’ requires a value of type Function ' +
                      'as its first argument; received "string"'));
```

Also, we could reduce the failing expression to `S.both('string')`.

test/index.js
```diff
@@ -1914,6 +1914,53 @@ describe('control', function() {
...
+
+      assert.throws(function() { S.both(S.test(/.*f$/), 'string', 'abcdef'); },
+          errorEq(TypeError,
+              '‘both’ requires a value of type Function as its ' +
+              'second argument; received "string"'));
+    });
+
+    it('is curried', function() {
+      eq(S.both(S.test(/^a/), S.test(/.*f$/)).length, 1);
+      eq(S.both(S.test(/^a/)).length, 2);
```
davidchambers: Let's flip the two lines above.

test/index.js
svozza: I wish I could take credit but I robbed the nub of the idea from the `R.both` tests. :see_no_evil:

svozza:
All done. :)

index.js
```diff
@@ -1497,6 +1497,25 @@
     return xBool !== yBool ? or(x, y) : xEmpty;
   });
 
+  //# both :: (a -> Boolean) -> (a -> Boolean) -> a -> Boolean
+  //.
+  //. Takes two unary predicates and a value of any type, and returns
+  //. true if the value satisfies both predicates; false otherwise.
```
davidchambers: Let's wrap "true" and "false" with backticks so they are rendered as `true` and `false`.

index.js
```diff
@@ -1497,6 +1497,25 @@
...
+  //.
+  //. ```javascript
+  //. > S.both(S.test(/a/), S.test(/b/), 'banana')
+  //. true
+  //.
+  //. > S.both(S.test(/a/), S.test(/b/), 'insouciant')
```
davidchambers: Nice word!

test/index.js
davidchambers: It looks as though this comment got lost in the shuffle.

index.js
svozza: Hehe. One of my faves!

svozza:
Is there something up with Github? I'm seeing some line notes in email but when I look in the browser they're not there.

davidchambers:
Reveal the hidden diff comments. ;)

svozza:
Aha. :mag:

svozza:
We should close this in favour of the `lift` approach.

PR #111, opened 2015-11-05
Add complement and "either" function

svozza:
Obviously merging this depends on the outcome of #102. I've renamed `either` to `oneOf` not because I think it's a good name but because I had to call it _something_. I can update the PR when we've made a decision.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1513,6 +1513,26 @@
     return !f(x);
   });
 
+  //# oneOf :: (a -> Boolean) -> (a -> Boolean) -> a -> Boolean
+  //.
+  //. Takes two unary predicates and a value of any type, and returns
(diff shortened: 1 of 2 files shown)


svozza:
I still think `complement` might be useful so I can amend this PR to remove the 'either' function?

davidchambers:
`complement` is `lift(not)`, so perhaps what we really want is to define `S.not :: Boolean -> Boolean` along with `S.lift{,2,3}`. I could work on this myself, but I plan to work on sanctuary-def this weekend.

svozza:
Yes, that makes sense. I can easily do `S.not` at some point over the weekend, it shouldn't be very difficult at all.

PR #112, opened 2015-11-06
Add ifElse function

svozza:
I tried to avoid using `S.test` again by using `Just#equals` but that function depends on `this` so the context got messed up when passed into `R.ifElse`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -150,7 +150,7 @@
                        R.nth(_, ['zero', 'one', 'two', 'three', 'four', 'five',
                                  'six', 'seven', 'eight', 'nine', 'ten']),
                        String),
-    '{ord}': R.nth(_, ['first', 'second', 'third']),
+    '{ord}': R.nth(_, ['first', 'second', 'third', 'fourth']),
     '{quote}': function(s) { return '\u2018' + s + '\u2019'; },
     '{repr}': R.toString,
     '{type}': functionName
@@ -212,6 +212,7 @@
       case 1: return function(a) { return f.apply(this, arguments); };
       case 2: return function(a, b) { return f.apply(this, arguments); };
       case 3: return function(a, b, c) { return f.apply(this, arguments); };
+      case 4: return function(a, b, c, d) { return f.apply(this, arguments); };
     }
   };
 
@@ -1495,6 +1496,25 @@
     var yBool = toBoolean(y);
     var xEmpty = empty(x);
     return xBool !== yBool ? or(x, y) : xEmpty;
+  });
+
+  //# ifElse :: (a -> Boolean) -> (a -> b) -> (a -> b) -> a -> b
+  //.
+  //. Takes a unary predicate, a unary "if" function, a unary "else"
+  //. function, and a value of any type, and returns the result of
+  //. applying the "if" function to the value if the value satisfies
+  //. the predicate; the result of applying the "else" function to the
+  //. value otherwise.
+  //.
+  // ```javascript
+  //. > S.ifElse(x => x < 0, Math.abs, Math.sqrt, -1)
+  //. 1
+  //.
+  //. > S.ifElse(x => x < 0, Math.abs, Math.sqrt, 16)
+  //. 4
+  //. ```
+  S.ifElse = def('ifElse', [Function, Function, Function, a], function(pred, f, g, x) {
+    return pred(x) ? f(x) : g(x);
   });
 
   //. ### List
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1497,6 +1498,23 @@
     return xBool !== yBool ? or(x, y) : xEmpty;
   });
 
+  //# ifElse :: (a -> Boolean) -> (a -> b) -> (a -> b) -> a -> b
+  //.
+  //. Takes a unary predicate, two unary functions and a value of any type
+  //. and applies the first function to the value if the predicate evaluates
+  //. to true; otherwise applies the second function to the value.
```
davidchambers: We could make it clear that the return value is the result of applying one of the `a -> b` functions. Perhaps:

_Takes a unary predicate, a unary "if" function, a unary "else" function, and a value of any type, and returns the result of applying the "if" function to the value if the value satisfies the predicate; the result of applying the "else" function to the value otherwise._

index.js
```diff
@@ -1497,6 +1498,23 @@
...
+  //.
+  // ```javascript
+  //. > S.ifElse(S.test(/light/), S.K('dark'), S.K('light'), 'light')
+  //. 'dark'
+  //.
+  //. > S.ifElse(S.test(/light/), S.K('dark'), S.K('light'), 'dark')
+  //. 'light'
+  //. ```
```
davidchambers: How about these?

```javascript
> S.ifElse(x => x < 0, Math.abs, Math.sqrt, -1)
1

> S.ifElse(x => x < 0, Math.abs, Math.sqrt, 16)
4
```

index.js
```diff
@@ -1497,6 +1498,23 @@
...
+  S.ifElse = def('ifElse', [Function, Function, Function, a], function(pred, f, g, x) {
+    return R.ifElse(pred, f, g)(x);
```
davidchambers: I'd prefer:

```javascript
return pred(x) ? f(x) : g(x);
```

It's just as short, and avoids the overhead of Ramda's currying.

test/index.js
```diff
@@ -1914,6 +1914,44 @@ describe('control', function() {
...
+
+    it('is a quaternary function', function() {
+      eq(typeof S.ifElse, 'function');
+      eq(S.ifElse.length, 4);
+    });
+
+    it('type checks its arguments', function() {
+      assert.throws(function() { S.ifElse('wrong'); },
+             errorEq(TypeError, '‘ifElse’ requires a value of type Function ' +
+                     'as its first argument; received "wrong"'));
```
davidchambers: Format like this, please:

```javascript
assert.throws(function() { S.ifElse('wrong'); },
              errorEq(TypeError,
                      '‘ifElse’ requires a value of type Function ' +
                      'as its first argument; received "wrong"'));
```

index.js
svozza: Yes, that would be great. I wasn't sure if we could use ES6 in the examples. It would be nice to be able to use it in the tests: no build step either if we use `mocha --compilers js:babel/register`. :wink:

index.js
svozza: And it looks nicer too...

test/index.js
svozza: Argh, I actually had that but my IDE keeps 'autocorrecting' the formatting!

svozza:
Is there anything else here that needs tidying up?

davidchambers:
Let's use the expressions used in the doctests in the unit tests themselves. `S.K('light')` ignores its argument, obfuscating the behaviour of `ifElse`.

Also, please squash the commits.

svozza:
Done!

davidchambers:
I just thought of one more thing. We should update this line:

```javascript
'{ord}': R.nth(_, ['first', 'second', 'third']),
```

We should include `'fourth'` here (even though there are no constraints on the type of the fourth argument to `ifElse`).

svozza:
Ah yes, I'd wondered about that.

davidchambers:
:deciduous_tree:

index.js
```diff
@@ -1497,6 +1498,25 @@
...
 
+  //# ifElse :: (a -> Boolean) -> (a -> b) -> (a -> b) -> a -> b
+  //.
+  //. Takes a unary predicate, a unary "if" function, a unary "else"
+  //. function, and a value of any type, and returns the result of
+  //. applying the "if" function to the value if the value satisfies
+  //. the predicate; the result of applying the "else" function to the
+  //. value otherwise.
+  //.
+  // ```javascript
```
davidchambers: I just spent some time investigating a README rendering issue. Spot the missing character. ;)

svozza: Ooops!

PR #113, opened 2015-11-07
regexp: add S.regexEscape

davidchambers:
This will obviate the need for dependencies such as [escape-regexp][1] when using Sanctuary.

I'm open to name suggestions. `escapeRegexMetachars` is more descriptive, but feels too long.

I copied the regular expression from [XRegExp][2].

This is the first time I've used [JSVerify][3]. I'm impressed. We should write more tests this way!

[1]: https://www.npmjs.com/package/escape-regexp
[2]: https://github.com/slevithan/xregexp/blob/v3.0.0/xregexp-all.js#L792-L806
[3]: https://github.com/jsverify/jsverify

### DIFF
--- a/index.js
+++ b/index.js
@@ -2035,6 +2035,23 @@
     return new RegExp(source, flags);
   });
 
+  //# regexEscape :: String -> String
+  //.
+  //. Takes a string which may contain regular expression metacharacters,
+  //. and returns a string with those metacharacters escaped.
+  //.
+  //. Properties:
+  //.
+  //.   - `forall s :: String. S.test(S.regex('', S.regexEscape(s)), s) = true`
+  //.
+  //. ```javascript
+  //. > S.regexEscape('-=*{XYZ}*=-')
+  //. '\\-=\\*\\{XYZ\\}\\*=\\-'
+  //. ```
+  S.regexEscape = def('regexEscape', [String], function(s) {
+    return s.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');
+  });
+
   //# test :: RegExp -> String -> Boolean
   //.
   //. Takes a pattern and a string, and returns `true` if the pattern
--- a/test/index.js
+++ b/test/index.js
@@ -5,6 +5,7 @@
 var assert = require('assert');
 var vm = require('vm');
 
+var jsc = require('jsverify');
 var R = require('ramda');
 
 var S = require('..');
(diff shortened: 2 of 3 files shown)


index.js
```diff
@@ -2015,6 +2015,23 @@
...
   });
 
+  //# regexEscape :: String -> String
+  //.
+  //. Takes a string which may contain regular expression metacharacters,
+  //. and returns a string with those metacharacters escaped.
+  //.
+  //. Properties:
+  //.
+  //.   - ∀ `s` ∈ String: `S.test(S.regex('', S.regexEscape(s)), s)`
```
davidchambers: > Properties:
>
>   - ∀ `s` ∈ String: `S.test(S.regex('', S.regexEscape(s)), s)`

Is this a sensible way to express this property? How could the formatting be improved? Should we use this approach in the Ramda documentation as well?

/cc @CrossEye, @raine, @scott-christopher

svozza:
Nice, I was only using escape-regexp last week. JSVerify looks great too.

index.js
scott-christopher: Looks good to me, though I wonder whether `∀ s ∈ String` may be a little too foreign for others. I'd also suggest still including `=` in the property statement, despite `S.test` already returning a `Boolean` value.

e.g:
```
forall s :: String. S.test(S.regex('', S.regexEscape(s)), s) = true
```
It would be interesting to see how much effort would be involved in parsing these comments to be executed as actual tests with JSVerify.

davidchambers: Thanks for the feedback, Scott. I :heart: both `forall s :: String.` and `= true`. I'll make these changes. :)

michaelckelly: The notation makes sense! I agree with @scott-christopher in that the plaintext version is a bit more accessible. :+1:

test/index.js
```diff
@@ -5,6 +5,7 @@
 var assert = require('assert');
 var vm = require('vm');
 
+var jsc = require('jsverify');
```
michaelckelly: I wonder why they picked `jsc` as the standard import name? For `check`, maybe?

michaelckelly:
LGTM, @davidchambers.

test/index.js
davidchambers: I found the [answer][1]:

> ### Why do all the examples import the library as jsc instead of jsv?
>
> Does JSC originate with [JSCheck][2]?
>
> __A:__ Yes


[1]: https://github.com/jsverify/jsverify#why-do-all-the-examples-import-the-library-as-jsc-instead-of-jsv
[2]: http://www.jscheck.org/

index.js

PR #114, opened 2015-11-07
Add allPass and anyPass

svozza:
The most elegant implementation of these would be something like:

```js
S.allPass = (preds, a) => R.reduce((acc, f) => f(a) && acc, true, preds);
```
But then we'd lose short-circuiting.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1624,6 +1624,48 @@
     return pred(x) ? f(x) : g(x);
   });
 
+  //# allPass :: [a -> Boolean] -> a -> Boolean
+  //.
+  //. Takes an array of unary predicates and a value of any type
+  //. and returns `true` if all the predicates pass; `false` otherwise.
+  //. None of the subsequent predicates will be evaluated after the
+  //. first failed predicate.
+  //.
+  //. ```javascript
+  //. > S.allPass([S.test(/q/), S.test(/u/), S.test(/i/)], 'quiessence')
+  //. true
+  //.
+  //. > S.allPass([S.test(/q/), S.test(/u/), S.test(/i/)], 'fissiparous')
+  //. false
+  //. ```
+  S.allPass = def('allPass', [Array, a], function(preds, x) {
+    for (var idx = 0; idx < preds.length; idx += 1) {
+      if (!preds[idx](x)) return false;
+    }
+    return true;
+  });
+
+  //# anyPass :: [a -> Boolean] -> a -> Boolean
+  //.
+  //. Takes an array of unary predicates and a value of any type
+  //. and returns `true` if any of the predicates pass; `false` otherwise.
+  //. None of the subsequent predicates will be evaluated after the
+  //. first passed predicate.
+  //.
+  //. ```javascript
+  //. > S.anyPass([S.test(/q/), S.test(/u/), S.test(/i/)], 'incandescent')
+  //. true
+  //.
+  //. > S.anyPass([S.test(/q/), S.test(/u/), S.test(/i/)], 'empathy')
+  //. false
+  //. ```
+  S.anyPass = def('anyPass', [Array, a], function(preds, x) {
+    for (var idx = 0; idx < preds.length; idx += 1) {
+      if (preds[idx](x)) return true;
+    }
+    return false;
+  });
+
   //. ### List
 
   //# slice :: Integer -> Integer -> [a] -> Maybe [a]
--- a/test/index.js
+++ b/test/index.js
@@ -2175,6 +2175,164 @@ describe('control', function() {
 
   });
 
+  describe('allPass', function() {
+
+    it('is a binary function', function() {
+      eq(typeof S.allPass, 'function');
+      eq(S.allPass.length, 2);
+    });
+
+    it('type checks its arguments', function() {
+      assert.throws(function() { S.allPass('wrong'); },
+                    errorEq(TypeError,
+                            '‘allPass’ requires a value of type Array ' +
+                            'as its first argument; received "wrong"'));
+    });
+
+    it('returns true when given an empty array of predicates', function() {
+      eq(S.allPass([], 'quiessence'), true);
+    });
+
+    it('returns true when all predicates pass', function() {
+      eq(S.allPass([S.test(/q/),
+                    S.test(/u/),
+                    S.test(/i/),
+                    S.test(/c/)],
+                   'quiessence'),
+         true);
+    });
+
+    it('returns false when some predicates fail', function() {
+      eq(S.allPass([S.test(/q/),
+                    S.test(/u/),
+                    S.test(/i/),
+                    S.test(/c/)],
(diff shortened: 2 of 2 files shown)


kedashoe:
> But then we'd lose short-circuiting.

You could use `R.reduced`

```js
let allPass = (preds, a) => R.reduce((acc, f) => f(a) || R.reduced(false), true, preds);
let count = 0;
[
  allPass([
    () => ++count && true,
    () => ++count && false,
    () => ++count && true
  ]),
  allPass([
    () => ++count && true,
    () => ++count && true,
    () => ++count && true
  ]),
  count
]
```
http://goo.gl/gl1pio

svozza:
Nice!

svozza:
Ah, that desn't seem to play well with Sanctuary's type checking. The original works fine but when wrapped in the `def` function:
```
TypeError: ‘test’ requires a value of type String as its second argument; received {"name": "a"}
```

index.js
```diff
@@ -1497,6 +1497,48 @@
     return xBool !== yBool ? or(x, y) : xEmpty;
   });
 
+  //# allPass :: [a -> Boolean] -> a -> Boolean
+  //.
+  //. Takes an array of unary predicates and a value of any type
+  //. and returns true if all the predicates pass; false otherwise
```
davidchambers: Let's replace "true" and "false" with "`true`" and "`false`".

index.js
```diff
@@ -1497,6 +1497,48 @@
...
+  //. (none of the subsequent predicates will be evaluated after the
+  //. first failed predicate).
```
davidchambers: Let's make this a separate sentence.

test/index.js
```diff
@@ -1914,6 +1914,166 @@ describe('control', function() {
...
+    it('is a binary function', function() {
+      eq(typeof S.allPass, 'function');
+      eq(S.allPass.length, 2);
+    });
+
+    it('type checks its arguments', function() {
+      assert.throws(function() { S.allPass('wrong'); },
+        errorEq(TypeError,
+          '‘allPass’ requires a value of type Array ' +
+          'as its first argument; received "wrong"'));
```
davidchambers: :triangular_ruler:

test/index.js
```diff
@@ -1914,6 +1914,166 @@ describe('control', function() {
...
+    });
+
+    it('returns true when all predicates pass', function() {
+      eq(S.allPass([
+        S.test(/q/),
+        S.test(/u/),
+        S.test(/i/),
+        S.test(/c/)
+      ], 'quiessence'), true);
```
davidchambers: I'd prefer this formatting:

```javascript
eq(S.allPass([S.test(/q/),
              S.test(/u/),
              S.test(/i/),
              S.test(/c/)],
             'quiessence'),
   true);
```

Or this:

```javascript
eq(S.allPass([S.test(/q/), S.test(/u/), S.test(/i/), S.test(/c/)],
             'quiessence'),
   true);
```

test/index.js
```diff
@@ -1914,6 +1914,166 @@ describe('control', function() {
...
+    it('is a binary function', function() {
+      eq(typeof S.anyPass, 'function');
+      eq(S.anyPass.length, 2);
+    });
+
+    it('type checks its arguments', function() {
+      assert.throws(function() { S.anyPass('wrong'); },
+        errorEq(TypeError,
+          '‘anyPass’ requires a value of type Array ' +
+          'as its first argument; received "wrong"'));
```
davidchambers: :triangular_ruler:

davidchambers:
We should test the handling of an empty predicates list.

svozza:
What should the expected output be? An error? Or given that no predicates were passed in then none passed so it should always be false.

davidchambers:
`S.allPass([], x)` should evaluate `true` for every `x`: all (zero) of the predicates are guaranteed to be satisfied.

`S.anyPass([], x)` should evaluate `false` for every `x`: it's impossible for at least one predicate to be satisfied.

svozza:
Cool. Makes sense.

index.js
```diff
@@ -1497,6 +1497,48 @@
...
+    for (var idx = 0; idx < preds.length; idx += 1) {
+      if (!preds[idx](x)) return false;
+    }
+    return true;
+  });
+
+  //# anyPass :: [a -> Boolean] -> a -> Boolean
+  //.
+  //. Takes an array of unary predicates and a value of any type
+  //. and returns true if any of the predicates pass; false otherwise
```
davidchambers: Let's replace "true" and "false" with "`true`" and "`false`".

index.js
```diff
@@ -1497,6 +1497,48 @@
...
+  //. (none of the subsequent predicates will be evaluated after the
+  //. first passed predicate).
```
davidchambers: Let's make this a separate sentence.

test/index.js
```diff
@@ -1914,6 +1914,172 @@ describe('control', function() {
...
+    it('returns true when given an empty array of predicates', function() {
+      eq(S.allPass([], 'quiessence'), true);
+    });
+
+    it('returns true when all predicates pass', function() {
+      eq(S.allPass([S.test(/q/),
+                    S.test(/u/),
+                    S.test(/i/),
+                    S.test(/c/)],
+                    'quiessence'),
```
davidchambers: :triangular_ruler: `'quiessence'` should be nudged one space to the left to align with the `[`. :)

test/index.js
```diff
@@ -1914,6 +1914,172 @@ describe('control', function() {
...
+      eq(evaluated, false);
+    });
+
+    it('is curried', function() {
+      eq(S.allPass([]).length, 1);
+      eq(S.allPass([S.test(/q/),
+                    S.test(/u/),
+                    S.test(/i/),
+                    S.test(/c/)])('quiessence'),
+         true);
```
davidchambers: We could use a shorter expression here since we're just testing currying. Perhaps:

```javascript
eq(S.allPass([S.test(/q/)])('quiessence'), true);
```

davidchambers:
I'm sorry for getting behind on reviewing your pull requests, Stefano. Could you rebase and resolve the merge conflicts?

svozza:
I missed this comment altogether, just rebased now.

test/index.js
```diff
@@ -2175,6 +2175,164 @@ describe('control', function() {
...
+                    S.test(/i/),
+                    evaluate],
+                   'incandescent'),
+         false);
+      eq(evaluated, false);
+    });
+
+    it('is curried', function() {
+      eq(S.allPass([]).length, 1);
+      eq(S.allPass([S.test(/q/)])('quiessence'), true);
```
davidchambers: I think it'd read better if the first argument were `[S.test(/q/)]` in both cases (we follow this convention elsewhere).

davidchambers:
:deciduous_tree:

PR #115, opened 2015-11-08
Add S.not

svozza:
I haven't added anything about `S.lift` to the docs, I guess we can hold off on merging this until that's in.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1512,6 +1512,22 @@
     }
   };
 
+  //# not :: Boolean -> Boolean
+  //.
+  //. Takes a Boolean and returns the negation of that value
+  //. (`false` for `true`; `true` for `false`).
+  //.
+  //. ```javascript
+  //. > S.not(true)
+  //. false
+  //.
+  //. > S.not(false)
+  //. true
+  //. ```
+  S.not = def('not', [Boolean], function(x) {
+    return !x.valueOf();
+  });
+
   //# and :: a -> a -> a
   //.
   //. Takes two values of the same type and returns the second value
--- a/test/index.js
+++ b/test/index.js
@@ -1885,6 +1885,31 @@ describe('either', function() {
 
 describe('control', function() {
 
+  describe('not', function() {
+
+    it('is a unary function', function() {
+      eq(typeof S.not, 'function');
+      eq(S.not.length, 1);
+    });
+
+    it('can be applied to Booleans', function() {
+      eq(S.not(false), true);
+      eq(S.not(true), false);
+      // jshint -W053
+      eq(S.not(new Boolean(false)), true);
+      eq(S.not(new Boolean(true)), false);
+      // jshint +W053
+    });
+
+    it('throws when applied to a non-Boolean value', function() {
+      assert.throws(function() { S.not(0); },
+                    errorEq(TypeError,
+                            '‘not’ requires a value of type Boolean ' +
+                            'as its first argument; received 0'));
+    });
+
+  });
+
   describe('and', function() {
 
     it('is a binary function', function() {


davidchambers:
I'd been thinking this function would be of type `Boolean -> Boolean`. `toBoolean` is a requirement for types which implement Alternative, but I don't see `not` as relating to Alternative. Can you convince me otherwise? :)

svozza:
Well I actually envisaged it working your way too but then I saw how `and` and `or` were implemented and second-guessed myself. I'm happy to make this function more restrictive.

svozza:
Done.

test/index.js
```diff
@@ -1723,6 +1723,26 @@ describe('either', function() {
...
+
+    it('can be applied to Booleans', function() {
+      eq(S.not(false), true);
+      eq(S.not(true), false);
+    });
+
+    it('throws when supplied a non-boolean value', function() {
+      assert.throws(function() { S.not(0); },
+             errorEq(TypeError,
+                     '‘not’ requires a value of type Boolean as its first argument; received 0'));
```
davidchambers:
```javascript
assert.throws(function() { S.not(0); },
              errorEq(TypeError,
                      '‘not’ requires a value of type Boolean ' +
                      'as its first argument; received 0'));
```

test/index.js
```diff
@@ -1723,6 +1723,26 @@ describe('either', function() {
...
+
+    it('is a unary function', function() {
+      eq(typeof S.not, 'function');
+      eq(S.not.length, 1);
+    });
+
+    it('can be applied to Booleans', function() {
+      eq(S.not(false), true);
+      eq(S.not(true), false);
+    });
```
davidchambers: Let's also test `new Boolean(false)` and `new Boolean(true)`.

index.js
```diff
@@ -1437,6 +1437,21 @@
     }
   };
 
+  //# not :: Boolean -> Boolean
+  //.
+  //. Takes a boolean value and negates that value.
```
davidchambers: "Boolean" should be capitalized, as it's named after a person.

I'd like to be explicit about what the function returns:

_Takes a Boolean and returns the negation of that value (`false` for `true`; `true` for `false`)._

svozza: Can't get more explicit than that!

index.js
```diff
@@ -1437,6 +1437,22 @@
...
+  //. ```javascript
+  //. > S.not(true)
+  //. false
+  //.
+  //. > S.not(false)
+  //. true
+  //. ```
+  S.not = def('not', [Boolean], function(x) {
+    return !x.valueOf();
+  });
```
davidchambers: I believe `.valueOf()` is redundant here, since `!` will trigger a call to that method:

```javascript
> new Boolean(true)
new Boolean(true)

> !new Boolean(true)
false
```

The function can become `function(x) { return !x; }`, which is fairly nice even without arrow functions. :)

test/index.js
```diff
@@ -1723,6 +1723,31 @@ describe('either', function() {
...
+    it('can be applied to Booleans', function() {
+      eq(S.not(false), true);
+      eq(S.not(true), false);
+      // jshint -W053
+      eq(S.not(new Boolean(false)), true);
+      eq(S.not(new Boolean(true)), false);
+      // jshint +W053
+    });
+
+    it('throws when supplied a non-boolean value', function() {
```
davidchambers: I'd prefer _applied to_ in place of _supplied_. Also, let's capitalize Boolean.

index.js
svozza: Without `valueOf` the function will fail to evaluate correctly when passed a Boolean created with `new` in this case:

```js
S.not(new Boolean(false)) // returns false instead of true
```
I guess it's because objects are truthy so !(new Boolean(false)) will always evaluate to `false`.

davidchambers: Oh, you're right! It's good that we have tests to prevent us from forgetting JavaScript's more questionable behaviour!

test/index.js
```diff
@@ -1723,6 +1723,31 @@ describe('either', function() {
...
+      eq(S.not(true), false);
+      // jshint -W053
+      eq(S.not(new Boolean(false)), true);
+      eq(S.not(new Boolean(true)), false);
+      // jshint +W053
+    });
+
+    it('throws when applied to a non-Boolean value', function() {
+      assert.throws(function() { S.not(0); },
+             errorEq(TypeError,
```
davidchambers: :triangular_ruler: `errorEq` should be aligned with `function`.

svozza:
Is there anything more I need to do with this one?

svozza:
Or do you want to hold off until the `sanctuary-def` PR is in?

davidchambers:
:deciduous_tree:

Issue #118, opened 2015-11-17
maybe: implement Traversable


davidchambers:
@scott-christopher defined [Foldable][1] for Maybe in #69. It would be useful to define [Traversable][2] as well, particularly in light of ramda/ramda#1467.

[1]: https://github.com/fantasyland/fantasy-land#foldable
[2]: https://github.com/fantasyland/fantasy-land#traversable

davidchambers:
Thanks again for the contribution, @scott-christopher!

PR #121, opened 2015-11-17
Add traversable instance for Maybe

scott-christopher:
Implements #118 

As per the included docs:

**Maybe#sequence :: Applicative f => Maybe (f a) ~> (b -> f b) -> f (Maybe a)**

Evaluates an applicative action contained within the Maybe, resulting in:
  - an applicative of `Just` the value of the evaluated action if `this`
    is a `Just`.
  - a pure applicative of `Nothing` if `this` is a `Nothing`.
```javascript
> S.Nothing().sequence(S.Either.of)
Right(Nothing())

> S.Just(Right(42)).sequence(S.Either.of)
Right(Just(42))
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -540,8 +540,8 @@
   //. The Maybe type represents optional values: a value of type `Maybe a` is
   //. either a Just whose value is of type `a` or a Nothing (with no value).
   //.
-  //. The Maybe type satisfies the [Monoid][], [Monad][], [Foldable][], and
-  //. [Extend][] specifications.
+  //. The Maybe type satisfies the [Monoid][], [Monad][], [Traversable][],
+  //. and [Extend][] specifications.
   var Maybe = S.Maybe = function Maybe() {
     if (arguments[0] !== sentinel) {
       throw new Error('Cannot instantiate Maybe');
@@ -829,6 +829,29 @@
          [$Maybe(a), $.Function, b, b],
          function(maybe, f, x) {
            return maybe.isJust ? f(x, maybe.value) : x;
+         });
+
+  //# Maybe#sequence :: Applicative f => Maybe (f a) ~> (a -> f a) -> f (Maybe a)
+  //.
+  //. Evaluates an applicative action contained within the Maybe, resulting in:
+  //.
+  //.   - a pure applicative of a Nothing if `this` is a Nothing; otherwise
+  //.
+  //.   - an applicative of Just the value of the evaluated action.
+  //.
+  //. ```javascript
+  //. > S.Nothing().sequence(S.Either.of)
+  //. Right(Nothing())
+  //.
+  //. > S.Just(Right(42)).sequence(S.Either.of)
+  //. Right(Just(42))
+  //. ```
+  Maybe.prototype.sequence =
+  method('Maybe#sequence',
+         {a: [Apply], b: [Apply]},
+         [$Maybe(a), $.Function, b],
+         function(maybe, of) {
+           return maybe.isJust ? R.map(Just, maybe.value) : of(maybe);
          });
 
   //# Maybe#toBoolean :: Maybe a ~> Boolean
@@ -2491,10 +2514,10 @@
 
 //. [Apply]:        https://github.com/fantasyland/fantasy-land#apply
 //. [Extend]:       https://github.com/fantasyland/fantasy-land#extend
-//. [Foldable]:     https://github.com/fantasyland/fantasy-land#foldable
 //. [Functor]:      https://github.com/fantasyland/fantasy-land#functor
 //. [Monad]:        https://github.com/fantasyland/fantasy-land#monad
 //. [Monoid]:       https://github.com/fantasyland/fantasy-land#monoid
+//. [Traversable]:  https://github.com/fantasyland/fantasy-land#traversable
 //. [R.equals]:     http://ramdajs.com/docs/#equals
 //. [R.map]:        http://ramdajs.com/docs/#map
 //. [R.type]:       http://ramdajs.com/docs/#type
--- a/test/index.js
+++ b/test/index.js
@@ -67,6 +67,79 @@ var squareRoot = function(n) {
                : S.Right(Math.sqrt(n));
 };
 
+//  Identity :: a -> Identity a
+var Identity = function(x) {
+  return {
+    map: function(fn) {
+      return Identity(fn(x));
+    },
+    ap: function(y) {
+      return Identity(x(y));
+    },
+    equals: function(other) {
+      return R.equals(x, other.value);
+    },
+    value: x
+  };
+};
+Identity.of = Identity;
+
+//  Compose :: Apply f, Apply g
+//          => { of: b -> f b } -> { of: c -> g c }
+//          -> f (g a) -> Compose f g a
+var Compose = function(F, G) {
+  function _Compose(x) {
+    return {
+      map: function(f) {
+        return _Compose(R.map(R.map(f), x));
+      },
+      ap: function(y) {
+        return _Compose(R.ap(R.map(R.ap, x), y.value));
+      },
+      equals: function(other) {
+        return R.equals(x, other.value);
+      },
+      value: x
+    };
+  }
+  _Compose.of = function(x) {
+    return _Compose(F.of(G.of(x)));
+  };
+  return _Compose;
+};
+
+//  eitherToMaybe :: Either a b -> Maybe b
+var eitherToMaybe = S.either(S.K(S.Nothing()), S.Just);
+
+//  JustArb :: Arbitrary a -> Arbitrary (Maybe a)
+var JustArb = function(arb) {
+  return arb.smap(S.Just, function(m) { return m.value; }, R.toString);
+};
+
+//  MaybeArb :: Arbitrary a -> Arbitrary (Maybe a)
+var MaybeArb = function(arb) {
+  return jsc.oneof(JustArb(arb), jsc.constant(S.Nothing()));
+};
+
+//  LeftArb :: Arbitrary a -> Arbitrary (Either a b)
+var LeftArb = function(arb) {
+  return arb.smap(S.Left, function(e) { return e.value; }, R.toString);
+};
+
+//  RightArb :: Arbitrary a -> Arbitrary (Either b a)
+var RightArb = function(arb) {
+  return arb.smap(S.Right, function(e) { return e.value; }, R.toString);
+};
+
+//  EitherArb :: Arbitrary a -> Arbitrary b -> Arbitrary (Either a b)
+var EitherArb = function(lArb, rArb) {
+  return jsc.oneof(LeftArb(lArb), RightArb(rArb));
+};
+
+//  IdentityArb :: Arbitrary a -> Arbitrary (Identity a)
+var IdentityArb = function(arb) {
+  return arb.smap(Identity, function(i) { return i.value; });
+};
 
 describe('invariants', function() {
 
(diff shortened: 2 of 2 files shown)


davidchambers:
That was quick, Scott! Thanks!

index.js
```diff
@@ -485,8 +485,8 @@
   //. The Maybe type represents optional values: a value of type `Maybe a` is
   //. either a Just whose value is of type `a` or a Nothing (with no value).
   //.
-  //. The Maybe type satisfies the [Monoid][], [Monad][], [Foldable][], and
-  //. [Extend][] specifications.
+  //. The Maybe type satisfies the [Monoid][], [Monad][], [Foldable][],
+  //. [Traversable][] and [Extend][] specifications.
```
davidchambers: Traversable implies Foldable, so we could remove Foldable from the list in the interest of brevity.

scott-christopher: Good point – updated.

index.js
```diff
@@ -485,7 +485,7 @@
   //. The Maybe type represents optional values: a value of type `Maybe a` is
   //. either a Just whose value is of type `a` or a Nothing (with no value).
   //.
-  //. The Maybe type satisfies the [Monoid][], [Monad][], [Foldable][], and
+  //. The Maybe type satisfies the [Monoid][], [Monad][], [Traversable][] and
```
davidchambers: Can we keep the Oxford comma? :)

scott-christopher: :zap:

index.js
```diff
@@ -2159,4 +2187,5 @@
 //. [Ramda]:        http://ramdajs.com/
 //. [RegExp]:       https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp
 //. [Semigroup]:    https://github.com/fantasyland/fantasy-land#semigroup
+//. [Traversable]:  https://github.com/fantasyland/fantasy-land#traversable
```
davidchambers: Can we remove the link definition for `[Foldable]`? I believe we're removing the only occurrence.

test/index.js
```diff
@@ -47,6 +47,45 @@ var squareRoot = function(n) {
...
+    };
+  }
+  _Compose.of = function(x) {
+    return _Compose(F.of(G.of(x)));
+  };
+  return _Compose;
+};
+
+//  eitherToMaybe :: Either a b -> Maybe b
+var eitherToMaybe = S.either(R.always(S.Nothing()), S.Just);
```
davidchambers: Let's use `S.K` in place of `R.always`.

test/index.js
```diff
@@ -579,6 +618,29 @@ describe('maybe', function() {
       assert(a.chain(a.of).equals(a));
     });
 
+    it('implements Traversable', function() {
+      var u = S.Nothing(),
+          C = Compose(Identity, S.Maybe),
+          lhs, rhs;
```
davidchambers: I prefer one `var` keyword per declaration, though I don't mind grouping the uninitialized variables:

```javascript
var u = S.Nothing();
var C = Compose(Identity, S.Maybe);
var lhs, rhs;
```

index.js
```diff
@@ -714,6 +714,23 @@
...
+  //.
+  //.   - a pure applicative of `Nothing` if `this` is a `Nothing`.
+  //.
+  //. ```javascript
+  //. > S.Nothing().sequence(S.Either.of)
+  //. Right(Nothing())
+  //.
+  //. > S.Just(Right(42)).sequence(S.Either.of)
+  //. Right(Just(42))
+  //. ```
```
davidchambers: Nice examples!

index.js
scott-christopher: :zap:

test/index.js
scott-christopher: :zap:

test/index.js
scott-christopher: :zap:

index.js
```diff
@@ -863,6 +884,13 @@
     return f(z, this.value);
   });
 
+  //  Just#sequence :: Applicative f => Maybe (f a) ~> (b -> f b) -> f (Maybe a)
+  Just.prototype.sequence =
+    def('Just#sequence', [Function], function(of) {
+      void of;
```
davidchambers: `make lint` passes for me without this line, thanks to #31. Was it failing for you and/or CircleCI?

test/index.js
```diff
@@ -825,6 +887,30 @@ describe('maybe', function() {
...
+
+      // composition
+      u = S.Just(Identity(S.Just(42)));
+      lhs = u.map(C).sequence(C.of);
+      rhs = C(u.sequence(Identity.of).map(function(x) {
+        return x.sequence(S.Maybe.of);
+      }));
+      assert(lhs.equals(rhs));
+    });
   });
```
davidchambers: Please add an empty line between lines 913 and 914.

index.js
```diff
@@ -714,6 +714,23 @@
...
   //. ```
 
+  //# Maybe#sequence :: Applicative f => Maybe (f a) ~> (b -> f b) -> f (Maybe a)
+  //.
+  //. Evaluates an applicative action contained within the Maybe, resulting in:
+  //.
+  //.   - an applicative of `Just` the value of the evaluated action if `this`
+  //.     is a `Just`.
+  //.
+  //.   - a pure applicative of `Nothing` if `this` is a `Nothing`.
```
davidchambers: It would be nice to mirror the structure of the description of `Maybe#reduce`:

> Evaluates an applicative action contained within the Maybe, resulting in:
>
>   - a pure applicative of a Nothing if `this` is a Nothing; otherwise
>
>   - an applicative of Just the value of the evaluated action.

davidchambers:
Lovely pull request, Scott. Now that I've appeased my pedantic side it's time to spend some time studying these laws to make sure I understand them. :)

index.js
scott-christopher: I've started doing this to make it clear that the argument isn't used. Happy to remove it.

index.js
scott-christopher: :zap:

test/index.js
scott-christopher: :zap:

index.js
scott-christopher: :zap: I've removed the `void of`

davidchambers: Okay. I wouldn't have minded keeping it in that case. I assumed we were appeasing some overeager linter. :)

scott-christopher:
This loose translation from [here](https://hackage.haskell.org/package/base-4.8.1.0/docs/Data-Traversable.html#t:Traversable) _might_ help with interpreting the laws:

---

A definition of `sequence` must satisfy the following laws:

**naturality**
* `compose(t, sequence) = compose(sequence, map(t))` for every applicative transformation `t`, where an applicative transformation is a function: `t :: (Applicative f, Applicative g) => f a -> g a`

**identity**
* `compose(sequence, map(Identity)) = Identity`

**composition**
* `compose(sequence, map(Compose)) = compose(Compose, map(sequence), sequence)`

test/index.js
```diff
@@ -47,6 +47,45 @@ var squareRoot = function(n) {
...
+    },
+    value: x
+  };
+};
+Identity.of = Identity;
+
+var Compose = function(F, G) {
+  function _Compose(x) {
+    return {
+      map: function(fn) {
```
davidchambers: I've a slight preference for `f` over `fn`. Mind changing this, Scott?

test/index.js
```diff
@@ -579,6 +618,29 @@ describe('maybe', function() {
       assert(a.chain(a.of).equals(a));
     });
 
+    it('implements Traversable', function() {
+      var u = S.Nothing();
+      var C = Compose(Identity, S.Maybe);
```
davidchambers: Should this be `S.Maybe.of` rather than `S.Maybe`?

test/index.js
scott-christopher: :zap:

test/index.js
scott-christopher: You can think of the arguments to `Compose` as the types that are expected to have an `of` method attached which are used to implement `of` for `Compose`, so we want to pass `Identity` and `S.Maybe` here.

```js
_Compose.of = function(x) {
  return _Compose(F.of(G.of(x)));
};
```

davidchambers: Oh I see. Thanks for the explanation. :)

davidchambers:
I'm sorry for neglecting this pull request, Scott! Would you mind rebasing?

scott-christopher:
> I'm sorry for neglecting this pull request

I forgot about it myself :)

I've rebased on to master and also updated the tests for `sequence` to use jsverify.

index.js
```diff
@@ -831,6 +831,29 @@
...
+  //. > S.Nothing().sequence(S.Either.of)
+  //. Right(Nothing())
+  //.
+  //. > S.Just(Right(42)).sequence(S.Either.of)
+  //. Right(Just(42))
+  //. ```
+  Maybe.prototype.sequence =
+  method('Maybe#sequence',
+         {a: [Apply], b: [Apply]},
+         [$Maybe(a), $.Function, b],
```
scott-christopher: Is there a better way to represent `Apply f => Maybe (f a) -> (a -> f a) -> f (Maybe a)` here?

davidchambers: I don't think so. We're quite a bit beyond the city limits of sanctuary-def at this point. ;)

davidchambers:
>       1) maybe Maybe implements Traversable:
>          Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.

Looks as though we need a bit more :hourglass:.

scott-christopher:
I've split them out into separate `it` blocks to give them each a little more time to run.

index.js
```diff
@@ -540,8 +540,8 @@
   //. The Maybe type represents optional values: a value of type `Maybe a` is
   //. either a Just whose value is of type `a` or a Nothing (with no value).
   //.
-  //. The Maybe type satisfies the [Monoid][], [Monad][], [Foldable][], and
-  //. [Extend][] specifications.
+  //. The Maybe type satisfies the [Monoid][], [Monad][], [Foldable][],
+  //. [Traversable][], and [Extend][] specifications.
```
davidchambers: This came back to life during the rebase:

> Traversable implies Foldable, so we could remove Foldable from the list in the interest of brevity.

I believe this comment still applies:

> Can we remove the link definition for `[Foldable]`? I believe we're removing the only occurrence.

test/index.js
```diff
@@ -67,6 +67,69 @@ var squareRoot = function(n) {
                : S.Right(Math.sqrt(n));
 };
 
+var Identity = function(x) {
```
davidchambers: A type signature here would be nice:

```javascript
//  Identity :: a -> Identity a
var Identity = ...;
```

test/index.js
```diff
@@ -478,6 +541,37 @@ describe('maybe', function() {
...
+          var lhs = u.map(C).sequence(C.of);
+          var rhs = C(u.sequence(Identity.of).map(function(x) {
+            return x.sequence(S.Maybe.of);
+          }));
+          return lhs.equals(rhs);
+        }));
+      });
+
+    });
+
```
davidchambers: I :heart: these tests!

index.js
```diff
@@ -831,6 +831,29 @@
...
+  //.
+  //. > S.Just(Right(42)).sequence(S.Either.of)
+  //. Right(Just(42))
+  //. ```
+  Maybe.prototype.sequence =
+  method('Maybe#sequence',
+         {a: [Apply], b: [Apply]},
+         [$Maybe(a), $.Function, b],
+         function(maybe, of) {
+           return maybe.isJust ? maybe.value.map(Just) : of(maybe);
```
davidchambers: I think we should use `R.map(Just, maybe.value)` here, to support environments without `Array.prototype.map`.

scott-christopher: Good idea. :zap:

test/index.js
scott-christopher: Sure. :zap: I've included signatures for `Compose` and the Arbitrary constructors too.

test/index.js
```diff
@@ -67,6 +67,79 @@ var squareRoot = function(n) {
...
+      return R.equals(x, other.value);
+    },
+    value: x
+  };
+};
+Identity.of = Identity;
+
+//  Compose :: Apply f, Apply g
+//          => { of: b -> f b } -> { of: c -> g c }
+//          -> f (g a) -> Compose f g a
```
davidchambers: This is instructive!

index.js
scott-christopher: Haha, I looked at this yesterday and couldn't remember why `Foldable` was removed previously and presumed `reduce` hadn't yet landed at that time. That'll teach me for not looking at the comment history :smile: 

In any case, I've removed it again.

davidchambers:
:deciduous_tree:

PR #122, opened 2015-12-09
Add an 'inspect' method to all types

CrisFeo:
This PR simply proxies the `toString` method onto the `inspect` method to
make the Sanctuary types more REPL-friendly.

The Node REPL uses `util.inspect` when rendering objects for display.
This method defers to the `inspect` method on an object if it exists.
[source](https://nodejs.org/api/repl.html#repl_customizing_object_displays_in_the_repl)

### DIFF
--- a/index.js
+++ b/index.js
@@ -738,6 +738,22 @@
   //. 'Just([1, 2, 3])'
   //. ```
 
+  //# Maybe#inspect :: Maybe a ~> String
+  //.
+  //. Returns the string representation of the Maybe. This method is used by
+  //. `util.inspect` and the REPL to format a Maybe for display.
+  //.
+  //. See also [`Maybe#toString`](#Maybe.prototype.toString).
+  //.
+  //. ```javascript
+  //. > S.Nothing().inspect()
+  //. 'Nothing()'
+  //.
+  //. > S.Just([1, 2, 3]).inspect()
+  //. 'Just([1, 2, 3])'
+  //. ```
+  Maybe.prototype.inspect = function() { return this.toString(); };
+
   //# Nothing :: -> Maybe a
   //.
   //. Returns a Nothing. Though this is a constructor function the `new`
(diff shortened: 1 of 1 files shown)


davidchambers:
This will be so helpful!

index.js
```diff
@@ -796,6 +796,9 @@
   Nothing.prototype.toString = def('Nothing#toString', [],
                                    R.always('Nothing()'));
 
+  //  Nothing#inspect :: Maybe a ~> String
+  Nothing.prototype.inspect = R.nAry(0, Nothing.prototype.toString);
+
```
davidchambers: Two questions:

  - Is there a reason to use `R.nAry(0)` here rather than simply assign `Nothing#toString` to `Nothing#inspect`?

  - Rather than define `Nothing#inspect` and `Just#inspect`, could we define `Maybe#inspect`?

    ```javascript
    Maybe.prototype.inspect = def('Maybe#inspect', [], R.invoker(0, 'toString'));
    ```

CrisFeo: * `inspect` will be provided with two arguments when called by the REPL and `toString` is nullary so we do need to control the arguments in some fashion.

* Agreed, it makes more sense attached to `Maybe`, :zap:ing.

index.js
```diff
@@ -738,6 +738,21 @@
   //. 'Just([1, 2, 3])'
   //. ```
 
+  //# Maybe#inspect :: Maybe a ~> String
+  //.
+  //. Returns the string representation of the Maybe. This function is
+  //. identical to the `toString` method. It is used by `util.inspect` and the
+  //. REPL to format a Maybe for display.
```
davidchambers: This is not *identical* to the `toString` method. We could remove the second sentence and add a second paragraph:

```javascript
  //. See also [`Maybe#toString`](#Maybe.prototype.toString).
```

CrisFeo: Ah, true :zap:

davidchambers:
:deciduous_tree: Lovely patch, @CrisFeo.

soroushjp:
Woohoo nice work @CrisFeo !

PR #117, opened 2015-11-10
Add converge function

svozza:

PR #126, opened 2016-01-05
add lift, lift2, lift3

benperez:
@davidchambers, Wanted to see you what thought of `S.lift` before I go ahead and throw in `S.lift2` and `S.lift3`.

`S.lift` is pretty much just a wrapper for `R.map`, but `S.lift2` and `S.lift3` also use `R.ap`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -419,6 +419,57 @@
     return f(y, x);
   });
 
+  //# lift :: Functor f => (a -> b) -> f a -> f b
+  //.
+  //. Promotes a unary function to a function which operates on a [Functor][].
+  //.
+  //. ```javascript
+  //. > S.lift(R.inc, S.Just(2))
+  //. S.Just(3)
+  //.
+  //. > S.lift(R.inc, S.Nothing())
+  //. S.Nothing()
+  //. ```
+  S.lift = def('lift', [Function, a], R.map);
+
+  //# lift2 :: Apply f => (a -> b -> c) -> f a -> f b -> f c
+  //.
+  //. Promotes a binary function to a function which operates on two
+  //. [Apply][]s.
+  //.
+  //. ```javascript
+  //. > S.lift2(R.add, S.Just(2), S.Just(3))
+  //. S.Just(5)
+  //.
+  //. > S.lift2(R.add, S.Just(2), S.Nothing())
+  //. S.Nothing()
+  //.
+  //. > S.lift2(S.and, S.Just(true), S.Just(true))
+  //. S.Just(true)
+  //.
+  //. > S.lift2(S.and, S.Just(true), S.Just(false))
+  //. S.Just(false)
+  //. ```
+  S.lift2 = def('lift2', [Function, a, b], function(f, x, y) {
+    return R.ap(R.map(f, x), y);
+  });
+
+  //# lift3 :: Apply f => (a -> b -> c -> d) -> f a -> f b -> f c -> f d
+  //.
+  //. Promotes a ternary function to a function which operates on three
+  //. [Apply][]s.
+  //.
+  //. ```javascript
+  //. > S.lift3(R.reduce, S.Just(R.add), S.Just(0), S.Just([1, 2, 3]))
+  //. S.Just(6)
+  //.
+  //. > S.lift3(R.reduce, S.Just(R.add), S.Just(0), S.Nothing())
+  //. S.Nothing()
+  //. ```
+  S.lift3 = def('lift3', [Function, a, b, c], function(f, x, y, z) {
+    return R.ap(R.ap(R.map(f, x), y), z);
+  });
+
   //. ### Composition
 
   //# compose :: (b -> c) -> (a -> b) -> a -> c
@@ -2296,8 +2347,10 @@
 
 }));
 
+//. [Apply]:        https://github.com/fantasyland/fantasy-land#apply
 //. [Extend]:       https://github.com/fantasyland/fantasy-land#extend
 //. [Foldable]:     https://github.com/fantasyland/fantasy-land#foldable
+//. [Functor]:      https://github.com/fantasyland/fantasy-land#functor
 //. [Monad]:        https://github.com/fantasyland/fantasy-land#monad
 //. [Monoid]:       https://github.com/fantasyland/fantasy-land#monoid
 //. [R.equals]:     http://ramdajs.com/docs/#equals
--- a/test/index.js
+++ b/test/index.js
@@ -293,6 +293,87 @@ describe('function', function() {
 
   });
 
+  describe('lift', function() {
+
+    it('is a binary function', function() {
+      eq(typeof S.lift, 'function');
+      eq(S.lift.length, 2);
+    });
+
+    it('type checks its arguments', function() {
+      assert.throws(function() { S.lift('wrong'); },
+                    errorEq(TypeError,
+                            '‘lift’ requires a value of type Function ' +
+                            'as its first argument; received "wrong"'));
+    });
+
+    it('lifts a function into the context of Functors', function() {
+      eq(S.lift(R.multiply(2), S.Just(3)), S.Just(6));
+      eq(S.lift(R.multiply(2), S.Nothing()), S.Nothing());
+
+      eq(S.lift(R.multiply(2), S.Left(3)), S.Left(3));
+      eq(S.lift(R.multiply(2), S.Right(3)), S.Right(6));
+
+      eq(S.lift(R.multiply(2), [1, 2, 3]), [2, 4, 6]);
+      eq(S.lift(R.multiply(2), []), []);
+    });
+
+  });
+
+  describe('lift2', function() {
+
+    it('is a ternary function', function() {
+      eq(typeof S.lift2, 'function');
+      eq(S.lift2.length, 3);
+    });
+
+    it('type checks its arguments', function() {
+      assert.throws(function() { S.lift2('wrong'); },
+                    errorEq(TypeError,
+                            '‘lift2’ requires a value of type Function ' +
+                            'as its first argument; received "wrong"'));
+    });
+
+    it('lifts a function into the context of Applys', function() {
+      eq(S.lift2(R.add, S.Just(3), S.Just(3)), S.Just(6));
+      eq(S.lift2(R.add, S.Nothing(), S.Just(3)), S.Nothing());
+
+      eq(S.lift2(R.add, S.Right(3), S.Left(4)), S.Left(4));
+      eq(S.lift2(R.add, S.Right(3), S.Right(4)), S.Right(7));
+
+      eq(S.lift2(R.add, [1, 2], [10, 20]), [11, 21, 12, 22]);
+      eq(S.lift2(R.add, [], [1, 2]), []);
+    });
+
+  });
+
+  describe('lift3', function() {
+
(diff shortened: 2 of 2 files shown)


davidchambers:
I'll add a few in-line comments, but this looks great! I'm really excited about these three functions. :)

benperez:
> I'll add a few in-line comments

I had a hunch :stuck_out_tongue_winking_eye:

index.js
```diff
@@ -419,6 +419,21 @@
...
+  //.
+  //. Takes a unary function and transforms it into a unary function in another
+  //. setting.
+  //.
+  //. ```javascript
+  //. > S.lift(R.add(1))(S.Just(2))
+  //. S.Just(3)
+  //. > S.lift(R.add(1))(S.Nothing())
+  //. S.Nothing()
+  //. ```
```
davidchambers: Please include an empty—well, `//.`—line between doctests.

Let's use `R.inc` rather than `R.add(1)`.

Let's use the `S.lift(f, x)` form rather than the `S.lift(f)(x)` one. This highlights a difference between `R.lift` and `S.lift`.

index.js
```diff
@@ -419,6 +419,21 @@
...
+  S.lift = def('lift', [Function, a], function(f, a) {
+    return R.map(f, a);
+  });
```
davidchambers: `function(f, a) { return R.map(f, a); }` looks like unnecessary function wrapping. ;)

test/index.js
```diff
@@ -293,6 +293,32 @@ describe('function', function() {
 
   });
 
+  describe('lift', function() {
+    it('is a binary function', function() {
+      eq(typeof S.lift, 'function');
+      eq(S.lift.length, 2);
+    });
+
+    it('throws if applied to values of different types', function() {
```
davidchambers: The description should be `'type checks its arguments'`.

test/index.js
```diff
@@ -293,6 +293,32 @@ describe('function', function() {
...
+    it('lifts a function into the context of functors', function() {
+      eq(S.lift(R.multiply(2))(S.Just(3)), S.Just(6));
+      eq(S.lift(R.multiply(2))(S.Nothing()), S.Nothing());
+
+      eq(S.lift(R.multiply(2))(S.Left(3)), S.Left(3));
+      eq(S.lift(R.multiply(2))(S.Right(3)), S.Right(6));
+
+      eq(S.lift(R.multiply(2))([1, 2, 3]), [2, 4, 6]);
+      eq(S.lift(R.multiply(2))([]), []);
+    });
```
davidchambers: Let's use the `S.lift(f, x)` form in these examples too.

index.js
```diff
@@ -419,6 +419,21 @@
     return f(y, x);
   });
 
+  //# lift :: Applicative f => (a -> b) -> f a -> f b
+  //.
+  //. Takes a unary function and transforms it into a unary function in another
+  //. setting.
```

_Takes a unary function and returns the corresponding unary function in another setting._

I'd love to say more, but this is as much of a description as my current understanding permits.

benperez:
Addressed comments and added `S.lift2` and `S.lift3`. :muscle: :muscle: :muscle:

index.js
```diff
@@ -419,6 +419,52 @@
...
+  //. ```
+  S.lift = def('lift', [Function, a], R.map);
+
+  //# Applicative f => (a -> b -> c) -> f a -> f b -> f c
+  //.
+  //. Takes a binary function and promotes it to a function which operates on
+  //. two applicative functors.
+  //.
+  //. ```javascript
+  //. > S.lift2(R.add, S.Just(2), S.Just(3))
```
svozza: Could we add an example using `S.and` in the example to show people how to roll their own typesafe version of `R.both` as discussed in #110?

svozza:
Nice. This means we can close #117 too.

index.js
```diff
@@ -419,6 +419,52 @@
     return f(y, x);
   });
 
+  //# lift :: Applicative f => (a -> b) -> f a -> f b
```
davidchambers: Should this be `Apply` rather than `Applicative`? We don't need the full power of Applicative, do we?

index.js
```diff
@@ -419,6 +419,52 @@
...
+  //.
+  //. Takes a unary function and promotes it to a function which operates on
+  //. a functor.
```
davidchambers: Nice! This is much better. We could shorten it a little:

_Promotes a unary function to a function which operates on a functor._

Also, should we say "an [Apply][1]" rather than "a functor"?


[1]: https://github.com/fantasyland/fantasy-land#apply

index.js
davidchambers: :+1:

benperez:
@davidchambers regarding `Apply` vs `Applicative`: I did some research and here is what I figured out—According to the Fantasy-Land spec it seems like the Haskell idea of `Applicative` is split into two parts in fantasy-land. 

Haskell's Applicative (see [minimal complete definition](https://hackage.haskell.org/package/base-4.8.1.0/docs/Control-Applicative.html))) calls for `pure` and `<*>` (or `ap`).

FL has `Apply`—which calls for `ap`—and `Applicative`—which calls for `of` (aka `pure`).

I suppose lift really only needs `ap` so we could just use `Apply` here. It jives with FL but it might confuse some people coming from Haskell.

davidchambers:
> I suppose lift really only needs `ap` so we could just use `Apply` here. It jives with FL but it might confuse some people coming from Haskell.

Aligning with Fantasy Land is more important than aligning with Haskell. We can link to the relevant section of the FL spec to explain what we mean by "Apply".

benperez:
> Aligning with Fantasy Land is more important than aligning with Haskell.

:+1: will :zap: now

index.js
davidchambers: In fact, am I correct in saying that `lift2` and `lift3` require [Apply][1] but `lift` merely requires [Functor][2]? This suggests:

```haskell
lift  :: Functor f => (a -> b) -> f a -> f b
lift2 :: Apply f => (a -> b -> c) -> f a -> f b -> f c
lift3 :: Apply f => (a -> b -> c -> d) -> f a -> f b -> f c -> f d
```


[1]: https://github.com/fantasyland/fantasy-land#apply
[2]: https://github.com/fantasyland/fantasy-land#functor

benperez:
:zap:

index.js
```diff
@@ -419,6 +419,55 @@
     return f(y, x);
   });
 
+  //# lift :: Apply f => (a -> b) -> f a -> f b
+  //.
+  //. Promotes a unary function to a function which operates on a Functor.
```
davidchambers: Let's use `[Functor][]` and include `[Functor]: https://...` at the bottom of the file.

index.js
```diff
@@ -419,6 +419,55 @@
...
+  //. S.Just(3)
+  //.
+  //. > S.lift(R.inc, S.Nothing())
+  //. S.Nothing()
+  //. ```
+  S.lift = def('lift', [Function, a], R.map);
+
+  //# lift2 :: Apply f => (a -> b -> c) -> f a -> f b -> f c
+  //.
+  //. Promotes a binary function to a function which operates on two Apply's.
```
davidchambers: Let's use `[Apply][]s`. I don't think an apostrophe is appropriate here (though I agree that "Applys" looks silly).

index.js
```diff
@@ -419,6 +419,55 @@
...
+  //. S.Just(5)
+  //.
+  //. > S.lift2(R.add, S.Just(2), S.Nothing())
+  //. S.Nothing()
+  //.
+  //. > S.lift2(S.and, S.Just(true), S.Just(true))
+  //. S.Just(true)
+  //.
+  //. > S.lift2(S.and, S.Nothing(), S.Just(false))
+  //. S.Nothing()
```
davidchambers: Let's replace this example with `S.lift2(S.and, S.Just(true), S.Just(false))` since one of the preceding examples demonstrates the `Nothing()` behaviour.

index.js
```diff
@@ -419,6 +419,55 @@
...
+  //.
+  //. Promotes a ternary function to a function which operates on three Apply's.
+  //.
+  //. ```javascript
+  //. > S.lift3(R.reduce, S.Just(R.add), S.Just(0), S.Just([1, 2, 3]))
+  //. S.Just(6)
+  //.
+  //. > S.lift3(R.reduce, S.Just(R.add), S.Just(0), S.Nothing())
+  //. S.Nothing()
+  //. ```
```
davidchambers: Very nice examples!

test/index.js
```diff
@@ -293,6 +293,84 @@ describe('function', function() {
...
+
+    it('type checks its arguments', function() {
+      assert.throws(function() { S.lift2('wrong'); },
+                    errorEq(TypeError,
+                            '‘lift2’ requires a value of type Function as its ' +
+                            'first argument; received "wrong"'));
+    });
+
+    it('lifts a function into the context of applicative functors', function() {
```
davidchambers: Let's remove references to applicative functors. It's confusing to mix Fantasy Land and Haskell terminology.

index.js
```diff
@@ -419,6 +419,55 @@
...
+  //. S.Just(3)
+  //.
+  //. > S.lift(R.inc, S.Nothing())
+  //. S.Nothing()
+  //. ```
+  S.lift = def('lift', [Function, a], R.map);
+
+  //# lift2 :: Apply f => (a -> b -> c) -> f a -> f b -> f c
+  //.
+  //. Promotes a binary function to a function which operates on two [Apply][]s.
```
davidchambers: Could you wrap this at 79 characters? It's not currently enforced by the linter, but I like my 80-character Vim buffers. :)

test/index.js
```diff
@@ -293,6 +293,84 @@ describe('function', function() {
 
   });
 
+  describe('lift', function() {
+    it('is a binary function', function() {
```
davidchambers: Nit: Please include an empty line between the two lines above.

test/index.js
```diff
@@ -293,6 +293,84 @@ describe('function', function() {
...
+      eq(typeof S.lift, 'function');
+      eq(S.lift.length, 2);
+    });
+
+    it('type checks its arguments', function() {
+      assert.throws(function() { S.lift('wrong'); },
+                    errorEq(TypeError,
+                            '‘lift’ requires a value of type Function as its ' +
+                            'first argument; received "wrong"'));
```
davidchambers: Nit: Let's wrap `... type Function ' +` to avoid exceeding 79 characters.

test/index.js
```diff
@@ -293,6 +293,87 @@ describe('function', function() {
...
+    });
+
+    it('lifts a function into the context of Applys', function() {
+      eq(S.lift2(R.add, S.Just(3), S.Just(3)), S.Just(6));
+      eq(S.lift2(R.add, S.Nothing(), S.Just(3)), S.Nothing());
+
+      eq(S.lift2(R.add, S.Right(3), S.Left(4)), S.Left(4));
+      eq(S.lift2(R.add, S.Right(3), S.Right(4)), S.Right(7));
+
+      eq(S.lift2(R.add, [1, 2], [1, 2]), [2, 3, 3, 4]);
```
davidchambers: It would be nice to use `[1, 2]` and `[10, 20]` here so the return value makes the order of permutations clear.

davidchambers:
:deciduous_tree:

svozza:
:tada:

PR #123, opened 2015-12-10
integrate sanctuary-def

davidchambers:
Depends on plaid/sanctuary-def#4

Type class constraints obviate the need for ad hoc type checking via `assertMethodExists`, as demonstrated by the changes to `Maybe#concat` below.

Before:

```javascript
//  Nothing#concat :: Maybe a ~> Maybe a -> Maybe a
Nothing.prototype.concat = def('Nothing#concat', [Maybe], function(maybe) {
  if (maybe.isJust) {
    assertMethodExists('concat', maybe.value);
  }
  return maybe;
});

//  Just#concat :: Maybe a ~> Maybe a -> Maybe a
Just.prototype.concat = def('Just#concat', [Maybe], function(maybe) {
  assertMethodExists('concat', this.value);
  if (maybe.isJust) {
    assertMethodExists('concat', maybe.value);
    return Just(this.value.concat(maybe.value));
  } else {
    return this;
  }
});
```

After:

```javascript
Maybe.prototype.concat =
method('Maybe#concat',
       {a: [Semigroup]},
       [$Maybe(a), $Maybe(a), $Maybe(a)],
       function(mx, my) {
         return mx.isNothing ? my :
                my.isNothing ? mx : Just(mx.value.concat(my.value));
       });
```

GitHub is refusing to render the diff for __index.js__, so I've included it below (and will keep it up to date as I update this pull request).

```diff
diff --git a/index.js b/index.js
index 5105303..2c95998 100644
--- a/index.js
+++ b/index.js
@@ -59,6 +59,13 @@
 //. (for any type `a`) with an argument of type `a -> b` (for any type `b`),
 //. it returns a value of type `Maybe b`._
 //.
+//. Sanctuary supports type classes: constraints on type variables. Whereas
+//. `a -> a` implicitly supports every type, `Functor f => (a -> b) -> f a ->
+//. f b` requires that `f` be a type which satisfies the requirements of the
+//. Functor type class. Type-class constraints appear at the beginning of a
+//. type signature, separated from the rest of the signature by a fat arrow
+//. (`=>`).
+//.
 //. ### Accessible pseudotype
 //.
 //. What is the type of values which support property access? In other words,
@@ -81,19 +88,6 @@
 //. The List pseudotype represents non-Function values with numeric `length`
 //. properties greater than or equal to zero, such as `[1, 2, 3]` and `'foo'`.
 //.
-//. ### RegexFlags pseudotype
-//.
[...]

PR #131, opened 2016-01-25
ramda@0.19.x

davidchambers:
See the [0.18.0][1] and [0.19.0][2] upgrade guides.

I typically upgrade Ramda one minor version at a time, but I had trouble with one of the doctests when running 0.18.0, so I decided to skip that version. The changes are minimal, at any rate.

[1]: https://github.com/ramda/ramda/issues/1436
[2]: https://github.com/ramda/ramda/issues/1558

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -24,7 +24,8 @@
     "types"
   ],
   "dependencies": {
-    "ramda": "0.17.x"
+    "ramda": "0.19.x",
+    "sanctuary-def": "0.3.x"
   },
   "ignore": [
     "/.git/",
--- a/circle.yml
+++ b/circle.yml
@@ -11,7 +11,6 @@ test:
     - make lint
     - make test
     - nvm install 0.12.7
-    - nvm alias default 0.12.7
-    - make test
-    - nvm use 4 && make test
-    - nvm use 5 && make test
+    - nvm use 0.12.7 && rm -r node_modules && make setup test
+    - nvm use 4 && rm -r node_modules && make setup test
+    - nvm use 5 && rm -r node_modules && make setup test
--- a/index.js
+++ b/index.js
@@ -2310,7 +2310,7 @@
                           R.all(R.pipe(R.toUpper,
                                        R.indexOf(_, charset),
                                        R.gte(_, 0))))),
-          R.map(R.partialRight(parseInt, radix)),
+          R.map(R.partialRight(parseInt, [radix])),
           R.filter(Integer.test)
         )(s);
       });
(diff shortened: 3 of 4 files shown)


bower.json
```diff
@@ -24,7 +24,8 @@
     "types"
   ],
   "dependencies": {
-    "ramda": "0.17.x"
+    "ramda": "0.19.x",
+    "sanctuary-def": "0.3.x"
```
davidchambers: I forgot to add this in #123.

svozza:
I was just in the middle of doing a PR for this because I need that `toIsoString` fix in Ramda 0.19 and I was getting the same doc test error when running locally as is in CircleCI now. It's very hard to debug.

svozza:
So it looks like the issue is actually in doctest. For some reason, on my install `Ramda` is not in the `doctest` `node_modules` folder. I'm guessing it was working before because `require` was recursively finding the `Ramda` package being used in `Sanctuary` and because they we're both expecting v0.17 then that was fine. When I physically went into the `doctest` directory in `node_modules` and ran `npm install` the errors disappeared.

svozza:
If you look in the CircleCI logs you'll see this:

>npm install
npm WARN unmet dependency /home/ubuntu/sanctuary/node_modules/doctest requires ramda@'0.17.x' but will load
npm WARN unmet dependency /home/ubuntu/sanctuary/node_modules/ramda,
npm WARN unmet dependency which is version 0.19.1

circle.yml
```diff
@@ -12,6 +12,6 @@ test:
     - make test
     - nvm install 0.12.7
     - nvm alias default 0.12.7
-    - make test
-    - nvm use 4 && make test
-    - nvm use 5 && make test
+    - rm -r node_modules && make setup test
+    - nvm use 4 && rm -r node_modules && make setup test
+    - nvm use 5 && rm -r node_modules && make setup test
```
davidchambers: This did the trick!

circle.yml
```diff
@@ -12,6 +12,6 @@ test:
     - make test
     - nvm install 0.12.7
     - nvm alias default 0.12.7
```
davidchambers: @svozza, could we `nvm use 0.12.7 && ...` here instead, or is there a reason this is different from the lines below?

svozza: Unfortunately not, CircleCI doesn't have 0.12.7 installed by default for some reason. If you do `nvm use 0.12` it will run against 0.12.0.

davidchambers: Thanks for the info. Let's leave it as is, in that case.

svozza: You could do `nvm use 0.12.7 && make test` once the install is done though. There's no need for that `default alias` bit.

davidchambers: :zap:

svozza:
LGTM.

PR #136, opened 2016-01-29
function: assume that Function satisfies the requirements of Apply

davidchambers:
[`R.map`][1] and [`R.ap`][2] support Function, so we can assume that Function satisfies the requirements of [Apply][3].

This allows us to define `complement`, `both`, and `either` via lifting, as discussed in #102.

```javascript
//    complement :: Functor f => f Boolean -> f Boolean
const complement = S.lift(S.not);

//    both :: Apply f => f Boolean -> f Boolean -> f Boolean
const both = S.lift2(S.and);

//    either :: Apply f => f Boolean -> f Boolean -> f Boolean
const either = S.lift2(S.or);

//    isEven :: Integer -> Boolean
const isEven = x => x % 2 === 0;

//    isOdd :: Integer -> Boolean
const isOdd = complement(isEven);

//    isPositive :: Number -> Boolean
const isPositive = x => x > 0;

both(isEven, isPositive)(42);   // => true
both(isEven, isPositive)(43);   // => false
both(isEven, isPositive)(-42);  // => false
both(isEven, isPositive)(-43);  // => false
```

/cc @benperez

[1]: https://github.com/ramda/ramda/blob/v0.19.1/src/map.js#L44-L47
[2]: https://github.com/ramda/ramda/blob/v0.19.1/src/ap.js#L30-L33
[3]: https://github.com/fantasyland/fantasy-land#apply

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -49,6 +49,9 @@ var parseHex = function(s) {
   return n !== n ? S.Left('Invalid hexadecimal string') : S.Right(n);
 };
 
+//  positive :: Number -> Boolean
+var positive = function(n) { return n > 0; };
+
 //  rem :: (Number, Number) -> Number !
 var rem = function(x, y) {
   if (y === 0) {
@@ -381,6 +384,9 @@ describe('function', function() {
 
       eq(S.lift(R.multiply(2), [1, 2, 3]), [2, 4, 6]);
       eq(S.lift(R.multiply(2), []), []);
+
+      eq(S.lift(S.not, S.even)(42), false);
+      eq(S.lift(S.not, S.even)(43), true);
     });
 
   });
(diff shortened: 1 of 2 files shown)


test/index.js
```diff
@@ -435,6 +449,8 @@ describe('function', function() {
 
       eq(S.lift3(R.reduce, [R.add], [0], [[1, 2, 3]]), [6]);
       eq(S.lift3(R.reduce, [R.add], [0], []), []);
+
+      eq(S.lift3(R.curry(area), R.dec, S.I, R.inc)(4), 6);
```
benperez: :triangular_ruler:

benperez:
:dromedary_camel: Nice fix!

test/index.js
davidchambers: It looks correctly aligned to me. What am I missing?

benperez: I just left a triangle because we were talking about triangles earlier. I now see why that would be confusing :stuck_out_tongue_winking_eye:

davidchambers: Ah, yes! I should have understood that reference!

davidchambers:
Does this LGTY, @svozza?

svozza:
Sorry, I thought I'd responded to this already. LGTM.

Although, my `map` function is not going to be simple for much longer!

PR #137, opened 2016-01-29
export Maybe and Either types

svozza:
As discussed in svozza/sanctuary-date-poc-2/pull/1.

### DIFF
--- a/index.js
+++ b/index.js
@@ -193,7 +193,7 @@
   var r = $.TypeVariable('r');
 
   //  $Either :: Type -> Type -> Type
-  var $Either = $.BinaryType(
+  var $Either = S.EitherType = $.BinaryType(
     'sanctuary/Either',
     function(x) { return x != null && x['@@type'] === 'sanctuary/Either'; },
     function(either) { return either.isLeft ? [either.value] : []; },
@@ -215,7 +215,7 @@
   );
 
   //  $Maybe :: Type -> Type
-  var $Maybe = $.UnaryType(
+  var $Maybe = S.MaybeType = $.UnaryType(
     'sanctuary/Maybe',
     function(x) { return x != null && x['@@type'] === 'sanctuary/Maybe'; },
     function(maybe) { return maybe.isJust ? [maybe.value] : []; }
@@ -533,14 +533,20 @@
       });
 
   //. ### Maybe type
-
-  //# Maybe :: TypeRep Maybe
   //.
   //. The Maybe type represents optional values: a value of type `Maybe a` is
   //. either a Just whose value is of type `a` or a Nothing (with no value).
   //.
   //. The Maybe type satisfies the [Monoid][], [Monad][], [Traversable][],
   //. and [Extend][] specifications.
+
+  //# MaybeType :: Type -> Type
+  //.
+  //. A [`UnaryType`][UnaryType] for use with [sanctuary-def][].
+
+  //# Maybe :: TypeRep Maybe
+  //.
+  //. The [type representative](#type-representatives) for the Maybe type.
   var Maybe = S.Maybe = function Maybe() {
     if (arguments[0] !== sentinel) {
       throw new Error('Cannot instantiate Maybe');
@@ -1122,8 +1128,6 @@
       });
 
   //. ### Either type
-
-  //# Either :: TypeRep Either
   //.
   //. The Either type represents values with two possibilities: a value of type
   //. `Either a b` is either a Left whose value is of type `a` or a Right whose
@@ -1131,6 +1135,14 @@
   //.
   //. The Either type satisfies the [Semigroup][], [Monad][], and [Extend][]
   //. specifications.
+
+  //# EitherType :: Type -> Type -> Type
+  //.
+  //. A [`BinaryType`][BinaryType] for use with [sanctuary-def][].
+
+  //# Either :: TypeRep Either
+  //.
+  //. The [type representative](#type-representatives) for the Either type.
   var Either = S.Either = function Either() {
     if (arguments[0] !== sentinel) {
       throw new Error('Cannot instantiate Either');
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -543,6 +543,9 @@
   //.
   //. The Maybe type satisfies the [Monoid][], [Monad][], [Traversable][],
   //. and [Extend][] specifications.
+  //.
+  //. This type is exported as `S.MaybeType` for use with [sanctuary-def][].
+  //.
```

index.js
```diff
@@ -1133,6 +1136,8 @@
   //.
   //. The Either type satisfies the [Semigroup][], [Monad][], and [Extend][]
   //. specifications.
+  //.
+  //. This type is exported as `S.EitherType` for use with [sanctuary-def][].
```
davidchambers: This deserves a separate section in the documentation. We can then have `Either :: TypeRep Either` and `EitherType :: Type -> Type -> Type`.

svozza: Where will I put it?

davidchambers: I think the two paragraphs we currently have should actually live until the `### Either type` heading:

```javascript
  //. ### Either type
  //.
  //. The Either type represents values with two possibilities: a value of type
  //. `Either a b` is either a Left whose value is of type `a` or a Right whose
  //. value is of type `b`.
  //.
  //. The Either type satisfies the [Semigroup][], [Monad][], and [Extend][]
  //. specifications.

  //# EitherType :: Type -> Type -> Type
  //.
  //. TK (should include a link to https://github.com/plaid/sanctuary-def#binarytype)

  //# Either :: TypeRep Either
  //.
  //. TK (should include a link to https://github.com/plaid/sanctuary#type-representatives)
```

What do you think of this structure?

svozza: I'm not really sure what to write for the last bit there, it seems like I'd just be repeating the first paragraph.

```javascript
  //# Either :: TypeRep Either
  //.
  //. The [type representative](#type-representatives) for the Either type.
```

svozza: Looks good. Will I do the same for `Maybe`?

davidchambers: Yes, please. :)

index.js
```diff
@@ -2673,17 +2682,19 @@
...
+//. [R.equals]:       http://ramdajs.com/docs/#equals
+//. [R.map]:          http://ramdajs.com/docs/#map
+//. [R.type]:         http://ramdajs.com/docs/#type
+//. [Ramda]:          http://ramdajs.com/
+//. [RegExp]:         https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp
+//. [RegexFlags]:     https://github.com/plaid/sanctuary-def#regexflags
+//. [sanctuary-def]:  https://github.com/plaid/sanctuary-def
+//. [BinaryType]:     https://github.com/plaid/sanctuary-def#binarytype
+//. [Semigroup]:      https://github.com/fantasyland/fantasy-land#semigroup
+//. [parseInt]:       https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt
```
davidchambers: Let's keep these links lexicographically ordered by identifier. In Vim, I type `vip:sort u` to achieve this. :)

test/index.js
```diff
@@ -564,6 +565,11 @@ describe('maybe', function() {
                     errorEq(Error, 'Cannot instantiate Maybe'));
     });
 
+    it('has its type definition exported', function() {
+      assert(S.MaybeType($.Number).test(S.Nothing()));
+      assert(S.MaybeType($.Number).test(S.Just(1)));
```
davidchambers: I prefer to write `eq(expr, true)` rather than `assert(expr)` as it's more rigorous. Also, it'd be nice to test a value which is not a member of `Maybe Number`.

index.js

index.js
svozza: :rocket:

svozza:
:zap: x 3

index.js
```diff
@@ -531,6 +531,15 @@
   //.
   //. The Maybe type satisfies the [Monoid][], [Monad][], [Traversable][],
   //. and [Extend][] specifications.
+  //.
```

svozza: :rocket:

test/index.js
```diff
@@ -564,6 +565,12 @@ describe('maybe', function() {
                     errorEq(Error, 'Cannot instantiate Maybe'));
     });
 
+    it('has its type definition exported', function() {
+      eq(S.MaybeType($.Number).test(S.Nothing()), true);
+      eq(S.MaybeType($.Number).test(S.Just(1)), true);
+      eq(S.MaybeType($.Number).test('abc'), false);
+    });
+
```
davidchambers: This could live in its own `describe` block (before the `describe('Maybe')` block, perhaps).

svozza: Cool. I moved the `EitherType` tests as well.

davidchambers: Thanks!

svozza: Weird, how come this comment hasn't collapsed after I updated the code last night.

davidchambers: GitHub only collapses comments on a diff if the commented line is changed or if one of the (two?) lines immediately above that line is changed. This is confusing because GitHub displays more context than it takes into account when deciding whether to collapse comments. It should use the same number of lines in both cases.

index.js
```diff
@@ -183,7 +183,7 @@
   var r = $.TypeVariable('r');
 
   //  $Either :: Type -> Type -> Type
-  var $Either = $.BinaryType(
+  var $Either = S.EitherType =  $.BinaryType(
```
davidchambers: Nit: There's a stray space. :)

index.js
```diff
@@ -531,6 +531,14 @@
   //.
   //. The Maybe type satisfies the [Monoid][], [Monad][], [Traversable][],
   //. and [Extend][] specifications.
+
+  //# MaybeType :: Type -> Type
+  //.
+  //. A [`UnaryType`][UnaryType] for use with [sanctuary-def][].
+
+  //# Maybe :: TypeRep Maybe
```
davidchambers: Let's delete the existing occurrence of this heading.

test/index.js
svozza: Yeah, that would make a lot more sense.

davidchambers:
:deciduous_tree:

Issue #127, opened 2016-01-05
Add Math Functions


svozza:
It would be nice to have type-safe versions of `R.add`, `R.subtract`, `R.multiply` and `R.divide`. If people think that's a good idea, I will do a PR once #123 has landed.

davidchambers:
Absolutely! You could also define `S.inc` and `S.dec` while you're at it. :)

svozza:
Yep, good shout.

PR #138, opened 2016-01-30
Add div and mult functions

svozza:
Last of the functions from #127. I never want to see another `concat` error message again!

### DIFF
--- a/index.js
+++ b/index.js
@@ -239,6 +239,7 @@
   //  env :: [Type]
   var env = $.env.concat([
     $.FiniteNumber,
+    $.NonZeroFiniteNumber,
     $Either,
     Integer,
     List,
@@ -2276,6 +2277,35 @@
       {},
       [$.FiniteNumber, $.FiniteNumber],
       function(a) { return a - 1; });
+
+  //# mult :: FiniteNumber -> FiniteNumber -> FiniteNumber
+  //.
+  //. Returns the product of two (finite) numbers.
+  //.
+  //. ```javascript
+  //. > S.mult(4, 2)
+  //. 8
+  //. ```
+  S.mult =
+  def('mult',
+      {},
+      [$.FiniteNumber, $.FiniteNumber, $.FiniteNumber],
+      function(a, b) { return a * b; });
+
+  //# div :: FiniteNumber -> NonZeroFiniteNumber -> FiniteNumber
+  //.
+  //. Returns the result of dividing its first argument (a finite number) by
+  //. its second argument (a non-zero finite number).
+  //.
+  //. ```javascript
+  //. > S.div(7, 2)
+  //. 3.5
+  //. ```
+  S.div =
+  def('div',
+      {},
+      [$.FiniteNumber, $.NonZeroFiniteNumber, $.FiniteNumber],
+      function(a, b) { return a / b; });
 
   //. ### Parse
 
--- a/test/index.js
+++ b/test/index.js
@@ -629,7 +629,8 @@ describe('maybe', function() {
       assert.throws(function() { S.Nothing().concat(S.Just(1)); },
                     errorEq(TypeError,
                             '‘Maybe#concat’ requires ‘a’ to implement ' +
-                            'Semigroup; Number and FiniteNumber and Integer ' +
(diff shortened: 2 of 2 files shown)


davidchambers:
> I never want to see another `concat` error message again!

:stuck_out_tongue_winking_eye:

index.js
```diff
@@ -2277,6 +2278,35 @@
       [$.FiniteNumber, $.FiniteNumber],
       function(a) { return a - 1; });
 
+  //# mult :: FiniteNumber -> FiniteNumber -> FiniteNumber
+  //.
+  //. Returns the product of two (finite) numbers.
+  //.
+  //. ```javascript
+  //. > S.mult(2, 2)
```
davidchambers: Let's use values with don't also *add* to give 4. ;)

index.js
```diff
@@ -2277,6 +2278,35 @@
...
+  //. 4
+  //. ```
+  S.mult =
+  def('mult',
+      {},
+      [$.FiniteNumber, $.FiniteNumber, $.FiniteNumber],
+      function(a, b) { return a * b; });
+
+  //# div :: FiniteNumber -> FiniteNumber -> FiniteNumber
```
davidchambers: The second argument's type is `NonZeroFiniteNumber`.

test/index.js
```diff
@@ -3374,6 +3382,100 @@ describe('number', function() {
...
+                            '‘mult’ expected a value of type FiniteNumber ' +
+                            'as its second argument; received Infinity'));
+
+      assert.throws(function() { S.mult(2, -Infinity); },
+                    errorEq(TypeError,
+                            '‘mult’ expected a value of type FiniteNumber ' +
+                            'as its second argument; received -Infinity'));
+    });
+
+    it('mults two numbers', function() {
```
davidchambers: `s/mults/multiplies/`

index.js
svozza: Ha. Good point.

index.js

test/index.js

index.js
```diff
@@ -2277,6 +2278,35 @@
...
+      [$.FiniteNumber, $.FiniteNumber, $.FiniteNumber],
+      function(a, b) { return a * b; });
+
+  //# div :: FiniteNumber -> NonZeroFiniteNumber -> FiniteNumber
+  //.
+  //. Returns the result of dividing its first argument (a finite number) by
+  //. its second argument (a non-zero finite number).
+  //.
+  //. ```javascript
+  //. > S.div(4, 2)
```
davidchambers: The same logic applies here, actually. Let's use arguments which result in different values for `R.div` and `R.sub`. Also, let's make it clear that this is not integer division.

svozza: Fair enough, although regarding making it clear that this isn't integer division, I'm not really sure how much clearer we can make it given the type signature at the top of the function description.

davidchambers: I was thinking something like `S.div(7, 2) = 3.5` (rather than `3` as would be the case with integer division).

index.js
```diff
@@ -2277,6 +2278,38 @@
...
+  //. Returns the result of dividing its first argument (a finite number) by
+  //. its second argument (a non-zero finite number).
+  //.
+  //. ```javascript
+  //. > S.div(8, 2)
+  //. 4
+  //.
+  //. > S.div(7, 2)
+  //. 3.5
+  //. ```
```
davidchambers: The information content of the second example is a strict superset of the information content of the first example, so we could remove the first example.

index.js
```diff
@@ -2277,6 +2278,36 @@
...
+      [$.FiniteNumber, $.FiniteNumber, $.FiniteNumber],
+      function(a, b) { return a * b; });
+
+  //# div :: FiniteNumber -> NonZeroFiniteNumber -> FiniteNumber
+  //.
+  //. Returns the result of dividing its first argument (a finite number) by
+  //. its second argument (a non-zero finite number).
+  //.
+  //. ```javascript
+  //.
```

svozza: :rocket:

davidchambers:
:deciduous_tree:

PR #140, opened 2016-01-31
integer: add S.even and S.odd

davidchambers:
This pull request introduces two straightforward functions:

```haskell
even :: Integer -> Boolean
odd  :: Integer -> Boolean
```

It also removes the internal `Integer` definition, which is unnecessary now that we have [`$.Integer`][1]. :)

Note: In many JavaScript libraries one would expect these functions to be named `isEven` and `isOdd`. In Haskell, though, they are simply `even` and `odd`. This feels right in Sanctuary since we also emphasize reasoning based on types rather than names.

It's wonderful how simple a function's implementation becomes when one is able to restrict its domain. We needn't decide what to do with inputs such as `0.5` and `Infinity`, in this case.

[1]: https://github.com/plaid/sanctuary-def#integer

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -3514,6 +3514,88 @@ describe('number', function() {
 
 });
 
+describe('integer', function() {
+
+  describe('even', function() {
+
+    it('is a unary function', function() {
+      eq(typeof S.even, 'function');
+      eq(S.even.length, 1);
+    });
+
+    it('type checks its arguments', function() {
+      assert.throws(function() { S.even(0.5); },
+                    errorEq(TypeError,
+                            '‘even’ expected a value of type Integer ' +
+                            'as its first argument; received 0.5'));
+
+      assert.throws(function() { S.even(Infinity); },
+                    errorEq(TypeError,
+                            '‘even’ expected a value of type Integer ' +
+                            'as its first argument; received Infinity'));
+    });
+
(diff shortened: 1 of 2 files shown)


test/index.js
```diff
@@ -3478,6 +3478,88 @@ describe('number', function() {
...
+      eq(S.odd.length, 1);
+    });
+
+    it('type checks its arguments', function() {
+      assert.throws(function() { S.odd(0.5); },
+                    errorEq(TypeError,
+                            '‘odd’ expected a value of type Integer ' +
+                            'as its first argument; received 0.5'));
+
+      assert.throws(function() { S.odd(Infinity); },
```
svozza: Maybe use `-Infinity` here to have some symmetry  with `even`?

svozza:
>It also removes the internal Integer definition, which is unnecessary now that we have $.Integer

:+1: Nice, I was going to submit a PR for this.

test/index.js
davidchambers: I don't follow. Could you elaborate?

svozza: I meant use `Infinity` in `even` and use `-Infinity` in `odd`. :)

davidchambers: :zap:

svozza:
LGTM.

PR #141, opened 2016-01-31
number: add S.negate

davidchambers:
```haskell
negate :: ValidNumber -> ValidNumber
```

[`R.negate`][1], but with type checking. :)

[1]: http://ramdajs.com/docs/#negate

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -631,7 +631,7 @@ describe('maybe', function() {
                             '‘Maybe#concat’ requires ‘a’ to implement ' +
                             'Semigroup; Number and FiniteNumber ' +
                             'and NonZeroFiniteNumber and Integer ' +
-                            'do not'));
+                            'and ValidNumber do not'));
     });
 
     it('provides an "equals" method', function() {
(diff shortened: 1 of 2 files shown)


test/index.js
```diff
@@ -847,19 +847,21 @@ describe('maybe', function() {
...
-                            'NonZeroFiniteNumber and Integer do not'));
+                            'NonZeroFiniteNumber and Integer and ' +
+                            'ValidNumber do not'));
 
       assert.throws(function() { S.Just([1, 2, 3]).concat(S.Just(3)); },
                     errorEq(TypeError,
                             '‘Maybe#concat’ requires ‘a’ to implement ' +
                             'Semigroup; Number and FiniteNumber and ' +
-                            'NonZeroFiniteNumber and Integer do not'));
+                            'NonZeroFiniteNumber and Integer and ' +
```
svozza: We have to find a better way of doing this...

svozza:
LGTM. Should we compile a list of Ramda functions we want to make type safe and create an issue so we can start ticking them off as they're implemented?

davidchambers:
> Should we compile a list of Ramda functions we want to make type safe and create an issue so we can start ticking them off as they're implemented?

Great idea!

PR #142, opened 2016-01-31
Add S.map function

svozza:
This comment in #140 struck a chord with me:

>It's wonderful how simple a function's implementation becomes when one is able to restrict its domain. 

When I look at the implementation of `map` in Ramda, what strikes me is how complicated it is. I think we should take advantage of the type checking we have and simplify wherever possible. In fact, we could simplify this further by not special casing arrays at the cost of losing IE8 compatibility.

One side effect of this PR is that it means `Sanctuary` now has an alias, i.e., `S.lift`; I'm not sure how we resolve this other than deprecating the unary version of `lift`.

Another benefit I can see here would be for performance should a user decide to run with type-checking disabled in production.

This version is more restrictive in that it doesn't operate on Objects as our definition of `Functor` does not include them. I actually prefer this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,7 +25,7 @@
 //. Sanctuary gives us a fighting chance of avoiding such errors. We might
 //. write:
 //.
-//.     R.map(S.toUpper, S.head(words))
+//.     S.map(S.toUpper, S.head(words))
 //.
 //. ## Types
 //.
@@ -36,9 +36,9 @@
 //. That is, it takes an argument of type `Number` and returns a value of
 //. type `Number`.
 //.
-//. [`R.map`][R.map] has type `(a -> b) -> [a] -> [b]`. That is, it takes
+//. [`S.map`](#map) has type `(a -> b) -> [a] -> [b]`. That is, it takes
 //. an argument of type `a -> b` and returns a value of type `[a] -> [b]`.
-//. `a` and `b` are type variables: applying `R.map` to a value of type
+//. `a` and `b` are type variables: applying `S.map` to a value of type
 //. `String -> Number` will give a value of type `[String] -> [Number]`.
 //.
 //. Sanctuary embraces types. JavaScript doesn't support algebraic data types,
@@ -233,7 +233,8 @@
   var Functor = $.TypeClass(
     'sanctuary/Functor',
     function(x) {
-      return R.contains(_type(x), ['Array', 'Function']) ||
+      return R.contains(_type(x), ['Array', 'Object', 'Function']) ||
+             // $.test(env, $.StrMap, x) ||
              hasMethod('map')(x);
     }
   );
@@ -315,6 +316,7 @@
     $.Integer,
     $Maybe,
     $.RegexFlags,
+    // $.StrMap,
     TypeRep,
     $.ValidDate,
     $.ValidNumber
@@ -442,7 +444,7 @@
   //. > S.K('foo', 'bar')
   //. 'foo'
   //.
-  //. > R.map(S.K(42), R.range(0, 5))
+  //. > S.map(S.K(42), R.range(0, 5))
   //. [42, 42, 42, 42, 42]
   //. ```
   S.K =
@@ -540,7 +542,7 @@
   //. See also [`C`](#C).
   //.
   //. ```javascript
-  //. > R.map(S.flip(Math.pow)(2), [1, 2, 3, 4, 5])
+  //. > S.map(S.flip(Math.pow)(2), [1, 2, 3, 4, 5])
   //. [1, 4, 9, 16, 25]
   //. ```
   S.flip =
@@ -548,6 +550,52 @@
       {},
       [$.Function, b, a, c],
       function(f, x, y) { return f(y, x); });
+
+  //# map :: Functor f => (a -> b) -> f a -> f b
+  //.
+  //. Takes a function and a [Functor][], applies the function to each of the
+  //. Functor's values, and returns a Functor of the same shape.
+  //.
+  //. ```javascript
+  //. > S.map(S.inc, [1, 2, 3])
+  //. [2, 3, 4]
+  //.
+  //. > S.map(S.inc, {a: 1, b: 2, c: 3})
+  //. {a: 2, b: 3, c: 4}
+  //.
+  //. > S.map(S.inc, S.Just(2))
+  //. S.Just(3)
+  //.
+  //. > S.map(S.not, S.odd)(2)
+  //. true
+  //. ```
+  var map = S.map =
+  def('map',
+      {a: [Functor], b: [Functor]},
+      [$.Function, a, b],
+      function(f, functor) {
+        var type = _type(functor);
+        var idx;
+        if (type === 'Array') {
+          var xs = [];
+          for (idx = 0; idx < functor.length; idx += 1) {
+            xs.push(f(functor[idx]));
+          }
+          return xs;
+        } else if (type === 'Function') {
+          return compose(f, functor);
+        } else if (hasMethod('map')(functor)) {
+          return functor.map(f);
+        } else {
+          var obj = {};
+          for (var key in functor) {
+            if (Object.prototype.hasOwnProperty.call(functor, key)) {
+              obj[key] = f(functor[key]);
+            }
+          }
+          return obj;
+        }
+      });
 
   //# lift :: Functor f => (a -> b) -> f a -> f b
   //.
@@ -564,7 +612,7 @@
   def('lift',
       {a: [Functor], b: [Functor]},
       [$.Function, a, b],
-      R.map);
+      map);
 
   //# lift2 :: Apply f => (a -> b -> c) -> f a -> f b -> f c
   //.
@@ -588,7 +636,7 @@
   def('lift2',
       {a: [Apply], b: [Apply], c: [Apply]},
       [$.Function, a, b, c],
-      function(f, x, y) { return R.ap(R.map(f, x), y); });
+      function(f, x, y) { return R.ap(map(f, x), y); });
 
   //# lift3 :: Apply f => (a -> b -> c -> d) -> f a -> f b -> f c -> f d
   //.
@@ -606,7 +654,7 @@
   def('lift3',
       {a: [Apply], b: [Apply], c: [Apply], d: [Apply]},
       [$.Function, a, b, c, d],
-      function(f, x, y, z) { return R.ap(R.ap(R.map(f, x), y), z); });
+      function(f, x, y, z) { return R.ap(R.ap(map(f, x), y), z); });
 
   //. ### Composition
 
@@ -683,7 +731,7 @@
       {},
       [$.Array($.Function), $.Function],
       function(fs) {
-        var n = 1 + sum(R.map(R.length, fs)) - fs.length;
+        var n = 1 + sum(map(R.length, fs)) - fs.length;
         return R.curryN(n, function() {
           var args = Array.prototype.slice.call(arguments);
           for (var idx = 0; idx < fs.length; idx += 1) {
@@ -942,13 +990,16 @@
   //.
   //. Takes a function and returns `this` if `this` is a Nothing; otherwise
   //. it returns a Just whose value is the result of applying the function to
-  //. this Just's value.
+  //. this Just's value. This method is compatible with [`map`](#map).
   //.
   //. ```javascript
   //. > S.Nothing().map(S.inc)
   //. Nothing()
   //.
   //. > S.Just([1, 2, 3]).map(S.sum)
+  //. Just(6)
+  //.
+  //. > S.map(R.sum, S.Just([1, 2, 3]))
   //. Just(6)
   //. ```
   Maybe.prototype.map =
@@ -1017,7 +1068,7 @@
          {a: [Apply], b: [Apply]},
          [$Maybe(a), $.Function, b],
          function(maybe, of) {
-           return maybe.isJust ? R.map(Just, maybe.value) : of(maybe);
+           return maybe.isJust ? map(Just, maybe.value) : of(maybe);
          });
 
   //# Maybe#toBoolean :: Maybe a ~> Boolean
@@ -1253,7 +1304,7 @@
   def('mapMaybe',
       {},
       [$.Function, $.Array(a), $.Array(b)],
-      function(f, xs) { return justs(R.map(f, xs)); });
+      function(f, xs) { return justs(map(f, xs)); });
 
   //# encase :: (a -> b) -> a -> Maybe b
   //.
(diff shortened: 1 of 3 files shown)


davidchambers:
> In fact, we could simplify this further by not special casing arrays at the cost of losing IE8 compatibility.

I don't think so. `Array.prototype.map` does not comply with the [Functor][1] specification, since it applies the provided function to three arguments rather than one:

```javascript
[1, 2, 3, 4, 5].map(S.inc);
// ! TypeError: ‘inc’ requires one argument; received three arguments
```

> This version is more restrictive in that it doesn't operate on Objects as our definition of `Functor` does not include them. I actually prefer this.

Why do you prefer this?


[1]: https://github.com/fantasyland/fantasy-land#functor

index.js
```diff
@@ -391,6 +391,37 @@
       [$.Function, b, a, c],
       function(f, x, y) { return f(y, x); });
 
+  //# map :: Functor f => (a -> b) -> f a -> f b
+  //.
+  //. Takes a function and a [Functor][], applies the function to each of the
+  //. functor's values, and returns a functor of the same shape.
+  //.
+  //. ```javascript
+  //. > S.map(S.mult(4), S.Just(2))
```
davidchambers: I'd prefer to use `S.inc` here rather than `S.mult(4)`, to reduce the cognitive load slightly.

index.js
```diff
@@ -391,6 +391,37 @@
...
+  //. S.Just(8)
+  //.
+  //. > S.map(S.mult(4), S.Nothing())
+  //. S.Nothing()
+  //.
+  //. > S.map(S.mult(2), [1, 2, 3])
+  //. [2, 4, 6]
+  //. ```
+  S.map =
```
davidchambers: Let's write `var map = S.map =` and use `map` (rather than `S.map`) elsewhere in the file.

svozza:
>I don't think so. Array.prototype.map does not comply with the Functor specification, since it applies the provided function to three arguments rather than one.

Ah yes, I was wondering why I had some doubts in my mind when I wrote that sentence.

>Why do you prefer this?

It's hard to articulate but the `R.map` change for objects always made me uneasy. Maybe it because of the ambiguous role that objects play in JavaScript as both maps and Records, which I know was done to death in the original Ramda issue but I was never fully convinced. I think it's telling that we didn't consider objects in our initial definition of the `Functor` typeclass and that no one picked up on it in the code review.

index.js

index.js

index.js
```diff
@@ -36,9 +36,9 @@
 //. That is, it takes an argument of type `Number` and returns a value of
 //. type `Number`.
 //.
-//. [`R.map`][R.map] has type `(a -> b) -> [a] -> [b]`. That is, it takes
+//. `S.map` has type `(a -> b) -> [a] -> [b]`. That is, it takes
```
davidchambers: It'd be nice to use ``[`S.map`](#map)`` here.

index.js
```diff
@@ -391,6 +391,37 @@
...
+  //.
+  //. ```javascript
+  //. > S.map(S.inc, S.Just(2))
+  //. S.Just(3)
+  //.
+  //. > S.map(S.inc, S.Nothing())
+  //. S.Nothing()
+  //.
+  //. > S.map(S.inc, [1, 2, 3])
+  //. [2, 3, 4]
```
davidchambers: Let's make this the first example.

index.js
```diff
@@ -391,6 +391,37 @@
       [$.Function, b, a, c],
       function(f, x, y) { return f(y, x); });
 
+  //# map :: Functor f => (a -> b) -> f a -> f b
+  //.
+  //. Takes a function and a [Functor][], applies the function to each of the
+  //. functor's values, and returns a functor of the same shape.
```
davidchambers: Let's `s/functor/Functor/g` to be consistent.

test/index.js
```diff
@@ -358,6 +358,38 @@ describe('function', function() {
...
+      eq(S.map.length, 2);
+    });
+
+    it('type checks its arguments', function() {
+      assert.throws(function() { S.map('xxx'); },
+                    errorEq(TypeError,
+                            '‘map’ expected a value of type Function ' +
+                            'as its first argument; received "xxx"'));
+
+      assert.throws(function() { S.map(S.mult(4), 'xxx'); },
```
davidchambers: Let's use `S.toUpper` in place of `S.mult(4)`.

index.js
```diff
@@ -25,7 +25,7 @@
 //. Sanctuary gives us a fighting chance of avoiding such errors. We might
 //. write:
 //.
-//.     R.map(R.toUpper, S.head(words))
+//.     S.map(R.toUpper, S.head(words))
```
davidchambers: It'd be nice to `s/R[.]toUpper/S.toUpper/`. :)

svozza:
:zap: x 5

davidchambers:
> It's hard to articulate but the `R.map` change for objects always made me uneasy. Maybe it because of the ambiguous role that objects play in JavaScript as both maps and Records, which I know was done to death in the original Ramda issue but I was never fully convinced.

I hope plaid/sanctuary-def#29 will put you at ease. :)

svozza:
Haha. Very good.

davidchambers:
Could you add a Function implementation now that #136 has been merged?

svozza:
I was just waiting on it to drop before I started. Quick question, do we want to the Function implementation to support functions of any arity?

davidchambers:
> [D]o we want to the Function implementation to support functions of any arity?

I'm happy to require a unary function (functions curried Ramda-style can be considered unary).

svozza:
Agreed. I think it makes the implementation much more elegant.

index.js
```diff
@@ -378,6 +378,45 @@
...
+          case 'Array':
+            var xs = [];
+            for (var idx = 0; idx < functor.length; idx += 1) {
+              xs.push(f(functor[idx]));
+            }
+            return xs;
+          case 'Function':
+            return function(x) {
+              return f(functor(x));
+            };
```
davidchambers: Could we write this as `compose(f, functor)`?

svozza: No probs. Points free ftw!

index.js
```diff
@@ -396,6 +396,55 @@
...
+          }
+          return xs;
+        } else if (type === 'Function') {
+          return compose(f, functor);
+        } else if (hasMethod('map')(functor)) {
+          return functor.map(f);
+        } else {
+          var obj = {};
+          for (var key in functor) {
+            /* istanbul ignore else */
```
davidchambers: We could test this by doing something along these lines:

```javascript
function Point(x, y) {
  this.x = x;
  this.y = y;
}
Point.prototype.color = 'black';

var point = new Point(0, 0);

S.map(S.inc, point);
// => {x: 1, y: 1}
```

davidchambers:
@Bpless has told me on at least one occasion that the way in which the various methods are documented is confusing because we do not explain why they exist as methods rather than as functions. Would using `S.map` in the [`Maybe#map`][1] and [`Either#map`][2] examples make this clearer? We could add a sentence to each method's description stating that the method is compatible with `S.map`.


[1]: https://github.com/plaid/sanctuary#Maybe.prototype.map
[2]: https://github.com/plaid/sanctuary#Either.prototype.map

svozza:
Yeah, I guess it wouldn't do any harm to be explicit. It might be obvious to us that any functor is compatible with `S.map` but that's not true for everyone.

svozza:
I've add examples for `S.map` to the above mentioned functions.

index.js
```diff
@@ -521,6 +521,54 @@
...
+  //. {a: 2, b: 3, c: 4}
+  //.
+  //. > S.map(S.inc, S.Just(2))
+  //. S.Just(3)
+  //.
+  //. > S.map(S.inc, S.Nothing())
+  //. S.Nothing()
+  //.
+  //. > S.map(S.not, S.odd)(2)
+  //. true
```
davidchambers: This is new to me. I didn't realize one could use `map` to compose functions!

index.js
```diff
@@ -914,7 +962,7 @@
   //.
   //. Takes a function and returns `this` if `this` is a Nothing; otherwise
   //. it returns a Just whose value is the result of applying the function to
-  //. this Just's value.
+  //. this Just's value. This method is compatible with [`S.map`](#map).
```
davidchambers: This is a very nice addition. :)

Elsewhere in the docs we don't use the `S.` prefix when referring to other Sanctuary functions. Let's drop it here.

index.js
```diff
@@ -36,9 +36,9 @@
 //. That is, it takes an argument of type `Number` and returns a value of
 //. type `Number`.
 //.
-//. [`R.map`][R.map] has type `(a -> b) -> [a] -> [b]`. That is, it takes
+//. [`S.map`](#map) has type `(a -> b) -> [a] -> [b]`. That is, it takes
```
davidchambers: This is the only reference to the `R.map` link identifier, so we can delete the relevant line from the bottom of the file.

index.js
```diff
@@ -521,6 +521,54 @@
...
+          }
+          return xs;
+        } else if (type === 'Function') {
+          return compose(f, functor);
+        } else if (hasMethod('map')(functor)) {
+          return functor.map(f);
+        } else {
+          var obj = {};
+          for (var key in functor) {
+            if (Object.prototype.hasOwnProperty.call(functor, key)) {
```
davidchambers: Let's use `R.keys` here (or define our own `keys` function).

svozza: `S.keys` is next on my hitlist once https://github.com/plaid/sanctuary-def/pull/33 lands.

index.js

index.js
svozza: Yeah, it's pretty neat. I wonder should I remove the `S.nothing` example there, there's probably no need for two `Maybe` examples.

davidchambers: > I wonder should I remove the `S.nothing` example there, there's probably no need for two `Maybe` examples.

Sounds good to me.

davidchambers:
Please rebase this branch, Stefano, and utilize [`$.StrMap`][1].


[1]: https://github.com/plaid/sanctuary-def#strmap

svozza:
:+1: Btw, can I change the treatment of objects back to a straight `for` loop without R.keys etc? I think the implementation of a function as fundamental as `map` should be as low-level as possible.

davidchambers:
A low-level implementation sounds good to me. I'd like to remove the Ramda dependency eventually. :)

svozza:
Hmmm, something odd happens when I add `StrMap` to the environment, the StrMap type get repeat for some reason:

```
TypeError: Invalid value

slice :: Integer -> Integer -> List a -> Maybe (List a)
                               ^^^^^^
                                 1

1)  {"length": -1} :: Object, StrMap Number, StrMap FiniteNumber, StrMap NonZeroFiniteNumber, StrMap Integer, StrMap ValidNumber, StrMap Number, StrMap FiniteNumber, StrMap NonZeroFiniteNumber, StrMap Integer, StrMap ValidNumber

The value at position 1 is not a member of ‘List a’.
```

svozza:
I have no idea how those other commit have made it in there either.

davidchambers:
```haskell
{"length": -1} :: Object, StrMap Number, StrMap FiniteNumber, StrMap NonZeroFiniteNumber, StrMap Integer, StrMap ValidNumber, StrMap Number, StrMap FiniteNumber, StrMap NonZeroFiniteNumber, StrMap Integer, StrMap ValidNumber
```

This is accurate. It's overly wordy, though. As a result of plaid/sanctuary-def#52 we no longer need to include all these numeric types in our environment. [`$.Number`][1] is a superset of all the other numeric types, so just having that in the environment is sufficient for polymorphic functions.


[1]: https://github.com/plaid/sanctuary-def#number

Avaq:
Hi, I'm not sure if this would be welcomed at this stage, so I wouldn't be upset if you disregard it, but I'm just going to throw my idea out here. I personally think Ramda's `map` is *overly* polymorphic. You can input just about anything, and `map` will do something to it. This leads to some surprising output at the end of a pipeline if the input was simply not the type you expected it to be.

We might be making the same mistake here. The definition of `Functor` is becoming super loose and therefore `map`s type-checking, and the type-checking of other functions using `Functor` is a lot less powerful. It might even come to the point that you create a function with a type-constraint on `Functor`, and then must do *additional* type checking inside the function body to ensure you've got the *right* functor.

I'd much rather see Sanctuary treat `Functor` strictly like an `Object` implementing `map`, and export different versions of map, something like:

```haskell
lmap :: (a -> b) -> [a] -> [b]
omap :: (a -> b) -> {k: a} -> {k: b}
fmap :: Functor m => (a -> b) -> m a -> m b
tmap :: Step f => (a -> b) -> f a -> f b
map :: union(Array, Object, Functor, Step) m => (a -> b) -> m a -> m b
```

I know I would use the specialized versions every time, just to get the benefits of type-checking. The normal `map` would just "pattern match" against the union, and dispatch to the right specialized map.

svozza:
> $.Number is a superset of all the other numeric types, so just having that in the environment is sufficient for polymorphic functions.

So I should remove the other types from the environment and then replace all uses of these number types with `$.Number`?

davidchambers:
> So I should remove the other types from the environment and then replace all uses of these number types with `$.Number`?

In my earlier comment I didn't do a good job of explaining the relationship between the environment and polymorphic functions. I'll elaborate.

For a function such as [`take`][1], we *do* want to use a more restrictive type—[`$.Integer`][2]—in place of [`$.Number`][3].

Let's consider a polymorphic function, `id :: a -> a`, defined in the empty environment. How does `id(42)` evaluate? Without the logic added in plaid/sanctuary-def#38, this would be a type error, since `42` is not a member of any type in the environment.

Now, consider `id :: a -> a` defined in the environment comprising `$.Number`. `42 :: Number`, so `id(42)` is not a type error.

Finally, consider `id :: a -> a` defined in the environment comprising `$.Number` and `$.Integer`. `42 :: Number, Integer`, so again, `id(42)` is not a type error.

Since every value which is a member of `$.Integer` is also a member of `$.Number`, including `$.Integer` in an environment does not prevent any type errors. All it does is affect error messages.

If we decide

```haskell
{"length": -1} :: Object, StrMap Number, StrMap FiniteNumber, StrMap NonZeroFiniteNumber, StrMap Integer, StrMap ValidNumber, StrMap Number, StrMap FiniteNumber, StrMap NonZeroFiniteNumber, StrMap Integer, StrMap ValidNumber
```

is overly wordy, we can use

```haskell
{"length": -1} :: Object, StrMap Number
```

instead simply by removing the other numeric types from the environment.


[1]: http://sanctuary.js.org/#take
[2]: https://github.com/plaid/sanctuary-def#integer
[3]: https://github.com/plaid/sanctuary-def#number

svozza:
Ah I get you. I think I'll leave it as it is for the moment. It feels like a separate pull request to deal with these log messages.

svozza:
Should there be an `$.StrMap.test` function? I'm getting an error when I try to invoke saying it doesn't exist.

davidchambers:
> Should there be an `$.StrMap.test` function?

[`$.StrMap`][1], like [`$.Array`][2], is a type *constructor*. It is not itself a Type since it is of type `Type -> Type`.

I imagine `$.StrMap(a).test(x)` is what you're after, but I just discovered a problem with this which I reported in plaid/sanctuary-def#56.


[1]: https://github.com/plaid/sanctuary-def#strmap
[2]: https://github.com/plaid/sanctuary-def#array

svozza:
Ah, I see. I am destined never to land this PR!

davidchambers:
> I am destined never to land this PR!

Another way to view it is that this pull request is driving sanctuary-def forward. :)

svozza:
I much prefer that interpretation!

svozza:
Oops, I accidently pushed some changes I was working on while I'm waiting for `$.test` to drop in the new version of `sanctuary-def`.

davidchambers:
Thank you for all your work on this, Stefano. Let's move the discussion to #216. :)

PR #144, opened 2016-02-03
either: add S.lefts and S.rights

davidchambers:
This pull request adds Haskell's [`lefts`][1] and [`rights`][2]:

```haskell
lefts  :: [Either a b] -> [a]
rights :: [Either a b] -> [b]
```

I recently reviewed some code @Bpless wrote which could be nicely expressed in terms of `rights`.

These functions are similar to [`S.catMaybes`][3], which should arguably be renamed `S.justs` for consistency.

[1]: http://hackage.haskell.org/package/base-4.8.2.0/docs/Data-Either.html#v:lefts
[2]: http://hackage.haskell.org/package/base-4.8.2.0/docs/Data-Either.html#v:rights
[3]: https://github.com/plaid/sanctuary#catMaybes

### DIFF
--- a/index.js
+++ b/index.js
@@ -1490,6 +1490,42 @@
         return either.isLeft ? l(either.value) : r(either.value);
       });
 
+  //# lefts :: [Either a b] -> [a]
+  //.
+  //. Takes a list of Eithers and returns a list containing each Left's value.
+  //.
+  //. See also [`rights`](#rights).
+  //.
+  //. ```javascript
+  //. > S.lefts([S.Right(20), S.Left('foo'), S.Right(10), S.Left('bar')])
+  //. ['foo', 'bar']
+  //. ```
+  S.lefts =
+  def('lefts',
+      {},
+      [$.Array($Either(a, b)), $.Array(a)],
+      R.chain(function(either) {
+        return either.isLeft ? [either.value] : [];
+      }));
+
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1490,6 +1490,38 @@
...
+  //.
+  //. ```javascript
+  //. > S.lefts([S.Right(20), S.Left('foo'), S.Right(10), S.Left('bar')])
+  //. ['foo', 'bar']
+  //. ```
+  S.lefts =
+  def('lefts',
+      {},
+      [$.Array($Either(a, b)), $.Array(a)],
+      R.chain(function(either) {
```
Bpless: In both of these cases, we could alternatively filter -> map to achieve the same ends.

davidchambers: We certainly could!

```diff
   def('lefts',
       {},
       [$.Array($Either(a, b)), $.Array(a)],
-      R.chain(function(either) {
-        return either.isLeft ? [either.value] : [];
-      }));
+      compose(R.pluck('value'), R.filter(isLeft)));
 
   //# rights :: [Either a b] -> [b]
   //.
```

I *slightly* prefer the `chain` version because it avoids an intermediate collection. :)

svozza: I prefer `chain` too.

svozza:
These a nice functions, I could see them being quite useful.

Issue #145, opened 2016-02-03
Type Safe Ramda Functions


svozza:
As mentioned in #141, let's start try to get a list of typesafe functions we want from Ramda. I've put together a list of way too many functions that we can whittle down here as people chip in with their views.

- [x] [`R.append`](http://ramdajs.com/docs/#append)
- [x] [`R.concat`](http://ramdajs.com/docs/#concat)
- [ ] [`R.contains`](http://ramdajs.com/docs/#contains)
- [ ] [`R.equals`](http://ramdajs.com/docs/#equals)
- [ ] [`R.filter`](http://ramdajs.com/docs/#filter)
- [ ] [`R.fromPairs`](http://ramdajs.com/docs/#fromPairs)
- [x] [`R.keys`](http://ramdajs.com/docs/#keys)
- [ ] <del>[`R.length`](http://ramdajs.com/docs/#length)</del>
- [x] [`R.max`](http://ramdajs.com/docs/#max)
- [x] [`R.mean`](http://ramdajs.com/docs/#mean)
- [ ] [`R.median`](http://ramdajs.com/docs/#median)
- [ ] <del>[`R.merge`](http://ramdajs.com/docs/#merge)</del>
- [ ] [`R.mergeAll`](http://ramdajs.com/docs/#mergeAll)
- [x] [`R.min`](http://ramdajs.com/docs/#min)
- [ ] <del>[`R.modulo`](http://ramdajs.com/docs/#modulo)</del>
- [ ] [`R.omit`](http://ramdajs.com/docs/#omit)
- [x] [`R.prepend`](http://ramdajs.com/docs/#prepend)
- [x] [`R.product`](http://ramdajs.com/docs/#product)
- [x] [`R.range`](http://ramdajs.com/docs/#range)
- [x] [`R.reduce`](http://ramdajs.com/docs/#reduce)
- [ ] [`R.reject`](http://ramdajs.com/docs/#reject)
- [x] [`R.reverse`](http://ramdajs.com/docs/#reverse)
- [ ] [`R.scan`](http://ramdajs.com/docs/#scan)
- [ ] [`R.split`](http://ramdajs.com/docs/#split)
- [x] [`R.sum`](http://ramdajs.com/docs/#sum)
- [x] [`R.toPairs`](http://ramdajs.com/docs/#toPairs)
- [x] [`R.trim`](http://ramdajs.com/docs/#trim)
- [x] [`R.values`](http://ramdajs.com/docs/#values)
[...]

PR #146, opened 2016-02-04
number, string: add S.min and S.max

davidchambers:
```haskell
min :: Ord a => a -> a -> a
max :: Ord a => a -> a -> a
```

[`R.min`][1] and [`R.max`][2], but with type checking. I want to tick some boxes in #145. :)

[1]: http://ramdajs.com/docs/#min
[2]: http://ramdajs.com/docs/#max

### DIFF
--- a/index.js
+++ b/index.js
@@ -169,6 +169,16 @@
     }
   );
 
+  //  Ord :: TypeClass
+  var Ord = $.TypeClass(
+    'sanctuary/Ord',
+    function(x) {
+      var type = _type(x);
+      return type === 'Number' && $.ValidNumber.test(x) ||
+             type === 'String';
+    }
+  );
+
   //  Semigroup :: TypeClass
   var Semigroup = $.TypeClass(
     'sanctuary/Semigroup',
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -169,6 +169,16 @@
...
+  //  Ord :: TypeClass
+  var Ord = $.TypeClass(
+    'sanctuary/Ord',
+    function(x) {
+      var type = _type(x);
+      return type === 'Number' && $.ValidNumber.test(x) ||
+             type === 'String';
+    }
+  );
+
```
davidchambers: We should document all the type classes at some point.

test/index.js
```diff
@@ -3575,6 +3575,80 @@ describe('number', function() {
...
+
+    it('may be applied to (valid) numbers', function() {
+      eq(S.min(42, 99), 42);
+      eq(S.min(0.1, 0.01), 0.01);
+      eq(S.min(Infinity, -Infinity), -Infinity);
+    });
+
+    it('may be applied to strings', function() {
+      eq(S.min('abc', 'xyz'), 'abc');
+      eq(S.min('xyz', 'abc'), 'abc');
```
svozza: Maybe add a case sensitive test?

index.js
svozza: Agreed. This will also be a very useful type.

test/index.js
davidchambers: :zap:

davidchambers:
Prompted by your comment, @svozza, I added several more assertions. I also updated the examples to highlight the difference between number comparison and string comparison, and added a paragraph to each function's description to explain the distinction.

svozza:
LGTM.

PR #147, opened 2016-02-04
Replace Instances of Ramda Functions in Docs With Sanctuary Equivalent

svozza:
There were a fair few instances of `R.inc` and friends in the docs...


davidchambers:
:deciduous_tree: Very nice!

PR #148, opened 2016-02-05
Add S.reduce function

svozza:
Those of you that know your Ramda history should appreciate the irony of this one...


index.js
```diff
@@ -522,6 +522,27 @@
...
+  //. the current value from the list, and then passing the result to the
+  //. next call.
+  //.
+  //. ```javascript
+  //. > S.foldl(S.add, 0, [1, 2, 3, 4, 5])
+  //. 15
+  S.foldl =
+  def('foldl',
+      {},
+      [$.Function, a, List(b), a],
```
davidchambers: I believe we want `$.Array(b)` rather than `List(b)`: `List` exists to support functions such as [`S.head`][1] which operate on arrays or strings. We should use `$.Array` rather than `List` in the definitions of `S.pipe` and `S.meld` as well.


[1]: https://github.com/plaid/sanctuary#head

svozza: Cool. Will I change `meld` and `pipe` in this PR or do it separately?

svozza: Actually, no it should be a separate PR.

davidchambers:
Since we define [`Maybe#reduce`][1] (in accordance with the requirements of [Foldable][2]), this function should dispatch to a `reduce` method if present. It'd make sense to update the function's name to match.

We could define `Foldable :: TypeClass` to constrain the relevant argument.


[1]: https://github.com/plaid/sanctuary#Maybe.prototype.reduce
[2]: https://github.com/fantasyland/fantasy-land#foldable

svozza:
Dammit, `foldl` is so much cooler as a name. ;) Not sure what to put for the documentation now that the type takes the more genral type `Foldable` rather than `Array`. Any suggestions?

index.js
```diff
@@ -175,6 +175,12 @@
     hasMethod('concat')
   );
 
+  //  Foldable :: TypeClass
+  var Foldable = $.TypeClass(
+    'sanctuary/Foldable',
+    hasMethod('reduce')
```
davidchambers: We should return `true` for Array values explicitly. Even setting aside the fact that `Array#reduce` isn't defined in ancient environments, we should make it clear that there are two cases:

  - Array values, which we handle ourselves since `Array#reduce` is not :rainbow:-compliant; and
  - non-Array values with `reduce` methods.

Also, for lack of a better system I've been alphabetizing the type-class definitions. ;)

index.js
```diff
@@ -522,6 +528,31 @@
         });
       });
 
+  //# reduce :: Foldable b => (a -> b -> a) -> a -> b -> a
```
davidchambers: This doesn't look quite right to me. In the Array case we take `[String]`, say, as `b`, but the function provided receives a value of type `String`—rather than `[String]`—as its `b`. I think we want:

```haskell
reduce :: Foldable f => (a -> b -> a) -> a -> f b -> a
```

index.js
```diff
@@ -151,6 +151,14 @@
     }
   );
 
+  //  Foldable :: TypeClass
+  var Foldable = $.TypeClass(
+    'sanctuary/Foldable',
+    function (x) {
```
davidchambers: Nit: No space after `function`, please. :)

index.js
svozza: Yep, you're right, I knew it looked a bit off.

index.js

index.js
```diff
@@ -522,6 +530,31 @@
...
+  //.
+  //. Returns a single value by iterating through a list, successively
+  //. calling an iterator function and passing it an accumulator value and
+  //. the current value from the list, and then passing the result to the
+  //. next call.
+  //.
+  //. ```javascript
+  //. > S.reduce(S.add, 0, [1, 2, 3, 4, 5])
+  //. 15
+  S.reduce =
```
davidchambers:
```diff
   //. 15
+  //. ```
   S.reduce =
```

index.js
```diff
@@ -522,6 +530,32 @@
...
+  //.
+  //. Returns a single value by iterating through a list, successively
+  //. calling an iterator function and passing it an accumulator value and
+  //. the current value from the list, and then passing the result to the
+  //. next call.
+  //.
+  //. ```javascript
+  //. > S.reduce(S.add, 0, [1, 2, 3, 4, 5])
+  //. 15
+  //. ```
```
davidchambers: It would be nice to also include an example in which `a` and `b` differ. How about this?

```javascript
> S.reduce((xs, x) => R.concat([x], xs), [], [1, 2, 3, 4, 5])
[5, 4, 3, 2, 1]
```

svozza: Can we use the native `concat` function instead of the Ramda one?

index.js
```diff
@@ -522,6 +530,32 @@
         });
       });
 
+  //# reduce :: Foldable f => (a -> b -> a) -> a -> f b -> a
+  //.
+  //. Returns a single value by iterating through a list, successively
+  //. calling an iterator function and passing it an accumulator value and
+  //. the current value from the list, and then passing the result to the
+  //. next call.
```
davidchambers: How about something like this?

_Takes a binary function, an initial value, and a [Foldable][1], and applies the function to the initial value and the Foldable's first value, then applies the function to the result of the previous application and the Foldable's second value. Repeats this process until each of the Foldable's values has been used. Returns the initial value if the Foldable is empty; the result of the final application otherwise._


index.js
davidchambers: `[x].concat(xs)` works for me. :)

index.js
svozza: Couldn't have put it better myself. ;)

index.js

davidchambers:
This function seems more at home in the [list][1] section than in the [function][2] section. Do you agree?

Also, you should edit the commit message. ;)


[1]: https://github.com/plaid/sanctuary#list
[2]: https://github.com/plaid/sanctuary#function

svozza:
Yeah, I wasn't sure where to put it, `List` seems more appropriate given the standard way the function willl be used.

index.js
```diff
@@ -1806,6 +1814,38 @@
...
+  //.
+  //. ```javascript
+  //. > S.reduce(S.add, 0, [1, 2, 3, 4, 5])
+  //. 15
+  //.
+  //. > S.reduce((xs, x) => [x].concat(xs), [], [1, 2, 3, 4, 5])
+  //. [5, 4, 3, 2, 1]
+  //. ```
+  S.reduce =
+    def('reduce',
```
davidchambers: :arrow_left: :arrow_left:

davidchambers:
Let's insert `S.reduce` immediately before `S.unfoldr`. These two functions are related, and beginning the list section with the various slicing functions seems appropriate.

svozza:
:zap: x 2

index.js
```diff
@@ -2171,6 +2179,38 @@
...
+  //. [5, 4, 3, 2, 1]
+  //. ```
+  S.reduce =
+  def('reduce',
+      {b: [Foldable]},
+      [$.Function, a, b, a],
+      function(f, acc, foldable) {
+        if (_type(foldable) === 'Array') {
+          for (var idx = 0; idx < foldable.length; idx += 1) {
+            acc = f(acc, foldable[idx]);
```
davidchambers: I'd prefer not to reassign to a parameter. Let's rename the parameter `initial` and add `var acc = initial;` before the loop.

test/index.js
```diff
@@ -2613,6 +2613,40 @@ describe('logic', function() {
 
 describe('list', function() {
 
+  describe('reduce', function() {
```
davidchambers: Let's move this to immediately before the `describe('unfoldr')` block.

svozza:
:zap: x 2

test/index.js
```diff
@@ -3197,6 +3197,40 @@ describe('list', function() {
...
+    it('folds over lists with the supplied accumulator', function() {
+      eq(S.reduce(S.add, 0, [1, 2, 3, 4, 5]), 15);
+      eq(S.reduce(S.add, 0, []), 0);
+      eq(S.reduce(S.lift2(S.add), S.Just(0),
+                  [S.Just(1), S.Just(2), S.Just(3), S.Just(4), S.Just(5)]),
+         S.Just(15));
+    });
+
+    it('dispatches to a reduce method if present', function() {
+      eq(S.reduce(function(a, b) { return a + b; }, 10, S.Just(5)), 15);
```
davidchambers: Is there a reason to use a lambda rather than `S.add` here? Are you asserting that the function needn't be curried?

test/index.js
```diff
@@ -3197,6 +3197,40 @@ describe('list', function() {
...
+
+    it('folds over lists with the supplied accumulator', function() {
+      eq(S.reduce(S.add, 0, [1, 2, 3, 4, 5]), 15);
+      eq(S.reduce(S.add, 0, []), 0);
+      eq(S.reduce(S.lift2(S.add), S.Just(0),
+                  [S.Just(1), S.Just(2), S.Just(3), S.Just(4), S.Just(5)]),
+         S.Just(15));
+    });
+
+    it('dispatches to a reduce method if present', function() {
```
davidchambers: `s/reduce/"reduce"/`

index.js
```diff
@@ -2171,6 +2179,39 @@
...
+      [$.Function, a, b, a],
+      function(f, initial, foldable) {
+        var acc = initial;
+        if (_type(foldable) === 'Array') {
+          for (var idx = 0; idx < foldable.length; idx += 1) {
+            acc = f(acc, foldable[idx]);
+          }
+          return acc;
+        } else {
+          return foldable.reduce(f, acc);
```
davidchambers: I think `foldable.reduce(f, initial)` reads slightly better. We could then move `var acc = initial;` into the `if` block.

test/index.js
svozza: No, not at all, I just copied and pasted from the Maybe#reduce tests.

svozza:
:zap: x 3

davidchambers:
:deciduous_tree: Feel free to merge this once the tests pass. :)

svozza:
:tada:

Issue #139, opened 2016-01-31
more combinators


davidchambers:
@Avaq raised the issue of combinators on Gitter. Sanctuary currently exports [`I`][1] and [`K`][2], but [fantasy-combinators][3] defines several more, neatly summarized in a [gist][4]:

> ```javascript
> const I = x => x;
> const K = x => _ => x;
> const A = f => x => f(x);
> const T = x => f => f(x);
> const C = f => y => x => f(x)(y);
> const B = f => g => x => f(g(x));
> const S = f => g => x => f(x)(g(x));
> const P = f => g => x => y => f(g(x))(g(y));
> const Y = f => (g => g(g))(g => f(x => g(g)(x)));
> ```

`A` is a strict version of [`R.call`][5] (since it expects a unary function).

`C` is a strict version of [`S.flip`][6] (since it expects a curried function).

`B` is [`S.compose`][7].

I'd like Sanctuary to provide more of these combinators. I'd welcome a pull request for one or more of these functions. :)

[1]: https://github.com/plaid/sanctuary#I
[2]: https://github.com/plaid/sanctuary#K
[3]: https://github.com/fantasyland/fantasy-combinators
[4]: https://gist.github.com/Avaq/1f0636ec5c8d6aed2e45#file-combinators-js
[5]: http://ramdajs.com/docs/#call
[6]: https://github.com/plaid/sanctuary#flip
[7]: https://github.com/plaid/sanctuary#compose

Avaq:
Hi @davidchambers!

In all honesty I have to say I'm still a bit skeptical about the decision to add *more combinators* to Sanctuary. There are several points I'd like to make.

----

Firstly, as I've come to understand, combinators have their roots in Lambda calculus. The Y combinator specifically -I believe- is a way to gain recursion in an environment where there is no variable assignment other than Lambda binding. That's the main problem is solves. The point I'm trying to make here is that we *can* do other types of variable assignment in JavaScript, so it's really quite easy to create recursive functions without help from the Y combinator:

```js
//With Y:
const factorial = Y(f => n => n < 1 ? 1 : n * f(n - 1));
//Easier to read and just as easy to write without Y:
const factorial = n => n < 1 ? 1 : n * factorial(n - 1);
```

So the only application it has in JavaScript (as far as I can see), is to enable writing point-free recursive functions:

```js
//Without Y:
const helper = n => factorial(n); //I need this pointful helper :(
[...]

PR #151, opened 2016-02-06
More combinators

Avaq:
For #139

* Add the A combinator
* Add the C combinator
* Add the B combinator
* Add the S combinator

### DIFF
--- a/index.js
+++ b/index.js
@@ -438,6 +438,24 @@
       [$.Function, $.Function, a, c],
       compose3);
 
+  //# S :: (a -> b -> c) -> (a -> b) -> a -> c
+  //.
+  //. The S combinator. Takes a curried binary function, a unary function,
+  //. and a value, and returns the result of applying the binary function to:
+  //.
+  //.   - the value; and
+  //.   - the result of applying the unary function to the value.
+  //.
+  //. ```javascript
+  //. > S.S(R.add, Math.sqrt, 100)
+  //. 110
+  //. ```
+  S.S =
+  def('S',
+      {},
+      [$.Function, $.Function, a, c],
+      function(f, g, x) { return f(x)(g(x)); });
+
   //. ### Function
 
   //# flip :: (a -> b -> c) -> b -> a -> c
--- a/index.js
+++ b/index.js
@@ -274,6 +274,10 @@
     };
   };
 
+  var compose3 = function(f, g, x) {
+    return f(g(x));
+  };
+
   var filter = function(pred, m) {
     return m.chain(function(x) {
       return pred(x) ? m.of(x) : m.empty();
@@ -418,6 +422,22 @@
       [$.Function, b, a, c],
       function(f, x, y) { return f(y)(x); });
 
+  //# B :: (b -> c) -> (a -> b) -> a -> c
+  //.
+  //. The B combinator. Takes two functions and a value, and returns the result
+  //. of applying the first function to the result of applying the second to the
+  //. value. Equivalent to [`compose`](#compose) and Haskell's `(.)` function.
+  //.
+  //. ```javascript
+  //. > S.B(Math.sqrt, S.inc, 99)
+  //. 10
+  //. ```
+  S.B =
+  def('B',
+      {},
+      [$.Function, $.Function, a, c],
+      compose3);
+
   //. ### Function
 
   //# flip :: (a -> b -> c) -> b -> a -> c
@@ -509,7 +529,7 @@
   //. In general terms, `compose` performs right-to-left composition of two
   //. unary functions.
   //.
-  //. See also [`pipe`](#pipe).
+  //. See also [`B`](#B) and [`pipe`](#pipe).
   //.
   //. ```javascript
   //. > S.compose(Math.sqrt, S.inc)(99)
@@ -519,7 +539,7 @@
   def('compose',
       {},
       [$.Function, $.Function, a, c],
-      function(f, g, x) { return f(g(x)); });
+      compose3);
 
   //# pipe :: [(a -> b), (b -> c), ..., (m -> n)] -> a -> n
   //.
--- a/index.js
+++ b/index.js
@@ -395,6 +395,29 @@
       [$.Function, a, b],
       function(f, x) { return f(x); });
 
+  //# C :: (a -> b -> c) -> b -> a -> c
+  //.
+  //. The C combinator. Takes a curried binary function and two values, and
+  //. returns the result of applying the function to the values in reverse.
+  //. Equivalent to Haskell's `flip` function.
+  //.
+  //. This function is very similar to [`flip`](#flip), except that its first
+  //. argument must be curried. This allows it to work with manually curried
+  //. functions.
+  //.
+  //. ```javascript
+  //. > S.C(R.concat, 'foo', 'bar')
+  //. 'barfoo'
+  //.
+  //. > R.filter(S.C(R.gt, 0), [-1, -2, 3, -4, 4, 2])
+  //. [3, 4, 2]
+  //. ```
+  S.C =
+  def('C',
+      {},
+      [$.Function, b, a, c],
+      function(f, x, y) { return f(y)(x); });
+
   //. ### Function
 
   //# flip :: (a -> b -> c) -> b -> a -> c
@@ -403,6 +426,8 @@
   //. applying the function - with its argument order reversed - to the
   //. values. `flip` may also be applied to a Ramda-style curried
   //. function with arity greater than two.
+  //.
+  //. See also [`C`](#C).
   //.
   //. ```javascript
   //. > R.map(S.flip(Math.pow)(2), [1, 2, 3, 4, 5])
--- a/index.js
+++ b/index.js
@@ -377,6 +377,24 @@
       [a, b, a],
       function(x, y) { return x; });
 
+  //# A :: (a -> b) -> a -> b
+  //.
+  //. The A combinator. Takes a function and a value, and returns the result of
+  //. applying the function to the value. Equivalent to Haskell's `($)` function.
+  //.
+  //. ```javascript
+  //. > S.A(R.inc, 1)
+  //. 2
+  //.
+  //. > R.map(S.A(R.__, 100), [R.inc, Math.sqrt])
+  //. [101, 10]
+  //. ```
+  S.A =
+  def('A',
+      {},
+      [$.Function, a, b],
+      function(f, x) { return f(x); });
+
   //. ### Function
 
   //# flip :: (a -> b -> c) -> b -> a -> c
--- a/test/index.js
+++ b/test/index.js
@@ -350,6 +350,26 @@ describe('combinator', function() {
 
   });
 
+  describe('C', function() {
+
+    it('is a ternary function', function() {
+      eq(typeof S.C, 'function');
+      eq(S.C.length, 3);
+    });
+
+    it('C(f, x, y) is equivalent to f(y)(x)', function() {
+      eq(S.C(R.concat, 'foo', 'bar'), 'barfoo');
+      eq(R.map(S.C(R.concat, '!'), ['BAM', 'POW', 'KA-POW']), ['BAM!', 'POW!', 'KA-POW!']);
+    });
+
+    it('is curried', function() {
+      eq(S.C(R.concat).length, 2);
+      eq(S.C(R.concat)('foo').length, 1);
+      eq(S.C(R.concat)('foo')('bar'), 'barfoo');
+    });
+
+  });
+
 });
 
 describe('function', function() {
--- a/test/index.js
+++ b/test/index.js
@@ -145,6 +145,25 @@ var IdentityArb = function(arb) {
   return arb.smap(Identity, function(i) { return i.value; });
 };
 
+var runCompositionTests = function(compose) {
+
+  it('is a ternary function', function() {
+    eq(typeof compose, 'function');
+    eq(compose.length, 3);
+  });
+
+  it('composes two functions assumed to be unary', function() {
+    eq(compose(R.map(Math.sqrt), JSON.parse, '[1, 4, 9]'), [1, 2, 3]);
+  });
+
+  it('is curried', function() {
+    eq(compose(R.map(Math.sqrt)).length, 2);
+    eq(compose(R.map(Math.sqrt))(JSON.parse).length, 1);
+    eq(compose(R.map(Math.sqrt))(JSON.parse)('[1, 4, 9]'), [1, 2, 3]);
+  });
+
+};
+
 describe('invariants', function() {
 
   it('f() is equivalent to f for every "regular" function', function() {
@@ -370,6 +389,12 @@ describe('combinator', function() {
 
   });
 
+  describe('B', function() {
+
+    runCompositionTests(S.B);
+
+  });
+
 });
 
 describe('function', function() {
@@ -498,20 +523,7 @@ describe('composition', function() {
 
   describe('compose', function() {
 
-    it('is a ternary function', function() {
-      eq(typeof S.compose, 'function');
-      eq(S.compose.length, 3);
-    });
-
-    it('composes two functions assumed to be unary', function() {
-      eq(S.compose(R.map(Math.sqrt), JSON.parse, '[1, 4, 9]'), [1, 2, 3]);
-    });
-
-    it('is curried', function() {
-      eq(S.compose(R.map(Math.sqrt)).length, 2);
-      eq(S.compose(R.map(Math.sqrt))(JSON.parse).length, 1);
-      eq(S.compose(R.map(Math.sqrt))(JSON.parse)('[1, 4, 9]'), [1, 2, 3]);
-    });
+    runCompositionTests(S.compose);
 
   });
 
--- a/test/index.js
+++ b/test/index.js
@@ -331,6 +331,25 @@ describe('combinator', function() {
 
   });
 
+  describe('A', function() {
+
+    it('is a binary function', function() {
+      eq(typeof S.A, 'function');
+      eq(S.A.length, 2);
+    });
+
+    it('A(f, x) is equivalent to f(x)', function() {
+      eq(S.A(R.inc, 1), 2);
+      eq(R.map(S.A(R.__, 100), [R.inc, Math.sqrt]), [101, 10]);
+    });
+
+    it('is curried', function() {
+      eq(S.A(R.inc).length, 1);
+      eq(S.A(R.inc)(1), 2);
+    });
+
+  });
+
 });
 
 describe('function', function() {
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -256,6 +256,10 @@
     };
   };
 
+  var compose3 = function(f, g, x) {
+    return f(g(x));
+  };
+
```
Avaq: I wasn't sure how to deal with aliases in Sanctuary Def. Just going `S.compose = S.I` would give `compose` the name of `I`, going `def('compose', ..., S.I)` would do double type-checking I think. So this is what I ended up doing.

test/index.js
```diff
@@ -144,6 +144,25 @@ var IdentityArb = function(arb) {
...
+  });
+
+  it('is curried', function() {
+    eq(compose(R.map(Math.sqrt)).length, 2);
+    eq(compose(R.map(Math.sqrt))(JSON.parse).length, 1);
+    eq(compose(R.map(Math.sqrt))(JSON.parse)('[1, 4, 9]'), [1, 2, 3]);
+  });
+
+};
+
```
Avaq: This is here because I didn't want to repeat existing code, yet I wanted both functions (`S.compose` and `S.B`) to be tested, because they are in fact different functions, despite being aliases.

index.js
davidchambers: Thanks for the explanation. I agree with your conclusion.

It's unfortunate that we have both `compose2` and `compose3`, but this doesn't bother me greatly.

index.js
```diff
@@ -359,6 +363,83 @@
       [a, b, a],
       function(x, y) { return x; });
 
+  //# A :: (a -> b) -> a -> b
+  //.
+  //. The A combinator. Takes a function and a value and returns the result of
+  //. applying the function to the value. Equivalent to Haskell's `($)` function.
+  //.
+  //. ```javascript
+  //. > S.A(R.add(1), 1)
```
davidchambers: Let's use `S.inc` in place of `R.add(1)`. :)

index.js
```diff
@@ -359,6 +363,83 @@
...
+
+  //# C :: (a -> b -> c) -> b -> a -> c
+  //.
+  //. The C combinator. Takes a curried binary function and two values and
+  //. returns the result of applying the function to the values in reverse.
+  //. Equivalent to Haskell's `flip` function.
+  //.
+  //. This function is very similar to `S.flip`, except that its first argument
+  //. must be curried. This also allows it to work with manually curried
+  //. functions.
```
davidchambers: I think we could drop the word "also".

index.js
```diff
@@ -359,6 +363,83 @@
...
+  //.
+  //. ```javascript
+  //. > R.gt(1, 2)
+  //. false
+  //.
+  //. > S.C(R.gt, 1, 2)
+  //. true
```
davidchambers: I'd like to avoid `R.gt` due to confusion surrounding its argument order. How about this instead?

```javascript
> S.C(R.concat, 'foo', 'bar')
'barfoo'
```

index.js
```diff
@@ -359,6 +363,83 @@
...
+  //.
+  //. > R.map(S.C(S.A, 3), [R.add(2), R.multiply(2)])
+  //. [5, 6]
```
davidchambers: I find this example mind-bending. I'd prefer to avoid using `S.A` in an example for `S.C`. How about this instead?

```javascript
> R.map(S.C(R.concat, '!'), ['BAM', 'POW', 'KA-POW'])
['BAM!', 'POW!', 'KA-POW!']
```

Using `R.concat`, a noncommutative function, nicely conveys the "flipping" nature of the C combinator.

index.js
```diff
@@ -359,6 +363,83 @@
...
+
+  //# B :: (b -> c) -> (a -> b) -> a -> c
+  //.
+  //. The B combinator. Takes two functions and a value and returns the result
+  //. of applying the first function to the result of applying the second to the
+  //. value. Equivalent to Haskells `(.)` function and to `S.compose`.
+  //.
+  //. ```javascript
+  //. > S.B(R.multiply(2), R.add(2), 0)
+  //. 4
```
davidchambers: I'd like to use the [`S.compose`][1] example here as well:

```javascript
> S.B(Math.sqrt, S.inc, 99)
10
```

Unary functions such as `Math.sqrt` and `S.inc` are slightly less noisy than partially applied binary functions, yet convey the same idea.


[1]: https://github.com/plaid/sanctuary#compose

index.js
```diff
@@ -359,6 +363,83 @@
...
+  def('C',
+      {},
+      [$.Function, b, a, c],
+      function(f, x, y) { return f(y)(x); });
+
+  //# B :: (b -> c) -> (a -> b) -> a -> c
+  //.
+  //. The B combinator. Takes two functions and a value and returns the result
+  //. of applying the first function to the result of applying the second to the
+  //. value. Equivalent to Haskells `(.)` function and to `S.compose`.
```
davidchambers: `s/Haskells/Haskell's/`

index.js
```diff
@@ -359,6 +363,83 @@
...
+
+  //# S :: (a -> b -> c) -> (a -> b) -> a -> c
+  //.
+  //. The S combinator. Takes a curried binary function, a unary function and a
+  //. value and returns the result of applying the binary function to the value
+  //. along with the result of applying the unary function to the value. This
+  //. is known as the substitution combinator.
+  //.
+  //. ```javascript
+  //. > S.S(R.add, R.multiply(2), 1)
```
davidchambers: Let's use a unary function here rather than a partially applied binary function.

index.js
```diff
@@ -458,7 +539,7 @@
   def('compose',
       {},
       [$.Function, $.Function, a, c],
-      function(f, g, x) { return f(g(x)); });
+      compose3);
```
davidchambers: We should make this change a few lines above:

```diff
-  //. See also [`pipe`](#pipe).
+  //. See also [`B`](#B) and [`pipe`](#pipe).
```

Let's include a "see also" link in the description of [`S.flip`][1] as well.


[1]: https://github.com/plaid/sanctuary#flip

test/index.js
davidchambers: As @svozza will confirm, I don't mind duplication in test suites. :)

Having these functions share their tests is reasonable, though.

test/index.js
```diff
@@ -330,6 +349,70 @@ describe('combinator', function() {
...
 
+  describe('A', function() {
+
+    it('is a binary function', function() {
+      eq(typeof S.A, 'function');
+      eq(S.A.length, 2);
+    });
+
+    it('returns the result of applying its first argument to its second', function() {
+      eq(S.A(R.add(1), 1), 2);
```
davidchambers: Let's use `S.inc` in place of `R.add(1)`.

test/index.js
```diff
@@ -330,6 +349,70 @@ describe('combinator', function() {
...
+      eq(R.map(function(f) { return S.A(f, 3); }, [R.add(2), R.multiply(2)]), [5, 6]);
```
davidchambers: We could use `S.A(R.__, 3)` in place of the lambda. What do you think?

test/index.js
```diff
@@ -330,6 +349,70 @@ describe('combinator', function() {
...
+
+  describe('C', function() {
+
+    it('is a ternary function', function() {
+      eq(typeof S.C, 'function');
+      eq(S.C.length, 3);
+    });
+
+    it('applies its third and second argument to its first and returns the result', function() {
+      eq(S.C(R.gt, 1, 2), true);
```
davidchambers: As suggested above, let's avoid `R.gt` here.

index.js
```diff
@@ -359,6 +363,83 @@
...
+  //# A :: (a -> b) -> a -> b
+  //.
+  //. The A combinator. Takes a function and a value and returns the result of
+  //. applying the function to the value. Equivalent to Haskell's `($)` function.
+  //.
+  //. ```javascript
+  //. > S.A(R.add(1), 1)
+  //. 2
+  //.
+  //. > R.map(S.flip(S.A, 3), [R.add(2), R.multiply(2)])
```
davidchambers: I think it would be nice to use `S.A(R.__, 3)` here. Do you agree?

index.js
Avaq: I actually had that with `"world!Hello "` before I changed it because I wanted to show a more practical example. I ended up showing how you can "fix" `R.gt`'s argument order using this combinator. I'll change it back if you think it's confusing though.

index.js
Avaq: Good idea, will do.

index.js
Avaq: :+1:

test/index.js
Avaq: Ah right. I ended up using a lambda because I didn't want to use `flip` or `C` before they were tested. The placeholder didn't come to mind.

index.js
Avaq: Agreed.

Avaq:
:information_source: I'm going to commit the changes as a result of your feedback on top. I'll neatly squash everything back down to the four commits once we're done. :)

index.js
Avaq: It goes to show how important we find composition, that we have four nearly identical implementations of it in a single code-base. ;)

index.js
Avaq: Should I add links by the combinators as well (`see also S.flip` etc)?

index.js
Avaq: I will admit `"barfoo"` is a lot less clumsy-looking than `"world!Hello "`.

test/index.js
```diff
@@ -330,6 +349,70 @@ describe('combinator', function() {
...
+  });
+
+  describe('C', function() {
+
+    it('is a ternary function', function() {
+      eq(typeof S.C, 'function');
+      eq(S.C.length, 3);
+    });
+
+    it('applies its third and second argument to its first and returns the result', function() {
```
Avaq: I think this sentence is wrong. I'm applying values to functions here.

test/index.js
```diff
@@ -330,6 +349,70 @@ describe('combinator', function() {
...
+  });
+
+  describe('S', function() {
+
+    it('is a ternary function', function() {
+      eq(typeof S.C, 'function');
+      eq(S.C.length, 3);
+    });
+
+    it('applies its third, and its third applied to its second, to its first argument', function() {
```
Avaq: I think this sentence is wrong. I'm applying values to functions here. Bah, it was quite a challenge to keep it short.

index.js
davidchambers: Ah, I see now why you chose this example. To show the problem with `R.gt` we'd need to partially apply the function in an expression such as:

```javascript
> R.map(S.C(R.gt, 0), [-1, -2, 3, -4, 4, 2])
[3, 4, 2]
```

I like the opening example to be trivial, even if it doesn't demonstrate realistic usage. I'd be happy with something like the above as the second example, so feel free to replace the `'KA-POW!'` one. :)

index.js
davidchambers: Yep! In cases where we already mention the similar function in the description this isn't necessary, though we should ensure that each reference is a link to the relevant section of the documentation.

test/index.js
davidchambers: How about this?

```javascript
it('S.S(f, g, x) is equivalent to f(x)(g(x))', ...);
```

It doesn't read as a sentence beginning with `it`, but I don't let this worry me. :)

index.js
```diff
@@ -359,6 +363,82 @@
...
+      [$.Function, a, b],
+      function(f, x) { return f(x); });
+
+  //# C :: (a -> b -> c) -> b -> a -> c
+  //.
+  //. The C combinator. Takes a curried binary function and two values and
+  //. returns the result of applying the function to the values in reverse.
+  //. Equivalent to Haskell's `flip` function.
+  //.
+  //. This function is very similar to `S.flip`, except that its first argument
```
davidchambers: Let's use ``[`flip`](#flip)`` in place of `` `S.flip` ``.

index.js
```diff
@@ -359,6 +363,82 @@
...
+  def('C',
+      {},
+      [$.Function, b, a, c],
+      function(f, x, y) { return f(y)(x); });
+
+  //# B :: (b -> c) -> (a -> b) -> a -> c
+  //.
+  //. The B combinator. Takes two functions and a value and returns the result
+  //. of applying the first function to the result of applying the second to the
+  //. value. Equivalent to Haskell's `(.)` function and to `S.compose`.
```
davidchambers: Let's use ``[`compose`](#compose)`` in place of `` `S.compose` ``.

test/index.js
```diff
@@ -330,6 +349,70 @@ describe('combinator', function() {
...
   });
 
+  describe('A', function() {
+
+    it('is a binary function', function() {
+      eq(typeof S.A, 'function');
+      eq(S.A.length, 2);
+    });
+
+    it('returns the result of applying its first argument to its second', function() {
```
davidchambers: For consistency, let's write `'A(f, x) is equivalent to f(x)'`.

index.js
```diff
@@ -359,6 +363,80 @@
       [a, b, a],
       function(x, y) { return x; });
 
+  //# A :: (a -> b) -> a -> b
+  //.
+  //. The A combinator. Takes a function and a value and returns the result of
```
davidchambers: I'd like a comma after "value".

index.js
```diff
@@ -359,6 +363,80 @@
...
+  //. ```
+  S.A =
+  def('A',
+      {},
+      [$.Function, a, b],
+      function(f, x) { return f(x); });
+
+  //# C :: (a -> b -> c) -> b -> a -> c
+  //.
+  //. The C combinator. Takes a curried binary function and two values and
```
davidchambers: I'd like a comma after "values".

index.js
```diff
@@ -359,6 +363,80 @@
...
+  //. ```
+  S.C =
+  def('C',
+      {},
+      [$.Function, b, a, c],
+      function(f, x, y) { return f(y)(x); });
+
+  //# B :: (b -> c) -> (a -> b) -> a -> c
+  //.
+  //. The B combinator. Takes two functions and a value and returns the result
```
davidchambers: I'd like a comma after "value".

index.js
```diff
@@ -359,6 +363,80 @@
...
+  S.B =
+  def('B',
+      {},
+      [$.Function, $.Function, a, c],
+      compose3);
+
+  //# S :: (a -> b -> c) -> (a -> b) -> a -> c
+  //.
+  //. The S combinator. Takes a curried binary function, a unary function and a
+  //. value and returns the result of applying the binary function to the value
```
davidchambers: I'd like to add two commas:

    Takes a curried binary function, a unary function, and a value, and returns...
                                                     ^            ^

index.js
```diff
@@ -359,6 +363,80 @@
...
+  def('C',
+      {},
+      [$.Function, b, a, c],
+      function(f, x, y) { return f(y)(x); });
+
+  //# B :: (b -> c) -> (a -> b) -> a -> c
+  //.
+  //. The B combinator. Takes two functions and a value and returns the result
+  //. of applying the first function to the result of applying the second to the
+  //. value. Equivalent to Haskell's `(.)` function and [`compose`](#compose).
```
davidchambers: The last sentence would be slightly clearer reordered:

```
Equivalent to `compose` and Haskell's `(.)` function.
```

The current ordering permits two interpretations:

```
Equivalent to (Haskell's `(.)` function) and `compose`.
```

```
Equivalent to Haskell's (`(.)` function and `compose`).
```

index.js
Avaq: Are you sure about that first comma? School taught me a list is written like `a, b, c and d`. I wouldn't want to break off the last item from the list by placing that comma. I totally agree with the second comma though, because the rest of the sentence shouldn't be confused as being part of the list.

---

Some closer research shows that this rule differs between the UK and the US. The US *does* place that extra "[oxford comma](https://en.wikipedia.org/wiki/Serial_comma)". Since you're in the US, and the rest of the documentation is probably in en_US, I'll just add that extra comma. :)

davidchambers: For what it's worth I grew up in New Zealand, which tends to be closer to the UK than the US on such matters. I don't recall learning this one way or the other, but since the comma is necessary to [resolve ambiguity][1] in some cases, I like to use it in all cases. When I say _a b c and d_ I do pause after the _c_, so the comma seems natural to me.

Regardless of the merits of one approach or the other, and regardless of my preferences, maintaining consistency with the style used elsewhere in the file is worthwhile. :)


[1]: https://en.wikipedia.org/wiki/Serial_comma#Ambiguity

index.js
```diff
@@ -359,6 +363,80 @@
...
+      {},
+      [$.Function, $.Function, a, c],
+      compose3);
+
+  //# S :: (a -> b -> c) -> (a -> b) -> a -> c
+  //.
+  //. The S combinator. Takes a curried binary function, a unary function, and a
+  //. value, and returns the result of applying the binary function to the value
+  //. along with the result of applying the unary function to the value. This
+  //. is known as the substitution combinator.
```
davidchambers: This description doesn't seem quite right to me. We say that `S` "returns the result of _this_ along with the result of _that_"; it sounds as though `S` returns a pair of values. Could you tweak the sentence to avoid this misinterpretation? I realize this function is tricky to describe in words!

davidchambers:
Lovely work, Aldwin! At some point I'd like you to squash your commits and `git rebase master`. :)

Avaq:
I'll get straight to it. Would you like them as the original four commits (one per combinator) or rather all in one?

index.js
Avaq: Hey, this only showed up once I refreshed the page. Is this still a concern? It appears as though you gave the :+1: after.

Avaq: This is the best human sentence I could come up with, unfortunately. I'm no native speaker. The following statement, however..:
`S(f, g, x)` is like `f(x, g(x))`
..Makes it very clear to me what's going on. We might consider adding that to the description to improve clarity, though I'm not sure if that would work for all people.

davidchambers: > I'm no native speaker.

I'd never have guessed. Your written English is excellent.

I thought of a way to format your description to remove the ambiguity:

```javascript
  //. The S combinator. Takes a curried binary function, a unary function,
  //. and a value, and returns the result of applying the binary function to:
  //.
  //.   - the value; and
  //.   - the result of applying the unary function to the value.
  //.
  //. This is also known as the substitution combinator.
```

davidchambers:
> Would you like them as the original four commits (one per combinator) or rather all in one?

One commit per combinator makes sense to me. I :heart: the care you've taken, and the notes you've included with each commit message. I tend to be a bit lazy with commit messages since I assume I'll always be able to refer to pull request descriptions, but having these notes in the Git history is even better. Thanks!

davidchambers:
:deciduous_tree: Thank you for the high-quality patches. Welcome to the Sanctuary team!

PR #152, opened 2016-02-07
Use $.Array Instead of List for S.pipe and S.meld

svozza:
As discussed in #148 .


davidchambers:
:deciduous_tree: Thanks!

PR #153, opened 2016-02-17
ValidDate satisfies the requirements of Ord

davidchambers:
This may be of interest to you, @svozza, given your recent work on various date libraries. :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -180,11 +180,7 @@
   //  Ord :: TypeClass
   var Ord = $.TypeClass(
     'sanctuary/Ord',
-    function(x) {
-      var type = _type(x);
-      return type === 'Number' && $.ValidNumber.test(x) ||
-             type === 'String';
-    }
+    R.anyPass([$.String.test, $.ValidDate.test, $.ValidNumber.test])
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -180,11 +180,7 @@
   //  Ord :: TypeClass
   var Ord = $.TypeClass(
     'sanctuary/Ord',
-    function(x) {
-      var type = _type(x);
-      return type === 'Number' && $.ValidNumber.test(x) ||
-             type === 'String';
-    }
+    R.anyPass([$.String.test, $.ValidDate.test, $.ValidNumber.test])
```
svozza: Any reason not to use `S.anyPass` here?

davidchambers: To be honest, I forgot about [`S.anyPass`][1]! At this point in the file, though, that function has not yet been defined so we can't reference it. I think sticking with the Ramda equivalent is reasonable under these circumstances.


[1]: https://github.com/plaid/sanctuary#anyPass

svozza: Cool. LGTM then.

PR #154, opened 2016-02-17
sanctuary-def@0.4.x

davidchambers:
`sanctuary-def@0.4.0` has not yet been released, but this needn't stop us from preparing.

In plaid/sanctuary-def#28 we made type checking optional. The most straightforward change would be to replace `$.create(env)` with `$.create(true, env)`. Sanctuary's slowness, though, was the motivation for making type checking optional in sanctuary-def, so Sanctuary must also provide a way to opt out of type checking.

The solution employed in this pull request is to provide two versions of the Sanctuary module: one with type checking; one without.

Sanctuary, with type checking:

```javascript
```

Sanctuary, without type checking:

```javascript
const S = require('sanctuary').unchecked;
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -102,6 +102,44 @@
 //.
 //. `Number` is the sole inhabitant of the TypeRep Number type.
 //.
+//. ## Type checking
+//.
+//. Sanctuary functions are defined via [sanctuary-def][] to provide run-time
+//. type checking. This is tremendously useful during development: type errors
+//. are reported immediately, avoiding circuitous stack traces (at best) and
+//. silent failures due to type coercion (at worst). For example:
+//.
+//. ```javascript
+//. S.inc('XXX');
+//. // ! TypeError: ‘inc’ expected a value of type FiniteNumber as its first argument; received "XXX"
+//. ```
+//.
+//. Compare this to the behaviour of Ramda's unchecked equivalent:
+//.
+//. ```javascript
+//. R.inc('XXX');
+//. // => '1XXX'
+//. ```
+//.
+//. There is a performance cost to run-time type checking. One may wish to
+//. disable type checking in certain contexts to avoid paying this cost.
+//. There are actually two versions of the Sanctuary module: one with type
+//. checking; one without. The latter is accessible via the `unchecked`
+//. property of the former.
+//.
+//. When application of `S.unchecked.<name>` honours the function's type
+//. signature the result will be the same as if `S.<name>` had been used
+//. instead. Otherwise, the behaviour is unspecified.
+//.
+//. In Node, one could use an environment variable to determine which version
+//. of the Sanctuary module to use:
+//.
+//. ```javascript
+//. const S = process.env.NODE_ENV === 'production' ?
+//.             require('sanctuary').unchecked :
+//.             require('sanctuary');
+//. ```
+//.
 //. ## API
 
 /* global define, self */
@@ -123,11 +161,34 @@
 
   'use strict';
 
-  var S = {};
-
   var _ = R.__;
 
   var sentinel = {};
+
+  //  _type :: a -> String
+  var _type = function(x) {
+    return x != null && R.type(x['@@type']) === 'String' ? x['@@type']
+                                                         : R.type(x);
+  };
+
+  //  compose2 :: ((b -> c), (a -> b)) -> a -> c
+  var compose2 = function(f, g) {
+    return function(x) {
+      return f(g(x));
+    };
+  };
+
+  //  compose3 :: ((b -> c), (a -> b), a) -> c
+  var compose3 = function(f, g, x) {
+    return f(g(x));
+  };
+
+  //  filter :: (Monad m, Monoid m) => ((a -> Boolean), m a) -> m a
+  var filter = function(pred, m) {
+    return m.chain(function(x) {
+      return pred(x) ? m.of(x) : m.empty();
+    });
+  };
 
   //  hasMethod :: String -> Any -> Boolean
   var hasMethod = function(name) {
@@ -135,6 +196,15 @@
       return x != null && typeof x[name] === 'function';
     };
   };
+
+  //  inspect :: -> String
+  var inspect = /* istanbul ignore next */ function() {
+    return this.toString();
+  };
+
+  //  negativeZero :: a -> Boolean
+  var negativeZero = R.either(R.equals(-0),
+                              R.equals(new Number(-0)));  // jshint ignore:line
 
   //  Accessible :: TypeClass
   var Accessible = $.TypeClass(
@@ -197,7 +267,7 @@
   var r = $.TypeVariable('r');
 
   //  $Either :: Type -> Type -> Type
-  var $Either = S.EitherType = $.BinaryType(
+  var $Either = $.BinaryType(
     'sanctuary/Either',
     function(x) { return x != null && x['@@type'] === 'sanctuary/Either'; },
     function(either) { return either.isLeft ? [either.value] : []; },
@@ -219,7 +289,7 @@
   );
 
   //  $Maybe :: Type -> Type
-  var $Maybe = S.MaybeType = $.UnaryType(
+  var $Maybe = $.UnaryType(
     'sanctuary/Maybe',
     function(x) { return x != null && x['@@type'] === 'sanctuary/Maybe'; },
     function(maybe) { return maybe.isJust ? [maybe.value] : []; }
@@ -250,7 +320,14 @@
     $.ValidNumber
   ]);
 
-  var def = $.create(env);
+  //  createSanctuary :: Boolean -> Module
+  var createSanctuary = function(checkTypes) {
+
+  //  To avoid excessive indentation this function's body is not indented.
+
+  var S = {EitherType: $Either, MaybeType: $Maybe};
+
+  var def = $.create(checkTypes, env);
 
   var method = function(name, constraints, types, _f) {
     var f = def(name, constraints, types, _f);
@@ -258,36 +335,6 @@
       return R.apply(f, R.prepend(this, arguments));
     });
   };
-
-  //  _type :: a -> String
-  var _type = function(x) {
-    return x != null && R.type(x['@@type']) === 'String' ? x['@@type']
-                                                         : R.type(x);
-  };
-
-  var compose2 = function(f, g) {
-    return function(x) {
-      return f(g(x));
-    };
-  };
-
-  var compose3 = function(f, g, x) {
-    return f(g(x));
-  };
-
-  var filter = function(pred, m) {
-    return m.chain(function(x) {
-      return pred(x) ? m.of(x) : m.empty();
-    });
-  };
-
-  var inspect = /* istanbul ignore next */ function() {
-    return this.toString();
-  };
-
-  //  negativeZero :: a -> Boolean
-  var negativeZero = R.either(R.equals(-0),
-                              R.equals(new Number(-0)));  // jshint ignore:line
 
   //. ### Classify
 
@@ -2898,6 +2945,14 @@
       [$.Array($.String), $.String],
       compose(R.join(''), R.map(R.concat(_, '\n'))));
 
+  return S;
+
+  };
+
+  //  Export two versions of the Sanctuary module: one with type checking;
+  //  one without.
+  var S       = createSanctuary(true);
+  S.unchecked = createSanctuary(false);
   return S;
 
 }));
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
   },
   "dependencies": {
     "ramda": "0.19.x",
-    "sanctuary-def": "0.3.x"
+    "sanctuary-def": "plaid/sanctuary-def#master"
```
davidchambers: TODO: Change this value to `"0.4.x"`.

davidchambers: :zap:

index.js
```diff
@@ -102,6 +102,43 @@
...
+//. ```
+//.
+//. There is a performance cost to run-time type checking. One may wish to
+//. disable type checking in certain contexts to avoid paying this cost.
+//. There are actually two versions of the Sanctuary module: one with type
+//. checking; one without. The latter is accessible via the `unchecked`
+//. property of the former.
+//.
+//. When used correctly, `S.unchecked.<name>` will behave just like `S.<name>`.
+//. When used *incorrectly*, its behaviour is unspecified.
```
svozza: What do we mean by incorrectly here?

davidchambers: What I'm trying to say is that `S.unchecked.inc(42)` is equivalent to `S.inc(42)`, but the behaviour of `S.unchecked.inc('XXX')` is unspecified (whereas the behaviour of `S.inc('XXX')` is covered by the test suite).

Can you think of a better way to phrase this?

svozza: How about:

>when the use of `S.unchecked.<name>` honours the function's type signature it will behave...otherwise its behaviour is unspecified.

davidchambers: Very nice! Thank you.

davidchambers: I made this change:

```diff
diff --git a/index.js b/index.js
index 9c732ef..ea85cd6 100644
--- a/index.js
+++ b/index.js
@@ -127,8 +127,9 @@
 //. checking; one without. The latter is accessible via the `unchecked`
 //. property of the former.
 //.
-//. When used correctly, `S.unchecked.<name>` will behave just like `S.<name>`.
-//. When used *incorrectly*, its behaviour is unspecified.
+//. When application of `S.unchecked.<name>` honours the function's type
+//. signature the result will be the same as if `S.<name>` had been used
+//. instead. Otherwise, the behaviour is unspecified.
 //.
 //. In Node, one could use an environment variable to determine which version
 //. of the Sanctuary module to use:
```

index.js
```diff
@@ -2900,6 +2947,14 @@
...
   return S;
 
+  };
+
+  //  Export two versions of the Sanctuary module: one with type checking;
+  //  one without.
+  var S       = createSanctuary(true);
+  S.unchecked = createSanctuary(false);
+  return S;
+
```
davidchambers: I'd love to know your thoughts on this matter of API design, @Avaq, as we discussed this on Gitter.

Avaq: I have many conflicting thoughts about it.

I like the API because

- It's easy to consume:
  * `var S = require('sanctuary').unchecked`
  * `import {unchecked as S} from 'sanctuary'`
- It's backwards compatible

I dislike it because:

- Consumption will become weird and fiddly when doing "unchecked in production":
  * `var S = isProduction ? require('sanctuary').unchecked : require('sanctuary')`
  * `import sanctuary from 'sanctuary'; const S = isProduction ? sanctuary.unchecked : sanctuary`
- Somewhere, knowing that there's an unnecessary sanctuary living in memory alongside the one I use doesn't sit well with me.
- It's not "future proofed" against additional options.

After spending some thought on it, a possible alternative could be:

```js
var creator = memoize(createSanctuary);
creator.create = createSanctuary;
return export;
```

to be consumed like:

```js
var S = require('sanctuary')(isProduction);
//or
import Sanctuary from 'sanctuary';
const S = Sanctuary(isProduction);
```

Or if you *must* create new instances:

```js
var S = require('sanctuary').create(isProduction);
```

Advantages of this approach:

- It's more recognizable: You're just instantiating the Sanctuary class, like with many other Node API's
- It's very easy to do "unchecked in production".
- It allows adding more options in the future
- You only need one instance in memory

Disadvantages are:

- It's horribly backwards incompatible

davidchambers: Thank you for your thoughtful response, Aldwin.

> - Consumption will become weird and fiddly when doing "unchecked in production":
>   * `var S = isProduction ? require('sanctuary').unchecked : require('sanctuary')`
>   * `import sanctuary from 'sanctuary'; const S = isProduction ? sanctuary.unchecked : sanctuary`

I agree that this is fiddly. The fiddliness can, though, be contained to a single file:

```javascript
// sanctuary.js

'use strict'

const env = require('./env');

module.exports = env.isProduction ? S.unchecked : S;
```

```javascript
// app.js

'use strict';

const S = require('./sanctuary');

// ...
```

> - Somewhere, knowing that there's an unnecessary sanctuary living in memory alongside the one I use doesn't sit well with me.

I agree that this is not ideal, though I give it a low weighting relative to the other considerations.

> - It's not "future proofed" against additional options.

I very much hope we won't ever need more options. :)

> ```javascript
> var S = require('sanctuary').create(isProduction);
> ```

The following comments relate to the approach quoted above.

> - It's more recognizable: You're just instantiating the Sanctuary class, like with many other Node API's

I find this a bit confusing. Am I creating an instance of some "class"? Does this "class" have a prototype? Is the idea that I can then mutate the instance (to add my own functions, say), or is the object shared in some way?

> - It's very easy to do "unchecked in production".

I can't argue with you on this one. :)

> - It allows adding more options in the future

As mentioned above, I very much hope this will never be necessary.

Were we to add options in future, we may at that point decide to switch to accepting a single options record rather than positional arguments. This would be particularly important if there were multiple Boolean arguments, to avoid the [Boolean trap][1]. I suggest we don't let extensibility influence this decision, since what will be best at some point in the future may be incompatible with what is best now.

> - You only need one instance in memory

This is a win, but a small one in my eyes. Sanctuary is inefficient in many respects. ;)

> - It's horribly backwards incompatible

I'm happy to break backwards compatibility in exchange for a significant improvement, so this point isn't very significant in and of itself. Only when we consider two options equally good should we favour the status quo.

I'd like to add two points for consideration:

  - Were one to `require('sanctuary').create(isProduction)` in N files, one would create N instances in memory. If N > 2, this is more wasteful than creating two modules up front. If, on the other hand, all N expressions result in references to a single module, I argue that the resemblance to instantiation hinders rather than aids understanding of the function's behaviour.

  - More significant than the question of backwards compatibility is the *experience* of importing Sanctuary. Currently a user can launch a REPL, `require('sanctuary')`, and begin interacting with the library. This is how we expect most packages to behave. Doing something else would be a minor inconvenience for experienced users; a source of confusion for new users. Expecting a new user to read documentation in order to have a satisfying first experience with the library is asking a lot.

Do you find this reasoning compelling, @Avaq? What are your thoughts, @svozza? @benperez?


[1]: http://ariya.ofilabs.com/2011/08/hall-of-api-shame-boolean-trap.html

kedashoe: The current solution is similar to using native bindings with node-postgres: https://github.com/brianc/node-postgres#native-bindings. I can't think of any other modules that have an alternate version like this off the top of my head.

It might be nice to have plain string versions that you can import directly. Something like

`import S from 'sanctuary/alt/unchecked'`
`import S from 'sanctuary/alt/checked'` (for completeness sake, same as just writing `import S from 'sanctuary'`)

This of course also does not scale well, but is backwards compatible.

Nothing to do with sanctuary, but you could easily write a babel plugin that creates the appropriate import for you so you could just write something like

`import S from 'env-import:sanctuary'`

and the plugin would change it to the appropriate import line

davidchambers: > The current solution is similar to using native bindings with node-postgres

That's a useful data point. It's reassuring to see that we independently reached the same conclusion.

> It might be nice to have plain string versions that you can import directly.

I'm reluctant to add complexity to the build step (which is currently non-existent). I'm not sure whether I'm being a pragmatic programmer (having learnt from experience) or whether I'm just being lazy. ;)

> Nothing to do with sanctuary, but you could easily write a babel plugin that creates the appropriate import for you

That's neat! I don't know much about Babel, but it seems very useful for front-end development.

kedashoe: > I'm reluctant to add complexity to the build step (which is currently non-existent). I'm not sure whether I'm being a pragmatic programmer (having learnt from experience) or whether I'm just being lazy. ;)

I don't know anything about publishing on npm, but I was just envisioning including actual files, eg

`alt/unchecked/index.js`
```js
module.exports = require('../../index.js').unchecked;
```
Does it not work like that?

davidchambers: > Does it not work like that?

For Node, we can do just as you suggest. For the browser, though, we want to provide a single source file so we'd need to use one of the various bundlers. There's nothing to prevent us from doing this, but there's an undeniable appeal to choosing an approach which works the same way for all environments without accommodation.

svozza:
I've sort of kept out of this because while I dislike initial the approach I hadn't got a better suggestion. I had thought that and options record had been ruled out by virtue of what was said in plaid/sanctuary-def/pull/28 but given David's comments in this thread I see it's still on the table.

I really think there's a danger of overthinking this, an options record with the type  `{ checkTypes :: Boolean }` is the simplest and most idiomatic way of achieving what we want here and avoids all the complexity of having two different versions of Sanctuary to manage. If we're worried about someone passing in an invalid option then we just do what we do everywhere else in the library and throw a descriptive type error. The options record could be mandatory if we don't want to rely on magic defaults.

```js
const options = R.objOf('checkTypes', NODE_ENV === 'production');
const S = Sanctuary(options);
```

Avaq:
> Were one to require('sanctuary').create(isProduction) in N files, one would create N instances in memory. If N > 2, this is more wasteful than creating two modules up front. If, on the other hand, all N expressions result in references to a single module, I argue that the resemblance to instantiation hinders rather than aids understanding of the function's behaviour. -- @davidchambers 

That's why I suggested the default export to be a memoized version of the factory. Maybe the non-memoized version could be a more descriptive name, along the lines of: `require('sanctuary').dangerouslyCreateANewInstanceEveryTime(env)`, you get the point. ;)

> I agree that this is fiddly. The fiddliness can, though, be contained to a single file -- @davidchambers

I would naturally put my sanctuary in a single file like that. I think one *more intuitively* reaches that solution if the API would *look like* other similar API's; like @svozza showed:

```js
// sanctuary.js
'use strict'
var Sanctuary = require('sanctuary');
var options = require('./env').sanctuary;
module.exports = Sanctuary(options);
```

I don't think this kind of API introduces confusion, or an OOP bias. Functional programmers can call it the memoized `createSanctuary` function, and people from an OOP background can call it the singleton `Sanctuary` class and even add `new` if they feel particularly old fashioned. The reason it's more intuitive in my opinion is because most other modules that have some kind of global configuration work like that, and I always create a separate file for them.

> It might be nice to have plain string versions that you can import directly. -- @kedashoe 

This can always be added to please some Node users who don't need to change things in different environments. All it takes is adding a `checked.js` and `unchecked.js` in the root of the project containing:

```js
//unchecked.js
module.exports = require('./').create({checkTypes: false})
```

At which point Node users can do `var S = require('sanctuary/unchecked')`, and still no extranous instances of Sanctuary are created! :clap:

davidchambers:
Thank you for your feedback, folks! I greatly appreciate it.

I raise issues for discussion because I value the opinions of others (this is particularly true when the Ramda/Sanctuary community is involved). Discussing an issue with others increases the likelihood of making a good decision. I can point to several occasions on which a comment on an issue or pull request has prevented me from going astray.

Discussion sometimes results in me convincing myself of my initially uncertain position. This has happened here. I've decided to merge this pull request—despite objections—for the following reasons:

  - I've decided it's very important for `require('sanctuary')` to continue to work, not for backwards compatibility but to avoid confusion for first-time users. I don't want to raise the barrier to entry.

  - I very much believe in this line from the _Zen of Python_:

    > There should be one—and preferably only one—obvious way to do it.

    I'm not in favour of supporting `require('sanctuary/unchecked')` because there would then be two ways to import the unchecked module. We can't use this method exclusively because it isn't applicable in the browser (at least, not without dictating the use of certain build tools, which Sanctuary has no right to do).

  - I don't find the memory-usage argument compelling. If and when we're able to demonstrate a problem arising from having two Sanctuary modules in memory, we'll reconsider this decision.

I plan to merge this pull request, preview the documentation changes since the last release, then publish v0.9.0. This will make lots of new functions available. :)

Avaq:
Awesome! :+1: Despite the minor disagreement I'm chuffed 0.9 is out. :)

PR #159, opened 2016-02-29
Add S.length function

svozza:
I went with the `Countable` suggestion as mentioned in #145.

### DIFF
--- a/index.js
+++ b/index.js
@@ -221,6 +221,14 @@
     }
   );
 
+  //  Countable :: TypeClass
+  var Countable = $.TypeClass(
+    'sanctuary/Countable',
+    function(x) {
+      return x != null && x.length != null && $.Integer.test(x.length);
+    }
+  );
+
   //  Foldable :: TypeClass
   var Foldable = $.TypeClass(
     'sanctuary/Foldable',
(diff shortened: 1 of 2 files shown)


davidchambers:
What's the appeal of this function over `R.prop('length')`? I'm curious. :)

svozza:
Hmmm, good question. Not a lot, aside from a more descriptive error message than:

> TypeError: Cannot read property 'length' of undefined

svozza:
How would that work? We couldn't do something like:

```js
def('prop', {}, [$.RecordType, $.Any]...)
```

davidchambers:
How about something like this?

```javascript
var prop =
def('prop',
    {a: [Accessible]},
    [$.String, a, b],
    function(name, obj) {
      if (name in obj) {
        return obj[name];
      } else {
        throw new TypeError('‘prop’ expected object to have a property named ' +
                            '‘' + name + '’; ' + show(obj) + ' does not');
      }
    });
```

svozza:
Do we want to traverse the prototype chain to find properties? I know not doing so would make this an O(n) function but I think it might be less surprising to users.

davidchambers:
How do you feel about `length` now that we have `prop`, Stefano?

svozza:
Only seeing your comment now, David. I think we can close this one. I really don't like this `Countable` type thing.

davidchambers:
Okay. Closing without prejudice.

PR #166, opened 2016-03-10
tweak doctests and implementation for vm.runInNewContext compatibility

davidchambers:
These changes are required in order for [sanctuary-site][1]'s build script to execute successfully.

[1]: https://github.com/plaid/sanctuary-site

### DIFF
--- a/index.js
+++ b/index.js
@@ -981,7 +981,7 @@
   //. > S.Nothing().sequence(S.Either.of)
   //. Right(Nothing())
   //.
-  //. > S.Just(Right(42)).sequence(S.Either.of)
+  //. > S.Just(S.Right(42)).sequence(S.Either.of)
   //. Right(Just(42))
   //. ```
   Maybe.prototype.sequence =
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -981,7 +981,7 @@
   //. > S.Nothing().sequence(S.Either.of)
   //. Right(Nothing())
   //.
-  //. > S.Just(Right(42)).sequence(S.Either.of)
+  //. > S.Just(S.Right(42)).sequence(S.Either.of)
```
davidchambers: Doctests have access to variables in scope, but this is not the case when evaluating these expressions in isolation via [`vm.runInNewContext`][1].


[1]: https://nodejs.org/api/vm.html#vm_vm_runinnewcontext_code_sandbox_options

index.js
```diff
@@ -2908,7 +2908,7 @@
   def('unwords',
       {},
       [$.Array($.String), $.String],
-      R.join(' '));
+      function(xs) { return xs.join(' '); });
```
svozza: :+1:

Issue #165, opened 2016-03-09
polymorphic functions cannot operate on "foreign" values


davidchambers:
#123 had one major unforeseen consequence.

To determine whether a value is a member of `Number`, say, sanctuary-def can ask `$.Number.test(x)`? To determine whether a value is a member of `a`, a type variable, sanctuary-def filters the set of types in the current environment to determine the set of types of which the value is a member, constraining `a`.

The problem is that Sanctuary provides a list of types to `$.create`, essentially stating _these are all the types that exist_. Sanctuary's polymorphic functions cannot operate on "unknown" types:

```javascript
S.I(document.body);
// ! TypeError: ‘I’ received [object HTMLBodyElement] as its first argument, but this value is not a member of any of the types in the environment:
//
//     - (Array ???)
//     - Boolean
//     ...
//     - ValidDate
//     - ValidNumber
```

```javascript
S.I({'@@type': 'my-package/MyType'});
// ! TypeError: ‘I’ received {"@@type": "my-package/MyType"} as its first argument, but this value is not a member of any of the types in the environment:
//
//     - (Array ???)
//     - Boolean
//     ...
//     - ValidDate
//     - ValidNumber
```

This is clearly a problem. What's more, [`S.type`][1] handles these "foreign" values:

```javascript
> S.type(document.body)
'HTMLBodyElement'

> S.type({'@@type': 'my-package/MyType'})
'my-package/MyType'
```

It's trivial to compare two type identifiers to determine whether two values are of the same type.

I believe the solution is as follows. When sanctuary-def determines that the value under interrogation is not a member of any of the types in the environment, it should not despair ([as it currently does][2]). Instead, it should calmly create a [`NullaryType`][3] and add it to the map:

```javascript
var typeIdentifier = $$type(value);
$typeVarMap[typeVarName] =
  [$.NullaryType(typeIdentifier, $$typeEq(typeIdentifier))];
```

It seems the fix should not be difficult. I hope to release a new version of sanctuary-def and a new version of Sanctuary this weekend.

Thank you @raqystyle for raising this issue on [Gitter][4].

[1]: https://github.com/plaid/sanctuary#type
[2]: https://github.com/plaid/sanctuary-def/blob/v0.4.0/index.js#L788-L790
[3]: https://github.com/plaid/sanctuary-def#nullarytype
[4]: https://gitter.im/plaid/sanctuary

soroushjp:
I have also run into this same issue while working on a project with @davidchambers.

[...]

PR #168, opened 2016-03-23
Change parseJson to perform type validation

### LINKED ISSUE #150: Even safer parseJson
> `parseJson(Array, '{}') -> Nothing()`

I'm very fond of the `get` and `gets` style of type validation, because it forces you to do it. It would've already saved me on multiple occasions if `parseJson` (or any `x -> Any` function for that matter) had this built-in, because after doing a "safe" JSON parse, it doesn't always occur to me the return value may still be of an unwanted type.

An alternative is to extract this behaviour out to its own function, which is used solely for type validation:

```js
validate :: TypeRep -> a -> Maybe a

const parseJsonArray = compose(chain(validate(Array)), parseJson);
```

I think this second solution is more elegant in terms of modularity but it lacks the safety that comes from forcing you to do the type validation.

Side note: `TypeRep` [may be changed to a predicate](https://github.com/plaid/sanctuary/issues/149), but the idea remains the same.

Avaq:
Adds a second argument to the `parseJson` function: A TypeRep which the result from parsing the JSON is validated against. This closes #150.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2760,24 +2760,28 @@
         )(s);
       });
 
-  //# parseJson :: String -> Maybe Any
+  //# parseJson :: TypeRep a -> String -> Maybe a
   //.
-  //. Takes a string which may or may not be valid JSON, and returns Just
-  //. the result of applying `JSON.parse` to the string if valid; Nothing
-  //. otherwise.
+  //. Takes a [type representative](#type-representatives) and a string which
+  //. may or may not be valid JSON, and returns Just the result of applying
+  //. `JSON.parse` to the string *if* the result is of the specified type
+  //. (according to [`is`](#is)); Nothing otherwise.
   //.
   //. ```javascript
-  //. > S.parseJson('["foo","bar","baz"]')
+  //. > S.parseJson(Array, '["foo","bar","baz"]')
   //. Just(['foo', 'bar', 'baz'])
   //.
-  //. > S.parseJson('[')
+  //. > S.parseJson(Array, '[')
+  //. Nothing()
+  //.
+  //. > S.parseJson(Object, '["foo","bar","baz"]')
   //. Nothing()
   //. ```
   S.parseJson =
   def('parseJson',
       {},
-      [$.String, $Maybe($.Any)],
-      encase(JSON.parse));
+      [TypeRep, $.String, $Maybe(a)],
+      function(type, s) { return filter(is(type), encase(JSON.parse, s)); });
 
   //. ### RegExp
 
--- a/test/index.js
+++ b/test/index.js
@@ -5125,19 +5125,31 @@ describe('parse', function() {
 
   describe('parseJson', function() {
 
-    it('is a unary function', function() {
+    it('is a binary function', function() {
       eq(typeof S.parseJson, 'function');
-      eq(S.parseJson.length, 1);
+      eq(S.parseJson.length, 2);
     });
 
     it('type checks its arguments', function() {
-      throws(function() { S.parseJson([1, 2, 3]); },
+      throws(function() { S.parseJson('String'); },
              errorEq(TypeError,
                      'Invalid value\n' +
                      '\n' +
-                     'parseJson :: String -> Maybe Any\n' +
-                     '             ^^^^^^\n' +
-                     '               1\n' +
+                     'parseJson :: TypeRep -> String -> Maybe a\n' +
+                     '             ^^^^^^^\n' +
+                     '                1\n' +
+                     '\n' +
+                     '1)  "String" :: String\n' +
+                     '\n' +
+                     'The value at position 1 is not a member of ‘TypeRep’.\n'));
+
+      throws(function() { S.parseJson(Array, [1, 2, 3]); },
+             errorEq(TypeError,
+                     'Invalid value\n' +
+                     '\n' +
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -2743,24 +2743,28 @@
...
+  //. > S.parseJson(Object, '["foo","bar","baz"]')
   //. Nothing()
   //. ```
   S.parseJson =
   def('parseJson',
       {},
-      [$.String, $Maybe($.Any)],
-      encase(JSON.parse));
+      [TypeRep, $.String, $Maybe(a)],
+      function(type, json) { return filter(is(type), encase(JSON.parse, json)); });
```
davidchambers: Let's name the second parameter `s` rather than `json`, as it's not necessarily a valid JSON string.

test/index.js
```diff
@@ -4123,24 +4123,34 @@ describe('parse', function() {
...
+      eq(S.parseJson(Array, '["foo","bar"]'), S.Just(['foo', 'bar']));
     });
 
     it('returns a Nothing when applied to an invalid JSON string', function() {
-      eq(S.parseJson('[Invalid JSON]'), S.Nothing());
+      eq(S.parseJson(Object, '[Invalid JSON]'), S.Nothing());
+    });
+
+    it('returns a Nothing when the parsed result is not a member of the given Type', function() {
+      eq(S.parseJson(Array, '{"foo": "bar"}'), S.Nothing());
```
davidchambers: Nit: Let's remove the space after the colon (`:`). I like to use the default `JSON.stringify` format.

davidchambers:
This is very nice, Aldwin!

index.js
```diff
@@ -2743,24 +2743,28 @@
...
 
-  //# parseJson :: String -> Maybe Any
+  //# parseJson :: TypeRep a -> String -> Maybe a
   //.
-  //. Takes a string which may or may not be valid JSON, and returns Just
-  //. the result of applying `JSON.parse` to the string if valid; Nothing
-  //. otherwise.
+  //. Takes a [type representative](#type-representatives) and a string which
+  //. may or may not be valid JSON, and returns Just the result of applying
+  //. `JSON.parse` to the string *if* the result is of the specified type
```
Avaq: What do you think about the cursive *if*? I thought it's a nice way to break up the long sentence.

test/index.js
```diff
@@ -4123,24 +4123,34 @@ describe('parse', function() {
...
-      eq(S.parseJson('["foo","bar"]'), S.Just(['foo', 'bar']));
+      eq(S.parseJson(Array, '["foo","bar"]'), S.Just(['foo', 'bar']));
     });
 
     it('returns a Nothing when applied to an invalid JSON string', function() {
-      eq(S.parseJson('[Invalid JSON]'), S.Nothing());
+      eq(S.parseJson(Object, '[Invalid JSON]'), S.Nothing());
+    });
+
+    it('returns a Nothing when the parsed result is not a member of the given Type', function() {
```
Avaq: This would be more correctly formulated as "when the parsed result is not a member of the type represented by the given TypeRep". But it was a little long. What do you think?

index.js
davidchambers: It works for me, but the one thing that's not clear from this description is the handling of invalid strings.

What do you think of this phrasing?

_… and returns Just `x` if the given string is a JSON representation of `x`, a value of the specified type (according to [`is`][1]); Nothing otherwise._


[1]: http://sanctuary.js.org/#is

test/index.js
davidchambers: The current description is fine, though we needn't capitalize "type" in this context.

index.js
Avaq: I thought about that, but then again I think only specifying the latter is good enough, because it implies the first. That combined with the code example makes it pretty clear. I'm worried going into the specifics take away from the ease of reading and clarity.

davidchambers: Okay. I'm happy with your current wording also.

Avaq:
@davidchambers Did you see I did the thing?

davidchambers:
Thanks, @Avaq. Now that we've merged plaid/sanctuary-def#38 I can publish a minor release of sanctuary-def and a patch release of this project (to fix #165). That will free us to start merging changes for the v0.10.0 release.

Avaq:
I see, so we're awaiting the release of `sanctuary-def` before we'll rebase and squash. Cool. :)

davidchambers:
Please rebase this branch when you get a chance.

Avaq:
:information_source: @davidchambers Notice the rebase? I'm not sure if it would've notified you.

test/index.js
```diff
@@ -5125,31 +5125,46 @@ describe('parse', function() {
...
-                     '             ^^^^^^\n' +
-                     '               1\n' +
+                     'parseJson :: TypeRep -> String -> Maybe a\n' +
+                     '             ^^^^^^^\n' +
+                     '                1\n' +
+                     '\n' +
+                     '1)  "String" :: String\n' +
+                     '\n' +
+                     'The value at position 1 is not a member of ‘TypeRep’.\n'));
+      throws(function() { S.parseJson(Array, [1, 2, 3]); },
```
davidchambers: Mind adding an empty line between these blocks?

Avaq: Done!

davidchambers:
:deciduous_tree: Thanks very much, @Avaq!

PR #171, opened 2016-04-04
sanctuary-def@0.5.x

### LINKED ISSUE #165: polymorphic functions cannot operate on "foreign" values

davidchambers:
Closes #165

/cc @Risto-Stevcev, @raqystyle, @soroushjp

> 1,582 additions, 216 deletions not shown because the diff is too large.

Since GitHub refuses to show the most exciting part of the patch—the improved error messages—I'll present a selection here.

#### Type-class constraint violation

Before:

```javascript
S.or(0, 1);
// ! TypeError: ‘or’ requires ‘a’ to implement Alternative; Number and FiniteNumber and Integer and ValidNumber do not
```

After:

```javascript
S.or(0, 1);
//
//   or :: Alternative a => a -> a -> a
//         ^^^^^^^^^^^^^    ^
//                          1
//
//   1)  0 :: Number, FiniteNumber, Integer, ValidNumber
//
//   ‘or’ requires ‘a’ to satisfy the Alternative type-class constraint; the value at position 1 does not.
```

#### Type-variable constraint violation

Before:

```javascript
S.or([], false);
// ! TypeError: ‘or’ expected a value of type (Array ???) or (List ???) as its second argument; received false
```

After:

```javascript
S.or([], false);
//
//   or :: Alternative a => a -> a -> a
//                          ^    ^
//                          1    2
//
//   1)  [] :: Array ???
//
//   2)  false :: Boolean
//
```

#### Invalid value

Before:

```javascript
S.isLeft([1, 2, 3]);
// ! TypeError: ‘isLeft’ expected a value of type (Either a b) as its first argument; received [1, 2, 3]
```

After:

```javascript
S.isLeft([1, 2, 3]);
// ! TypeError: Invalid value
//
//   isLeft :: Either a b -> Boolean
//             ^^^^^^^^^^
//                 1
//
//   1)  [1, 2, 3] :: Array Number, Array FiniteNumber, Array NonZeroFiniteNumber, Array Integer, Array ValidNumber
//
//   The value at position 1 is not a member of ‘Either a b’.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   },
   "dependencies": {
     "ramda": "0.21.x",
-    "sanctuary-def": "0.4.x"
+    "sanctuary-def": "0.5.x"
   },
   "devDependencies": {
     "doctest": "0.10.x",
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ README.md: index.js
 	  --heading-level 4 \
 	  --url 'https://github.com/plaid/sanctuary/blob/v$(VERSION)/{filename}#L{line}' \
 	  -- $^ \
-	| sed 's/<h4 name="\(.*\)#\(.*\)">\(.*\)\1#\2/<h4 name="\1.prototype.\2">\3\1#\2/' >'$@'
+	| LC_ALL=C sed 's/<h4 name="\(.*\)#\(.*\)">\(.*\)\1#\2/<h4 name="\1.prototype.\2">\3\1#\2/' >'$@'
 
 
 .PHONY: lint
--- a/bower.json
+++ b/bower.json
@@ -25,7 +25,7 @@
   ],
   "dependencies": {
     "ramda": "0.21.x",
-    "sanctuary-def": "0.4.x"
+    "sanctuary-def": "0.5.x"
   },
   "ignore": [
     "/.git/",
--- a/index.js
+++ b/index.js
@@ -111,7 +111,15 @@
 //.
 //. ```javascript
 //. S.inc('XXX');
-//. // ! TypeError: ‘inc’ expected a value of type FiniteNumber as its first argument; received "XXX"
+//. // ! TypeError: Invalid value
+//. //
+//. //   inc :: FiniteNumber -> FiniteNumber
+//. //          ^^^^^^^^^^^^
+//. //               1
+//. //
+//. //   1)  "XXX" :: String
+//. //
+//. //   The value at position 1 is not a member of ‘FiniteNumber’.
 //. ```
 //.
 //. Compare this to the behaviour of Ramda's unchecked equivalent:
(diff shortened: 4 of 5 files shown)


package.json
```diff
@@ -12,7 +12,7 @@
   },
   "dependencies": {
     "ramda": "0.19.x",
-    "sanctuary-def": "0.4.x"
+    "sanctuary-def": "plaid/sanctuary-def#dc-strmap"
```
davidchambers: TODO: Change to `"0.5.x"` once `sanctuary-def@0.5.0` is available.

PR #173, opened 2016-04-08
S.is() returned wrong result in IE 11

kahless:
In the current sanctuary version this code returns wrong result in IE 11:
`   S.is(Array, []) // It would be false`

The problem is in how sanctuary determines the type. More precisely how it compares user's constructor (Array) and the value from R.type() in 386th line:
`   R.type(x) === R.nth(1, R.match(/^function (\w*)/, String(type)))`

The actual problem is that this code returns different results in different browsers:
```
   String(Array)
   // In chrome — "function Array() { [native code] }"
   // In IE 11  — "\nfunction Array() {\n    [native code]\n}\n"
```

ECMA 262 5.1 says that the result of String(Array) is not determined and the count of whitespaces depends on implementation:

> 15.3.4.2 Function.prototype.toString ( )
> An implementation-dependent representation of the function is returned. This representation has the syntax of a FunctionDeclaration. Note in particular that the use and placement of white space, line terminators, and semicolons within the representation String is implementation-dependent.

That is why it is very unsafe to obtain constructor's name in this way. In future we can use a new (ES2015) 'Array.name' property: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Function/name But for now I don't think there is a way to make this regular expression safe. We can accept the string "array" instead of an Array to elude this problem.

### DIFF
--- a/index.js
+++ b/index.js
@@ -386,7 +386,7 @@
         return x != null && (
           R.type(type.prototype['@@type']) === 'String' ?
             x['@@type'] === type.prototype['@@type'] :
-            R.type(x) === R.nth(1, R.match(/^function (\w*)/, String(type)))
+            R.type(x) === R.nth(1, R.match(/function (\w*)/, String(type)))
         );
       });
 


davidchambers:
Thank you for submitting this patch, @kahless, and for writing a terrific description!

I know that @michaelficarra and others are [working][1] towards standardizing this behaviour, but it'll be some time before we can rely on that in our `is` implementation. ;)


[1]: https://github.com/tc39/Function-prototype-toString-revision

index.js
```diff
@@ -386,7 +386,7 @@
         return x != null && (
           R.type(type.prototype['@@type']) === 'String' ?
             x['@@type'] === type.prototype['@@type'] :
-            R.type(x) === R.nth(1, R.match(/^function (\w*)/, String(type)))
+            R.type(x) === R.nth(1, R.match(/function (\w*)/, String(type)))
```
davidchambers: Is there a reason to favour `/function (\w*)/` over `/^\s*function (\w*)/`, or vice versa?

kahless: This is a strange question.
`/function (\w*)/` matchs `"\nfunction Array() {\n    [native code]\n}\n"` in IE 11. And `/^function (\w*)/` don't. This is the only reason I made this change :)

davidchambers: I think you missed the `\s*` after the `^` in the second pattern. ;)

kahless: I thought about such pattern but insertion of "\s*" still does not make it safe. Because according to specification there can be any combination of spaces, newlines and semicolons. So it should be something like this: `/^[;\s]*function\s*(\w*)/`

davidchambers: Okay. I'm happy with either approach, in that case. :)

Did you see my comment about changing `S.if` to `S.is` in the commit message and pull request title?

kahless: Yes. I have not done this before but I think I will deal with it. I will force-push a new commit instead of this one. Then I will make a new pull request with the right title.

davidchambers: Actually, you can `git commit --amend` and tweak the commit message. Then `git push origin <branch> --force` to update this pull request. You can then update the pull request's title in the GitHub UI.

kahless: I see, thanks :) Replaced S.if() with S.is()

davidchambers:
:deciduous_tree: Thanks again!

PR #179, opened 2016-04-14
remove unnecessary types from environment

davidchambers:
Commit message:

> The environment is only necessary to support polymorphic functions. If the environment contains type A, it need not also contain type B if B is a subset of A.

I'm not convinced this is a good change. I look forward to hearing from @svozza and others.

### DIFF
--- a/index.js
+++ b/index.js
@@ -315,17 +315,7 @@
   );
 
   //  env :: [Type]
-  var env = $.env.concat([
-    $.FiniteNumber,
-    $.NonZeroFiniteNumber,
-    $Either,
-    $.Integer,
-    $Maybe,
-    $.RegexFlags,
-    TypeRep,
-    $.ValidDate,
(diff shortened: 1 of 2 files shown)


svozza:
What misgivings do you have?

davidchambers:
This is the change I find most concerning:

```diff
 Invalid value

 div :: FiniteNumber -> NonZeroFiniteNumber -> FiniteNumber
                        ^^^^^^^^^^^^^^^^^^^
                                 1

-1)  0 :: Number, FiniteNumber, Integer, ValidNumber
+1)  0 :: Number

 The value at position 1 is not a member of ‘NonZeroFiniteNumber’.
```

The argument is required to be a member of NonZeroFiniteNumber, but does not even appear to be a member of FiniteNumber in this case.

davidchambers:
I'm going to close this. It's not sitting right with me.

PR #180, opened 2016-04-14
array: add section for array-specific functions

davidchambers:
Commit message:

> This commit also updates type signatures to use `Array _` rather than `[_]` to represent the type of a value which really must be an array.

### DIFF
--- a/index.js
+++ b/index.js
@@ -83,11 +83,6 @@
 //. Sanctuary's run-time type checking asserts that a valid Number value is
 //. provided wherever an Integer value is required.
 //.
-//. ### List pseudotype
-//.
-//. The List pseudotype represents non-Function values with numeric `length`
-//. properties greater than or equal to zero, such as `[1, 2, 3]` and `'foo'`.
-//.
 //. ### Type representatives
 //.
 //. What is the type of `Number`? One answer is `a -> Number`, since it's a
@@ -314,7 +309,7 @@
     }
   );
 
-  //  env :: [Type]
+  //  env :: Array Type
   var env = $.env.concat([
     $.FiniteNumber,
     $.NonZeroFiniteNumber,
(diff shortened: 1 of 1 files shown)


Avaq:
Maybe it's worth adding a section near [`### List pseudotype`](https://github.com/plaid/sanctuary/blob/dc-array/index.js#L86-L89) about how the differences between Array and List are expressed throughout type signatures (`[_]` for `List`, `Array _` for Array). I had to figure out the difference by reading the source. :)

davidchambers:
Excellent suggestion, Aldwin! I will do so. :)

davidchambers:
Please have another look at this pull request, @Avaq. :)

Avaq:
I see you switched to using regular brackets to remove ambiguity in nested type signatures. I'm curious about your reasoning. Other than that it looks good to me. :)

davidchambers:
> I see you switched to using regular brackets to remove ambiguity in nested type signatures. I'm curious about your reasoning.

I don't follow, I'm sorry.

I did, for example, change `justs :: [Maybe a] -> [a]` to `justs :: Array (Maybe a) -> Array a`. I did so because this function really does require an *array* of maybes. It would reject, for example, an arguments object of maybes.

Avaq:
Oh, whoops! I must have misread. It seems you were already using regular brackets before. I was talking about how I would write: `Either[Error, Array String]` where I think you would write `Either Error (Array String)`. I somehow thought you changed between the two styles in *this* PR, but I was wrong.

davidchambers:
I like to stay as true to the Haskell notation as possible, so I use `Either a b` rather than `Either[a, b]`.

You're happy for this pull request to be merged otherwise?

Avaq:
I don't know where I got the square brackets from. I'll probably switch to using your haskellesk notation as well :). The changes look good to me by the way, feel free to merge.

Issue #163, opened 2016-03-05
Curry-safe reduce?


rjmk:
Is there space in sanctuary for a reduce-like function that calls the reducing function once on the accumulator and then the result on the current element? As it stands, Sanctuary, like in Ramda, doesn't work for manually curried functions.

davidchambers:
I like this proposal. Having higher-order functions operate on non-unary functions makes me uneasy. I'm open to the possibility of supporting both. For example:

```haskell
reduce  :: Foldable f => (a -> b -> a) -> a -> f b -> a
reduce_ :: Foldable f => ((a, b) -> a) -> a -> f b -> a
```

`reduce` would support manually curried functions, `reduce_` would support true binary functions, and both would support functions curried Ramda-style.

If there's support for this approach I suggest we apply it throughout the library.

rjmk:
> If there's support for this approach I suggest we apply it throughout the library.

I would really like to see this! When I'm working, it more often occurs to me to use sanctuary so I can declutter and write a better pointfree definition than for the type-safety (not that I care more about pointfree than type-safety, it's just more salient)

davidchambers:
Here are the higher-order functions in question:

```console
$ pcregrep '//# .* :: .*[(][^()]* -> [^()]* -> [^()]*[)]' index.js
  //# C :: (a -> b -> c) -> b -> a -> c
  //# S :: (a -> b -> c) -> (a -> b) -> a -> c
  //# flip :: (a -> b -> c) -> b -> a -> c
  //# lift2 :: Apply f => (a -> b -> c) -> f a -> f b -> f c
  //# lift3 :: Apply f => (a -> b -> c -> d) -> f a -> f b -> f c -> f d
  //# meld :: [** -> *] -> (* -> * -> ... -> *)
  //# Maybe#reduce :: Maybe a ~> (b -> a -> b) -> b -> b
  //# encase2 :: (a -> b -> c) -> a -> b -> Maybe c
  //# encase3 :: (a -> b -> c -> d) -> a -> b -> c -> Maybe d
  //# encaseEither2 :: (Error -> l) -> (a -> b -> r) -> a -> b -> Either l r
  //# encaseEither3 :: (Error -> l) -> (a -> b -> c -> r) -> a -> b -> c -> Either l r
  //# reduce :: Foldable f => (a -> b -> a) -> a -> f b -> a
```

rjmk:
[...]

PR #182, opened 2016-04-19
Changes higher order functions to expect curried functions

rjmk:
Opened this to restart the conversation in #163 

Does not currently expose additional functions that are uncurry-safe

### DIFF
--- a/index.js
+++ b/index.js
@@ -530,7 +530,7 @@
 
   //. ### Function
 
-  //# flip :: (a -> b -> c) -> b -> a -> c
+  //# flip :: ((a, b) -> c) -> b -> a -> c
   //.
   //. Takes a binary function and two values and returns the result of
   //. applying the function - with its argument order reversed - to the
@@ -1286,31 +1286,69 @@
   //# encase2 :: (a -> b -> c) -> a -> b -> Maybe c
   //.
   //. Binary version of [`encase`](#encase).
+  //.
+  //. See also [`encase2_`](#encase2_).
   S.encase2 =
   def('encase2',
       {},
       [$.Function, a, b, $Maybe(c)],
       function(f, x, y) {
         try {
-          return Just(f(x, y));
+          return Just(f(x)(y));
         } catch (err) {
           return Nothing();
         }
       });
 
+  //# encase2_ :: ((a, b) -> c) -> a -> b -> Maybe c
+  //.
+  //. Version of [`encase2`](#encase2) accepting uncurried functions.
+  S.encase2_ =
+  def('encase2_',
+      {},
+      [$.Function, a, b, $Maybe(c)],
+      function(f_, x, y) {
+        var f = function(x) {
+          return function(y) {
+            return f_(x, y);
+          };
+        };
+        return S.encase2(f, x, y);
+      });
+
   //# encase3 :: (a -> b -> c -> d) -> a -> b -> c -> Maybe d
   //.
   //. Ternary version of [`encase`](#encase).
+  //.
+  //. See also [`encase3_`](#encase3_).
   S.encase3 =
   def('encase3',
       {},
       [$.Function, a, b, c, $Maybe(d)],
       function(f, x, y, z) {
         try {
-          return Just(f(x, y, z));
+          return Just(f(x)(y)(z));
         } catch (err) {
           return Nothing();
         }
+      });
+
+  //# encase3_ :: ((a, b, c) -> d) -> a -> b -> c -> Maybe d
+  //.
+  //. Version of [`encase3`](#encase3) accepting uncurried functions.
+  S.encase3_ =
+  def('encase3_',
+      {},
+      [$.Function, a, b, c, $Maybe(d)],
+      function(f_, x, y, z) {
+        var f = function(x) {
+          return function(y) {
+            return function(z) {
+              return f_(x, y, z);
+            };
+          };
+        };
+        return S.encase3(f, x, y, z);
       });
 
   //. ### Either type
@@ -1771,31 +1809,70 @@
   //# encaseEither2 :: (Error -> l) -> (a -> b -> r) -> a -> b -> Either l r
   //.
   //. Binary version of [`encaseEither`](#encaseEither).
+  //.
+  //. See also [`encaseEither2_`](#encaseEither2_).
   S.encaseEither2 =
   def('encaseEither2',
       {},
       [$.Function, $.Function, a, b, $Either(l, r)],
       function(f, g, x, y) {
         try {
-          return Right(g(x, y));
+          return Right(g(x)(y));
         } catch (err) {
           return Left(f(err));
         }
       });
 
+  //# encaseEither2_ :: (Error -> l) -> ((a, b) -> r) -> a -> b -> Either l r
+  //.
+  //. Version of [`encaseEither2`](#encaseEither2) accepting uncurried
+  //. functions.
+  S.encaseEither2_ =
+  def('encaseEither2_',
+      {},
+      [$.Function, $.Function, a, b, $Either(l, r)],
+      function(f, g_, x, y) {
+        var g = function(x) {
+          return function(y) {
+            return g_(x, y);
+          };
+        };
+        return S.encaseEither2(f, g, x, y);
+      });
+
   //# encaseEither3 :: (Error -> l) -> (a -> b -> c -> r) -> a -> b -> c -> Either l r
   //.
   //. Ternary version of [`encaseEither`](#encaseEither).
+  //.
+  //. See also [`encaseEither3_`](#encaseEither3_).
   S.encaseEither3 =
   def('encaseEither3',
       {},
       [$.Function, $.Function, a, b, c, $Either(l, r)],
       function(f, g, x, y, z) {
         try {
-          return Right(g(x, y, z));
+          return Right(g(x)(y)(z));
         } catch (err) {
           return Left(f(err));
         }
+      });
+
+  //# encaseEither3_ :: (Error -> l) -> ((a, b, c) -> r) -> a -> b -> c -> Either l r
+  //.
+  //. Version of [`encaseEither3`](#encaseEither3) accepting uncurried functions.
+  S.encaseEither3_ =
+  def('encaseEither3',
+      {},
+      [$.Function, $.Function, a, b, c, $Either(l, r)],
+      function(f, g_, x, y, z) {
+        var g = function(x) {
+          return function(y) {
+            return function(z) {
+              return g_(x, y, z);
+            };
+          };
+        };
+        return S.encaseEither3(f, g, x, y, z);
       });
 
   //# maybeToEither :: a -> Maybe b -> Either a b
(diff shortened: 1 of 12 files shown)


davidchambers:
Thanks for the pull request, @rjmk!

I think we'll want variants of many of these functions as suggested in #163. We could provide both

```haskell
reduce :: Foldable f => (a -> b -> a) -> a -> f b -> a
```

and

```haskell
reduce_ :: Foldable f => ((a, b) -> a) -> a -> f b -> a
```

for example, and define the latter in terms of the former.

Do others agree this is a good idea?

index.js
```diff
@@ -535,7 +535,7 @@
 
   //. ### Function
 
-  //# flip :: (a -> b -> c) -> b -> a -> c
+  //# flip :: ((a -> b) -> c) -> b -> a -> c
```
davidchambers: I think this should be `(a, b) -> c` rather than `(a -> b) -> c`.

index.js
```diff
@@ -2360,8 +2360,8 @@
 
   //# reduce :: Foldable f => (a -> b -> a) -> a -> f b -> a
   //.
-  //. Takes a binary function, an initial value, and a [Foldable][], and
-  //. applies the function to the initial value and the Foldable's first
+  //. Takes a  curried binary function, an initial value, and a [Foldable][],
```
davidchambers: `s/a /a/`

test/index.js
```diff
@@ -73,6 +79,12 @@ var squareRoot = function(n) {
                : S.Right(Math.sqrt(n));
 };
 
+//  highArity :: a -> b -> c -> d -> a
+var highArity = function(a, b, c, d) { return a; };
+
+//  thunk :: (a -> b) -> c -> a -> b
+var thunk = function(fn) { return function() { return fn; }; };
```
davidchambers: The type signature can be more general:

```javascript
//  thunk :: a -> b -> a
var thunk = function(x) { return function(y) { return x; }; };
```

In fact, this is just [`S.K`][1], is it not?


[1]: http://sanctuary.js.org/#K

test/index.js
```diff
@@ -73,6 +79,12 @@ var squareRoot = function(n) {
                : S.Right(Math.sqrt(n));
 };
 
+//  highArity :: a -> b -> c -> d -> a
+var highArity = function(a, b, c, d) { return a; };
```
davidchambers: Let's represent the type as `(a, b, c, d) -> a`.

test/index.js
rjmk: I wondered whether it was better no to use `K` within the test suite. Could always use `R.const` if so. Happy to use `K` if no concerns

index.js
rjmk: Oops!

rjmk:
@davidchambers I do view the `reduce_`s etc. as a bit crufty. `reduce`, for example, will have its counterpart in ramda and arguments to `encase`s can always be uncurried

rjmk:
What would be the best approach to exposing things like `encase2_`. A function like

```js
function(f, x, y) {
  return encase2(R.uncurryN(2, f), x, y);
}
```
?

davidchambers:
Thanks for the updates, @rjmk.

I imagine the implementation of `encase2_` looking something like this:

```javascript
function(f_, x, y) {
  var f = function(x) {
    return function(y) {
      return f_(x, y);
    };
  };
  return encase2(f, x, y);
}
```

I'd like to curry the function manually as one of my long-term goals is to remove the Ramda dependency.

rjmk:
My pleasure, @davidchambers.

Would you rather have each done by hand, or is an internal `_uncurry` function (or dep) acceptable?

davidchambers:
> Would you rather have each done by hand, or is an internal `_uncurry` function (or dep) acceptable?

To minimize indirection in stack traces I'd like to curry these functions manually. Since the `_` functions are inherently inelegant, I don't mind inelegant implementations. ;)

rjmk:
Ha ha!

OK, sounds reasonable. I'll get to it

rjmk:
Also, tests? Coverage testing doesn't complain as we have full branch coverage. If we go for tests, is a Ramda devDependency OK? Just to avoid some duplication on util functions

davidchambers:
> If we go for tests, is a Ramda devDependency OK?

Absolutely. :)

rjmk:
All done. A question has occurred to me though: where do we want the responsibility to lie for the expected arity when dispatching to `reduce`? Does `reduce_` need to have an escape for dispatching as well?

davidchambers:
I don't quite follow, @rjmk. If `reduce_` is defined in terms of `reduce`, it will dispatch on its third argument as `reduce` does. I feel I'm missing something.

rjmk:
My thought is that we don't want users of the library to have to keep track of whether `reduce(a => b => a + b)` will be called on an array, or a type with a reduce implementation that expects its reducing function to be called as `f(a, b)` or as `f(a)(b)`

One option would be to Ramda-style curry the reducing function, to hide that implementation detail

davidchambers:
I see your point now. Thanks for the explanation. :)

We *could* change…

```javascript
if (_type(foldable) === 'Array') {
  // ...
} else {
  return foldable.reduce(f, initial);
}
```

to something like…

```javascript
if (_type(foldable) === 'Array') {
  // ...
} else if (foldable.reduce.length === 2) {
  return foldable.reduce(f, initial);
} else {
  return foldable.reduce(f)(initial);
}
```

I don't think this is necessary, though. According to the [spec][1]:

> A value which has a Foldable must provide a `reduce` method. The `reduce` method takes two arguments:
>
>     u.reduce(f, x)


rjmk:
I don't think that's quite the concern, @davidchambers.

The issue is that `f` might need to be binary, not `#reduce`. I personally would like a library like sanctuary to help me use a manually curried `f`. So perhaps in the case of `reduce`, unlike all the others, we should have `reduce` dispatch to `reduce_` rather than the other way around

davidchambers:
> I personally would like a library like sanctuary to help me use a manually curried `f`.

After three attempts I believe I've grasped this. We'd like to ensure that `S.reduce(a => b => a + b, 0)` can be applied to any Foldable, but according to the spec "`f` must be a binary function" (I forgot to include this line in the quote earlier). Have I understood correctly?

It seems we could wrap the `a -> b -> a` function to provide a binary function to the `reduce` method:

```javascript
function(f, initial, foldable) {
  if (_type(foldable) === 'Array') {
    // ...
  } else {
    return foldable.reduce(function(acc, x) { return f(acc)(x); }, initial);
  }
}
```

rjmk:
Yep! The only outstanding question on my end is whether it's better to do as you suggest, and then in `reduce_` wrap the binary function in a function that allows it be partially applied only to rewrap it to regain the original functionality, or whether to just have `reduce_` be the place where the main function definition takes place, and then define `reduce` in terms of `reduce_`

rjmk:
Also, I think I should revert `Maybe#reduce` as well, so that the datatype is still to spec.

davidchambers:
Defining `S.reduce` in terms of `S.reduce_` sounds fine to me. I'll defer to you, though, as you're at least three steps ahead of me on this one. ;)

rjmk:
I kept the tests for the `_` functions fairly brief, as I thought it would be better to not have too much duplication in tests that would have to be chased down in face of future changes

index.js
```diff
@@ -1292,12 +1292,28 @@
...
       function(f, x, y) {
         try {
-          return Just(f(x, y));
+          return Just(f(x)(y));
         } catch (err) {
           return Nothing();
         }
       });
 
+  //# encase2__ :: ((a, b) -> c) -> a -> b -> Maybe c
```
davidchambers: `s/__/_/`

index.js
```diff
@@ -1777,12 +1811,28 @@
...
-          return Right(g(x, y));
+          return Right(g(x)(y));
         } catch (err) {
           return Left(f(err));
         }
       });
 
+  //# encaseEither2_ :: (Error -> l) -> ((a, b) -> r) -> a -> b -> Either l r
+  //.
+  //. Version of [`encaseEither2`](#encaseEither2) accepting uncurried functions.
```
davidchambers: Let's wrap at 79 characters. :)

index.js
```diff
@@ -2404,13 +2472,27 @@
...
   //. ```
   S.reduce =
   def('reduce',
       {b: [Foldable]},
       [$.Function, a, b, a],
+      function(f_, initial, foldable) {
+        var f = function(a, b) {
+          return f_(a)(b);
+        };
+        return S.reduce_(f, initial, foldable);
```
davidchambers: Let's reference this as `reduce_` (by introducing an alias) rather than as `S.reduce_`.

test/encase2_.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+var uncurry2 = require('ramda').uncurryN(2);
+
+var eq = require('./utils').eq;
+var rem = uncurry2(require('./utils').rem);
+var S = require('..');
```
davidchambers: I'd prefer not to intermingle imports and other code. I suggest:

```javascript
'use strict';

var R = require('ramda');

var S = require('..');
var utils = require('./utils');


var eq = utils.eq;
var rem = R.uncurryN(2, utils.rem);
```

I realize you're simply doing as we do in other files, though, so you needn't change this.

test/reduce.js
```diff
@@ -6,6 +6,12 @@ var eq = require('./utils').eq;
 var errorEq = require('./utils').errorEq;
 var S = require('..');
 
+var unary_add = function(a) { return function(b) { return a + b; }; };
```
davidchambers: Camel case, please. :)

test/utils.js
```diff
@@ -8,13 +8,17 @@ var S = require('../index.js');
 
 
 //      area :: (Number, Number, Number) -> Number !
```
davidchambers: Let's update this signature.

test/utils.js
```diff
@@ -45,12 +49,20 @@ exports.parseHex = function(s) {
 };
 
 //      rem :: (Number, Number) -> Number !
```
davidchambers: Let's update this signature.

test/utils.js
```diff
@@ -45,12 +49,20 @@ exports.parseHex = function(s) {
...
+  return function(y) {
+    if (y === 0) {
+      throw new Error('Cannot divide by zero');
+    } else {
+      return x % y;
+    }
+  };
+};
+
+exports.highArity = function(a) {
```
davidchambers: Let's include a signature: `a -> ((b, c, d) -> b)`.

rjmk:
Thanks for the through review, @davidchambers! 👍

test/encase2_.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+var R = require('ramda');
```
davidchambers: Please include an empty line after the `'use strict'` pragma.

davidchambers:
Looking good, @rjmk! Thank you for all your work on this. There are two more things I'd like you to do:

  - add "see also" links from each unsuffixed function to its corresponding suffixed function (e.g. from `reduce` to `reduce_`) in the style used elsewhere; and
  - `git rebase master` and squash the commits.

rjmk:
Awesome! Just to clarify, shall I squash all the commits and force push onto my branch?

davidchambers:
> Just to clarify, shall I squash all the commits and force push onto my branch?

That would be wonderful. GitHub has a fancy new auto-squashing feature, but I prefer to let the author squash the commits and adjust the combined commit message as he or she sees fit.

davidchambers:
:deciduous_tree: Thanks so much for this contribution, @rjmk.

rjmk:
@davidchambers My pleasure. Will there be a blog post on the next version bump?

davidchambers:
> Will there be a blog post on the next version bump?

No.

How do you like to be notified of new releases? GitHub? Gitter? Twitter? Email? Something else? We haven't done much to announce releases in the past, but I'm happy to do so in future if there's interest.

rjmk:
Probably the order for me is twitter -> github (I have only just started watching the repo) -> gitter (easy to lose the significance)

davidchambers:
Okay. I'll create Ramda-style [upgrade guides][1] from this point forward.

I created a Sanctuary account on Twitter. I'm open to using it in an automated fashion to announce releases, though I'm reluctant to add another communication channel. ;)


[1]: https://github.com/ramda/ramda/issues?q=label%3A%22upgrade+guide%22

PR #183, opened 2016-04-19
Replace jshint/jscs with eslint

gilligan:
#### what is it ?
This replaces jshint and jscs with eslint. 

#### why ?
* Ramda also [switched to eslint](https://github.com/ramda/ramda/pull/1529) some time ago. So far nobody complained ;)
* jscs will soon have its last release as they are [merging with eslint](http://eslint.org/blog/2016/04/welcoming-jscs-to-eslint)
* jshint is not very active and not even anywhere close to the abilities of eslint

#### how ?
The initial .eslintrc was created using [polyjuice](https://github.com/brenolf/polyjuice). The only change I did in the code was to add some missing indentation level. I suppose there are quite some eslint settings that could be added over time. We could even use the same config as Ramda if that is desirable? I also cannot guarantee that the tool I used did indeed pick up everything or does cover everything but the code is accepted as it is and the settings seem to make sense to me. All further adjustments could be made over time I suppose.

Looking forward to hear what folks think about this ;-)


gilligan:
Removed the added indentation in `index.js` and disabled the indent config in `.eslintrc`

davidchambers:
I was unaware of the ESLint–JSCS merge. Thanks for informing me of this, and for submitting this pull request. :)

.eslintrc
```diff
@@ -0,0 +1,62 @@
...
+    "no-bitwise": 0,
+    "camelcase": 0,
+    "curly": 0,
+    "eqeqeq": 0,
+    "guard-for-in": 0,
+    "no-extend-native": 0,
+    "wrap-iife": [
+        2,
+      "any"
+    ],
```
davidchambers: I'd prefer `"wrap-iife": [2, "any"]` (on one line).

.eslintrc
```diff
@@ -0,0 +1,62 @@
+{
+  "env": {
+    "node": true
```
davidchambers: What does this mean? We need to be compatible with browsers as well.

Makefile
```diff
@@ -20,8 +19,7 @@ README.md: index.js
 
 .PHONY: lint
 lint:
-	$(JSHINT) -- index.js test/index.js
-	$(JSCS) -- index.js test/index.js
+	$(ESLINT) -- index.js test/index.js index.js
```
davidchambers: Did you intend to list `index.js` twice?

.eslintrc
```diff
@@ -0,0 +1,62 @@
...
+        99
+    ],
+    "max-depth": [
+        2,
+        99
+    ],
+    "complexity": [
+        2,
+        99
+    ],
```
davidchambers: Let's remove `max-params`, `max-depth`, and `complexity`. I've used arbitrarily large values for such options in the past and exceeded the limits (in test suites with lots of assertions).

.eslintrc
Avaq: This is like a template for `"globals"`, automatically populating it with the globals expected to be in Node. It's probably why the linter isn't complaining about the [use of `module.exports`](https://github.com/plaid/sanctuary/blob/master/index.js#L161).

gilligan: @Avaq yep exactly. It just defines nodejs globals.

davidchambers: Thanks for the explanation. :)

Makefile
```diff
@@ -1,7 +1,6 @@
 DOCTEST = node_modules/.bin/doctest --nodejs '--harmony' --module commonjs --prefix .
 ISTANBUL = node_modules/.bin/istanbul
-JSCS = node_modules/.bin/jscs
-JSHINT = node_modules/.bin/jshint
+ESLINT = node_modules/.bin/eslint
```
davidchambers: :abc:

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+  },
+  "globals": {},
+  "rules": {
+    "no-bitwise": 0,
+    "camelcase": 0,
+    "curly": 0,
+    "eqeqeq": 0,
+    "guard-for-in": 0,
+    "no-extend-native": 0,
+    "wrap-iife": [ 2, "any" ],
```
davidchambers: No spaces inside the `[` and `]`, please.

Who lints the linter? ;)

.eslintrc
```diff
@@ -0,0 +1,44 @@
+{
+  "env": {
+    "node": true
+  },
+  "globals": {},
+  "rules": {
+    "no-bitwise": 0,
```
davidchambers: Let's make this `2`, unless we're already using bitwise operators.

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+    "camelcase": 0,
```
davidchambers: Is there any reason to mention rules we are not enabling? It seems cleaner to simply omit them.

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+    "curly": 0,
+    "eqeqeq": 0,
```
davidchambers: Let's make this `[2, "allow-null"]`.

.eslintrc
davidchambers: Let's make this `[2, "outside"]`.

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+    "guard-for-in": 0,
+    "no-extend-native": 0,
+    "wrap-iife": [ 2, "any" ],
+    "no-use-before-define": 0,
+    "new-cap": 0,
+    "no-caller": 2,
+    "no-empty": 2,
```
davidchambers: Let's not enable this rule. There are valid use cases for `for (…; …; …) {}`.

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+    "no-new": 2,
+    "no-plusplus": 2,
+    "no-undef": 2,
+    "no-unused-vars": [2, {"vars": "all", "args": "none"}],
+    "strict": 0,
```
davidchambers: Could we enable this rule?

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+    "semi": 0,
```
davidchambers: Let's make this `2`.

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+    "no-cond-assign": 0,
```
davidchambers: Let's make this `[2, "except-parens"]`.

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+    "no-debugger": 0,
```
davidchambers: Let's make this `2`.

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+    "no-eq-null": 0,
+    "no-eval": 0,
```
davidchambers: Let's make this `2`.

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+    "no-unused-expressions": 0,
```
davidchambers: Let's make this `2`.

.eslintrc
gilligan: Hehe.. at work we actually are explicit about all the rules we configure - have a look at https://github.com/holidaycheck/eslint-config-holidaycheck/blob/master/baseConfig.json

I suppose the upside to that is that you can look at the configuration and actually see what the rules for the respective project are. Instead of writing some code to see if it lints. Of course that only applies if either the rule is obvious or you have some prior knowledge of the eslint rules.

I don't mind removing the disabled ones if you prefer that. TBH that is the result of that converting tool. I am not entirely sure why it picked the ones that are included and disabled. I think it does not make to explicitly disable *some* rules so I'll go ahead and just remove all disabled rules.

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+    "block-scoped-var": 0,
```
davidchambers: Could we enable this rule?

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+    "no-iterator": 0,
+    "linebreak-style": 0,
```
davidchambers: Let's make this `[2, "unix"]`.

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+    "comma-style": [ 2, "last" ],
+    "no-loop-func": 0,
```
davidchambers: Let's make this `2`.

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+    "no-multi-str": 0,
```
davidchambers: Let's make this `2`.

gilligan:
@davidchambers well the config is pretty tiny now ;-} I do think it would sense to add more rules over time. Also I think that pretty much all rules from the [possible errors](http://eslint.org/docs/rules/#possible-errors) set should be just fine. Because I'm pretty sure nobody wants unnecessary semicolons, invalid regexes or the other stuff that is just plain "wrong".

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+    "valid-typeof": 0,
```
davidchambers: Let's make this `2`.

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+    "no-proto": 0,
+    "no-script-url": 0,
+    "no-shadow": 0,
+    "dot-notation": 2,
```
davidchambers: Let's make this `[2, {"allowKeywords": false}]`.

.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+    "no-new-func": 0,
+    "no-new-wrappers": 0,
+    "no-invalid-this": 0
```
davidchambers: Let's make this `2`.

davidchambers:
I'd like to close this pull request and create a separate repository for our ESLint configuration, to allow it to be shared by the various Sanctuary projects. Does that sound okay to you, @gilligan?

gilligan:
@davidchambers That sounds perfect actually!

davidchambers:
sanctuary-js/sanctuary-style#1

PR #184, opened 2016-04-19
Add S.append

gilligan:
This adds S.append as suggested by #145 

So this is my first contribution and I just naively implemented it as `a -> [a] -> [a]`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2344,6 +2344,24 @@
 
   //. ### Array
 
+  //# append :: a -> Array a -> Array a
+  //.
+  //. Takes a value of any type and an array of values of that type, and
+  //. returns the result of appending the value to the array.
+  //.
+  //. ```javascript
+  //. > S.append(3, [1, 2])
+  //. [1, 2, 3]
+  //.
+  //. > S.append([3, 4], [[1], [2]])
+  //. [[1], [2], [3, 4]]
+  //. ```
+  S.append =
+  def('append',
+      {},
+      [a, $.Array(a), $.Array(a)],
+      function(x, xs) { return xs.concat([x]); });
+
   //# find :: (a -> Boolean) -> Array a -> Maybe a
   //.
   //. Takes a predicate and an array and returns Just the leftmost element of
--- a/test/append.js
+++ b/test/append.js
@@ -1,0 +1,64 @@
+'use strict';
+
+var throws = require('assert').throws;
+
+var errorEq = require('./utils').errorEq;
+var eq = require('./utils').eq;
+var S = require('..');
+
+
+describe('append', function() {
+
+   it('is a binary function', function() {
+     eq(typeof S.append, 'function');
+     eq(S.append.length, 2);
+   });
+
+   it('type checks its arguments', function() {
+     throws(function() { S.append('c', 'ab'); },
+            errorEq(TypeError,
+                    'Invalid value\n' +
+                    '\n' +
+                    'append :: a -> Array a -> Array a\n' +
+                    '               ^^^^^^^\n' +
+                    '                  1\n' +
+                    '\n' +
+                    '1)  "ab" :: String\n' +
+                    '\n' +
+                    'The value at position 1 is not a member of ‘Array a’.\n'));
+
+     throws(function() { S.append('c', [1, 2]); },
+            errorEq(TypeError,
+                    'Type-variable constraint violation\n' +
+                    '\n' +
+                    'append :: a -> Array a -> Array a\n' +
+                    '          ^          ^\n' +
+                    '          1          2\n' +
+                    '\n' +
+                    '1)  "c" :: String\n' +
+                    '\n' +
+                    '2)  1 :: Number, FiniteNumber, NonZeroFiniteNumber, Integer, ValidNumber\n' +
+                    '    2 :: Number, FiniteNumber, NonZeroFiniteNumber, Integer, ValidNumber\n' +
+                    '\n' +
+                    'Since there is no type of which all the above values are members, the type-variable constraint has been violated.\n'));
+   });
+
+  it('adds the element to the end of the list', function() {
+    eq(S.append('c', ['a', 'b']), ['a', 'b', 'c']);
+    eq(S.append({x: 3}, [{x: 1}, {x: 2}]), [{x: 1}, {x: 2}, {x: 3}]);
+  });
+
+  it('adds a list to a list of lists', function() {
+    eq(S.append([3, 4], [[1], [2]]), [[1], [2], [3, 4]]);
+  });
+
+  it('works on empty list', function() {
+    eq(S.append(1, []), [1]);
+  });
+
+  it('is curried', function() {
+    eq(S.append('c').length, 1);
+    eq(S.append('c')(['a', 'b']), ['a', 'b', 'c']);
+  });
+
+});


svozza:
We could restrict `a` to `Semigroup`, perhaps?

davidchambers:
Thanks, @gilligan!

Let's move the code and tests to the [List][1] section. It'd feel at home between [`dropLast`][2] and [`find`][3], I'd say.


[1]: http://sanctuary.js.org/#list
[2]: http://sanctuary.js.org/#dropLast
[3]: http://sanctuary.js.org/#find

index.js
```diff
@@ -3032,6 +3032,22 @@
...
+  //.
+  //. ```javascript
+  //. > S.append(3, [1, 2])
+  //. [1, 2, 3]
+  //. ```
+  S.append =
+  def('append',
+      {},
+      [a, $.Array(a), $.Array(a)],
+      function(x, xs) { return xs.concat(x); });
```
kedashoe: I think we need to wrap `x` here, currently we have `S.append([3, 4], [[1], [2]]); //=> [[1], [2], 3, 4]`

davidchambers: Let's use `xs.concat([x])`, to handle nested arrays correctly:

```javascript
> [1, 2].concat(3)
[1, 2, 3]

> [1, 2].concat([3])
[1, 2, 3]

> [[1], [2]].concat([3])
[[1], [2], 3]

> [[1], [2]].concat([[3]])
[[1], [2], [3]]
```

Let's also include a test case to cover this.

davidchambers: Beaten to the punch! Good spotting, Kevin. :)

gilligan:
@svozza I guess [semigroup](https://github.com/fantasyland/fantasy-land#semigroup) could make sense yes.
@davidchambers sure, will update the PR tomorrow.

davidchambers:
> We could restrict `a` to `Semigroup`, perhaps?

I don't think so, Stefano. The function should work `forall a`. The container must support `concat`, of course, but that's ensured by our use of `$.Array`.

gilligan:
@davidchambers  right..  so `S.append(() => {}, [])` should of course work as well.

svozza:
> I don't think so, Stefano. The function should work forall a. The container must support concat, of course, but that's ensured by our use of $.Array.

Do we not want this to work on strings too?

davidchambers:
> Do we not want this to work on strings too?

I'm not sure this is necessary: `S.concat('foo', '!')` behaves as `S.append('!', 'foo')` presumably would. One thing that worries me about supporting strings here is that `S.append('!')` would do different things depending on the type of the second argument:

```javascript
> S.append('!', 'abc')
'abc!'

> S.append('!', ['a', 'b', 'c'])
['a', 'b', 'c', '!']
```

There's also the question of whether `S.append('bar', 'foo')` is a type error, since it'd append more than one character to `'foo'`. Deferring to `S.concat` is a reasonable solution to these conundrums.

gilligan:
@davidchambers 
- [x] changed to `concat([x])`
- [x] added unit test / doc test to append list to list of lists
- [x] moved implementation list section
- [x] moved test to list section

anything else?

index.js
```diff
@@ -3032,6 +3065,8 @@
       [$.Array($.String), $.String],
       compose(R.join(''), R.map(concat(_, '\n'))));
 
+
+
```

index.js
```diff
@@ -1983,6 +1983,22 @@
...
+  //. > S.append(3, [1, 2])
+  //. [1, 2, 3]
+  //. > S.append([3, 4], [[1], [2]])
+  //. [[1], [2], [3, 4]]
+  //. ```
+  S.append =
+  def('append',
+      {},
+      [a, $.Array(a), $.Array(a)],
+      function(x, xs) { return xs.concat([x]); });
```
davidchambers: Let's delete this duplicate definition.

index.js
```diff
@@ -2245,6 +2261,23 @@
...
 
+  //# append :: a -> [a] -> [a]
+  //.
+  //. Takes a value and a list of values of the same type and appends
+  //. the value to the list.
+  //.
+  //. ```javascript
+  //. > S.append(3, [1, 2])
+  //. [1, 2, 3]
+  //. > S.append([3, 4], [[1], [2]])
```
davidchambers: Please include an "empty" line between doctests, as we do elsewhere.

index.js
```diff
@@ -2245,6 +2261,23 @@
         return n < 0 || negativeZero(n) ? Nothing() : slice(0, -n, xs);
       });
 
+  //# append :: a -> [a] -> [a]
+  //.
+  //. Takes a value and a list of values of the same type and appends
+  //. the value to the list.
```
davidchambers: Let's mention that the function *returns* a list to reduce the likelihood of a user assuming this modifies the given array. How about the following?

_Returns the result of appending the given value to the given list._

This is less precise than many of the other descriptions, but @joneshf has rightly called me out for writing descriptions which simply restate their corresponding type signatures. For example:

_Takes a value of any type and a list of values of that type, and returns the result of appending the value to the list._

test/index.js
```diff
@@ -3823,6 +3823,23 @@ describe('list', function() {
...
+      it('adds a list to a list of lists', function() {
+          eq(S.append([3, 4], [[1], [2]]), [[1], [2], [3, 4]]);
+      });
+
+      it('works on empty list', function() {
+        eq(S.append(1, []), [1]);
+      });
+
+  });
+
```
davidchambers: The indentation here is inconsistent. The body of the `it` blocks are indented by 3, 4, and 2 spaces respectively. The body of the `describe` block is indented by 4 spaces. Let's use 2 spaces in all cases.

This makes me think checking indentation in #183 is actually a good idea. We could enable it for __test/index.js__ but not for __index.js__.

test/index.js
```diff
@@ -3823,6 +3823,23 @@ describe('list', function() {
 
   });
 
+  describe('append', function() {
+
+      it('adds the element to the end of the list', function() {
+         eq(S.append('c', ['a', 'b']), ['a', 'b', 'c']);
+         eq(S.append({x:3}, [{x:1}, {x:2}]), [{x:1}, {x:2}, {x:3}]);
```
davidchambers: Let's use `{k: v}` rather than `{k:v}`. Can we enforce this via ESLint?

Avaq: Yes: [`"key-spacing": [2, {"beforeColon": false, "afterColon": true}]`](http://eslint.org/docs/rules/key-spacing) :)

davidchambers: Excellent!

index.js
gilligan: github seems to be drunk. This is not the correct contents of that file. The duplicate entry of `S.append` is gone. The only one is [here](https://github.com/plaid/sanctuary/pull/184/files#diff-168726dbe96b3ce427e7fedce31bb0bcR2277)

davidchambers: I no longer see a duplicate. :)

index.js
```diff
@@ -2274,6 +2274,24 @@
         return n < 0 || negativeZero(n) ? Nothing() : slice(0, -n, xs);
       });
 
+  //# append :: a -> [a] -> [a]
+  //.
+  //. Takes a value of any type and a list of values of that type, and
+  //. returns the result of appending the value to the list.
```
davidchambers: In light of #180, we need to make a few changes to preserve consistency:

  - change the signature to `a -> Array a -> Array a`;
  - `s/a list/an array/`;
  - `s/the list/the array/`; and
  - move the function to the Array section (at the very beginning, let's say).

Much has changed in the few days since you submitted this pull request. I realize I've already asked you to move this function once! There won't usually be this much friction, I promise. :)

davidchambers:
It looks as though your tests disappeared in the rebase!

gilligan:
@davidchambers ugh.. tests added again. This whole PR took much longer than I would have anticipated ;)

index.js
```diff
@@ -2274,6 +2274,7 @@
         return n < 0 || negativeZero(n) ? Nothing() : slice(0, -n, xs);
       });
 
+
```
Avaq: Is this newline intentional?

test/append.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+var eq = require('./utils').eq;
+var S = require('..');
+
+describe('append', function() {
```
Avaq: I believe the convention held within this project is to add two newlines after the last import statement.

gilligan: That is something that should/could be added to a sanctuary eslint config. I'll note that for later.

Avaq:
Looks good to me, though I haven't kept up with the discussions leading up to this point. I'll leave the final call to @davidchambers.

index.js
```diff
@@ -2292,6 +2292,24 @@
                R.pipe(R[name], Just, R.filter(R.gte(_, 0))));
   };
 
+  //# append :: a -> Array a -> Array a
```
davidchambers: This is still in the List section. This function is Array-specific so belongs in the Array section. Let's do this:

```diff
   //. ### Array

+  //# append :: a -> Array a -> Array a
+  //.
+  //. Takes a value of any type and an array of values of that type, and
+  //. returns the result of appending the value to the array.
+  //.
+  //. ```javascript
+  //. > S.append(3, [1, 2])
+  //. [1, 2, 3]
+  //.
+  //. > S.append([3, 4], [[1], [2]])
+  //. [[1], [2], [3, 4]]
+  //. ```
+  S.append =
+  def('append',
+      {},
+      [a, $.Array(a), $.Array(a)],
+      function(x, xs) { return xs.concat([x]); });
+
   //# find :: (a -> Boolean) -> Array a -> Maybe a
```

test/append.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+var eq = require('./utils').eq;
+var S = require('..');
+
+
+describe('append', function() {
+
+  it('adds the element to the end of the list', function() {
```
davidchambers: Currently the `[a, $.Array(a), $.Array(a)]` portion of the definition is untested. Let's add this:

```diff
 describe('append', function() {

+  it('is a binary function', function() {
+    eq(typeof S.append, 'function');
+    eq(S.append.length, 2);
+  });
+
+  it('type checks its arguments', function() {
+    throws(function() { S.append('c', 'ab'); },
+           errorEq(TypeError,
+                   'Invalid value\n' +
+                   '\n' +
+                   'append :: a -> Array a -> Array a\n' +
+                   '               ^^^^^^^\n' +
+                   '                  1\n' +
+                   '\n' +
+                   '1)  "ab" :: String\n' +
+                   '\n' +
+                   'The value at position 1 is not a member of ‘Array a’.\n'));
+
+    throws(function() { S.append('c', [1, 2]); },
+           errorEq(TypeError,
+                   'Type-variable constraint violation\n' +
+                   '\n' +
+                   'append :: a -> Array a -> Array a\n' +
+                   '          ^          ^\n' +
+                   '          1          2\n' +
+                   '\n' +
+                   '1)  "c" :: String\n' +
+                   '\n' +
+                   '2)  1 :: Number, FiniteNumber, NonZeroFiniteNumber, Integer, ValidNumber\n' +
+                   '    2 :: Number, FiniteNumber, NonZeroFiniteNumber, Integer, ValidNumber\n' +
+                   '\n' +
+                   'Since there is no type of which all the above values are members, the type-variable constraint has been violated.\n'));
+  });
+
   it('adds the element to the end of the list', function() {
     eq(S.append('c', ['a', 'b']), ['a', 'b', 'c']);
```

test/append.js
```diff
@@ -0,0 +1,22 @@
...
+
+  it('adds a list to a list of lists', function() {
+    eq(S.append([3, 4], [[1], [2]]), [[1], [2], [3, 4]]);
+  });
+
+  it('works on empty list', function() {
+    eq(S.append(1, []), [1]);
+  });
+
+});
```
davidchambers: It would be nice to assert that the arguments may be provided one at a time:

```javascript
it('is curried', function() {
  eq(S.append('c').length, 1);
  eq(S.append('c')(['a', 'b']), ['a', 'b', 'c']);
});
```

index.js
gilligan: wow i'm close to giving up x__X  - The one huge file is driving me bonkers. I don't think I've ever needed so many attempts to get something so trivial done. Sorry.

davidchambers: Even with separate files we'd need to specify the readme order *somewhere*, but a list of filenames in the makefile would certainly be easier to read. ;)

gilligan:
Are we good to go now? :-) *fingers crossed*

index.js
```diff
@@ -2292,6 +2292,7 @@
                R.pipe(R[name], Just, R.filter(R.gte(_, 0))));
   };
 
+
```

davidchambers:
Delete the stray line and I'll merge. :)

davidchambers:
:deciduous_tree: Thanks, Tobias!

PR #185, opened 2016-04-20
Split up test file

svozza:
I'm not going to get this finished tonight but I figure that we may as well start the review now given there's so much in here. I've only got the List, String and Number tests left to do. I'll change the name of the issue when I'm finished too.


davidchambers:
Thanks! This'll be a nice change. :)

test/A.js
```diff
@@ -0,0 +1,23 @@
...
+    eq(S.A(S.inc, 1), 2);
+    eq(R.map(S.A(R.__, 100), [S.inc, Math.sqrt]), [101, 10]);
+  });
+
+  it('is curried', function() {
+    eq(S.A(S.inc).length, 1);
+    eq(S.A(S.inc)(1), 2);
+  });
+
+});
```
davidchambers: Regarding spacing:

  - Let's include one empty line after the `'use strict'` pragma.

  - Let's include one empty line between third-party imports and local imports.

  - Let's include two empty lines after the last import.

  - Let's end each file with exactly one trailing `\n`. This script may be useful:

    ```bash
    #!/usr/bin/env bash
    set -e

    # Update the specified files so that each nonempty file ends with
    # exactly one "\n".
    # http://unix.stackexchange.com/q/81685
    for file ; do
      data="$(<"$file")"
      [[ $data == '' ]] || printf '%s\n' "$data" >"$file"
    done
    ```

Bring it all together, we'd get:

```javascript
'use strict';

var R = require('ramda');

var S = require('..');
var eq = require('./shared/utils').eq;


describe('A', function() {

  it('is a binary function', function() {
    eq(typeof S.A, 'function');
    eq(S.A.length, 2);
  });

  it('A(f, x) is equivalent to f(x)', function() {
    eq(S.A(S.inc, 1), 2);
    eq(R.map(S.A(R.__, 100), [S.inc, Math.sqrt]), [101, 10]);
  });

  it('is curried', function() {
    eq(S.A(S.inc).length, 1);
    eq(S.A(S.inc)(1), 2);
  });

});
```

svozza: You're a day late with this advice! ;)

davidchambers: Sorry! You could write a program to update the headers. I don't know whether it'd save time, though.

svozza: Hehe. No bother, I knew what I was signing up for when I offered to do this. Will be a finicky one to get right!

svozza:
All done but for some reason the build is taking ages.

test/shared/utils.js
```diff
@@ -0,0 +1,93 @@
+'use strict';
+var assert = require('assert');
+var R = require('ramda');
+
+var S = require('../..');
```
davidchambers:
```javascript
'use strict';

var assert = require('assert');     // } group for standard library imports

var R = require('ramda');           // } group for third-party imports

var S = require('../..');           // } group for local imports
```

test/shared/utils.js
```diff
@@ -0,0 +1,93 @@
...
+
+
+//  area :: (Number, Number, Number) -> Number !
+var area = function(a, b, c) {
```
davidchambers: Let's export this (and other exported functions) directly:

```javascript
exports.area = function(a, b, c) {…};
```

test/slice.js
```diff
@@ -0,0 +1,118 @@
...
+    eq(S.slice(3, -3, 'ramda'), S.Nothing());
+  });
+
+  it('is curried', function() {
+    eq(S.slice(1).length, 2);
+    eq(S.slice(1)(-1).length, 1);
+    eq(S.slice(1)(-1)(['a', 'b', 'c', 'd', 'e']), S.Just(['b', 'c', 'd']));
+  });
+
+});
```
davidchambers: The script I posted a day or two ago will fix these. ;)

test/shared/utils.js
svozza: Hmmm, I was sure I fixed these already. I even ran the extra line script just before I did my last push.

svozza:
:zap: x `\n`

davidchambers:
LGTM! My last question is whether we __test/shared__ is necessary. Do you envisage other files living in the directory one day? If not, I suggest moving __test/shared/utils.js__ to __test/shared.js__.

svozza:
Yeah, initially I thought there would be more shared functions and that we might want to split them into different files but there were much less than I expected. :zap:

Makefile
```diff
@@ -20,8 +20,8 @@ README.md: index.js
 
 .PHONY: lint
 lint:
-	$(JSHINT) -- index.js test/index.js
-	$(JSCS) -- index.js test/index.js
+	$(JSHINT) -- index.js test/**/*.js
+	$(JSCS) -- index.js test/**/*.js
```
davidchambers: Should this be `test/*.js` now? `make lint` is failing in CircleCI.

svozza:
Interesting. I've been run `npm test` all along thinking that it was running the linter but it's not.

svozza:
I've added `-W053: false` to `jshintrc` because I don't want to add the annotation into 40+ files when we're going to moving to ESLint soon anyway.

davidchambers:
:deciduous_tree: Thanks so much, Stefano! This was a big undertaking. :)

svozza:
No probs! :smile:

PR #187, opened 2016-04-22
rename files to support file systems which are not case-sensitive

davidchambers:
#185 caused problems for me on OS X, because __test/Either.js__ and __test/either.js__ are considered the same filename.

This pull request introduces __test/Either__ and __test/Maybe__ directories for the benefit of case-insensitive file systems.


svozza:
Ah yes, this had crossed my mind but I thought it would only affect Windows.

PR #190, opened 2016-04-23
maybe: add S.maybeToNullable

### LINKED ISSUE #189: maybeToNullable :: Maybe a -> Nullable a
`S.fromMaybe(null, S.Just(42))` is a type error, since there is no type of which `null` and `42` are both members.

@Avaq and I have found ourselves adding this to our projects:

```javascript
//    maybeToNullable :: Maybe a -> Nullable a
const maybeToNullable = m => m.isJust ? m.value : null;
```

It seems reasonable to include this function in Sanctuary, as dealing with JSON at the edges of one's program is often a requirement. It would be nice to have the function defined in one place to save us all from defining it in our various projects.

davidchambers:
Closes #189

### DIFF
--- a/index.js
+++ b/index.js
@@ -1151,6 +1151,8 @@
   //. Takes a default value and a Maybe, and returns the Maybe's value
   //. if the Maybe is a Just; the default value otherwise.
   //.
+  //. See also [`maybeToNullable`](#maybeToNullable).
+  //.
   //. ```javascript
   //. > S.fromMaybe(0, S.Just(42))
   //. 42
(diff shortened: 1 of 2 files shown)


Avaq:
Ooh you did that quick! :)
Maybe add a mention about the `Nullable` pseudotype somewhere?

index.js
```diff
@@ -1164,6 +1166,26 @@
       [a, $Maybe(a), a],
       function(x, maybe) { return maybe.isJust ? maybe.value : x; });
 
+  //# maybeToNullable :: Maybe a -> Nullable a
+  //.
+  //. Returns the given Maybe's value if the Maybe is a Just; `null` otherwise.
+  //. [Nullable][] is defined in sanctuary-def.
```
davidchambers: > Maybe add a mention about the `Nullable` pseudotype somewhere?

Is the sentence above sufficient, do you think?

Avaq: Yep! :+1:

PR #191, opened 2016-04-23
Change type validators to use predicate instead of TypeRep

### LINKED ISSUE #149: Type validation using predicates rather than TypeRep's
Since most type-checking libraries provide convenient ways to get the predicate for a type, this would allow for much greater interoperability with run-time type systems.

With Sanctuary:

``` js
const isInt = allPass([is(Number), x => x % 1 === 0]);

get(isInt, 'x', {x: 0.5}); //-> Nothing
```

With TComb:

``` js
const Int = t.refinement(t.Number, x => x % 1 === 0, 'Int');

get(Int.is, 'x', {x: 0.5}); //-> Nothing
```

With Sanctuary Def:

``` js
const Int = $.NullaryType('my/Int', allPass([is(Number), x => x % 1 === 0]));

get(Int.test, 'x', {x: 0.5}); //-> Nothing
```

The possibilities are _endless!_ Another option is to accept Sanctuary Def types, which would make the code a little more concise, but the possibilities slightly less endless.

Avaq:
Commit description:

> This commit will close #149 by changing all functions
> that have built-in type validation to use predicates
> instead of TypeReps to perform their validations.
> 
> The functions affected are:
> - get()
> - gets()
> - parseJson()
> 
> The only function untouched is `is()`, for it is used
> to turn a TypeRep into a predicate.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2787,64 +2787,62 @@
   }
   S.prop = def('prop', {a: [Accessible]}, [$.String, a, b], prop);
 
-  //# get :: Accessible a => TypeRep b -> String -> a -> Maybe b
+  //# get :: Accessible a => (b -> Boolean) -> String -> a -> Maybe c
   //.
-  //. Takes a [type representative](#type-representatives), a property
-  //. name, and an object and returns Just the value of the specified object
-  //. property if it is of the specified type (according to [`is`](#is));
-  //. Nothing otherwise.
-  //.
-  //. The `Object` type representative may be used as a catch-all since most
-  //. values have `Object.prototype` in their prototype chains.
+  //. Takes a predicate, a property name, and an object and returns Just the
+  //. value of the specified object property if it exists and the value
+  //. satisfies the given predicate; Nothing otherwise.
   //.
   //. See also [`gets`](#gets) and [`prop`](#prop).
   //.
   //. ```javascript
-  //. > S.get(Number, 'x', {x: 1, y: 2})
+  //. > S.get(S.is(Number), 'x', {x: 1, y: 2})
   //. Just(1)
   //.
-  //. > S.get(Number, 'x', {x: '1', y: '2'})
+  //. > S.get(S.is(Number), 'x', {x: '1', y: '2'})
   //. Nothing
   //.
-  //. > S.get(Number, 'x', {})
+  //. > S.get(S.is(Number), 'x', {})
   //. Nothing
   //. ```
-  function get(typeRep, key, obj) {
-    var x = obj[key];
-    return is(typeRep, x) ? Just(x) : Nothing;
+  function get(pred, key, obj) {
+    var x = null;
+    return key in obj && pred(x = obj[key]) ? Just(x) : Nothing;
   }
   S.get =
-  def('get', {a: [Accessible]}, [TypeRep, $.String, a, $Maybe(b)], get);
+  def('get', {a: [Accessible]}, [$.Function, $.String, a, $Maybe(c)], get);
 
-  //# gets :: Accessible a => TypeRep b -> Array String -> a -> Maybe b
+  //# gets :: Accessible a => (b -> Boolean) -> Array String -> a -> Maybe c
   //.
-  //. Takes a [type representative](#type-representatives), an array of
-  //. property names, and an object and returns Just the value at the path
-  //. specified by the array of property names if such a path exists and
-  //. the value is of the specified type; Nothing otherwise.
+  //. Takes a predicate, an array of property names, and an object and returns
+  //. Just the value at the path specified by the array of property names if
+  //. such a path exists and the value satisfies the given predicate; Nothing
+  //. otherwise.
   //.
   //. See also [`get`](#get).
   //.
   //. ```javascript
-  //. > S.gets(Number, ['a', 'b', 'c'], {a: {b: {c: 42}}})
+  //. > S.gets(S.is(Number), ['a', 'b', 'c'], {a: {b: {c: 42}}})
   //. Just(42)
   //.
-  //. > S.gets(Number, ['a', 'b', 'c'], {a: {b: {c: '42'}}})
+  //. > S.gets(S.is(Number), ['a', 'b', 'c'], {a: {b: {c: '42'}}})
   //. Nothing
   //.
-  //. > S.gets(Number, ['a', 'b', 'c'], {})
+  //. > S.gets(S.is(Number), ['a', 'b', 'c'], {})
   //. Nothing
   //. ```
-  function gets(typeRep, keys, obj) {
+  function gets(pred, keys, obj) {
     var maybe = keys.reduce(function(m, k) {
-      return m.chain(function(x) { return x == null ? Nothing : Just(x[k]); });
+      return m.chain(function(x) {
+        return x != null && k in x ? Just(x[k]) : Nothing;
+      });
     }, Just(obj));
-    return filter(function(x) { return is(typeRep, x); }, maybe);
+    return filter(pred, maybe);
   }
   S.gets =
   def('gets',
       {a: [Accessible]},
-      [TypeRep, $.Array($.String), a, $Maybe(b)],
+      [$.Function, $.Array($.String), a, $Maybe(c)],
       gets);
 
   //# keys :: StrMap a -> Array String
@@ -3255,29 +3253,27 @@
   S.parseInt =
   def('parseInt', {}, [$.Integer, $.String, $Maybe($.Integer)], parseInt_);
 
-  //# parseJson :: TypeRep a -> String -> Maybe a
+  //# parseJson :: (a -> Boolean) -> String -> Maybe b
   //.
-  //. Takes a [type representative](#type-representatives) and a string which
-  //. may or may not be valid JSON, and returns Just the result of applying
-  //. `JSON.parse` to the string *if* the result is of the specified type
-  //. (according to [`is`](#is)); Nothing otherwise.
+  //. Takes a predicate and a string which may or may not be valid JSON, and
+  //. returns Just the result of applying `JSON.parse` to the string *if* the
+  //. result satisfies the predicate; Nothing otherwise.
   //.
   //. ```javascript
-  //. > S.parseJson(Array, '["foo","bar","baz"]')
+  //. > S.parseJson(S.is(Array), '["foo","bar","baz"]')
   //. Just(['foo', 'bar', 'baz'])
   //.
-  //. > S.parseJson(Array, '[')
+  //. > S.parseJson(S.is(Array), '[')
   //. Nothing
   //.
-  //. > S.parseJson(Object, '["foo","bar","baz"]')
+  //. > S.parseJson(S.is(Object), '["foo","bar","baz"]')
   //. Nothing
   //. ```
-  function parseJson(typeRep, s) {
-    return filter(function(x) { return is(typeRep, x); },
-                  encase(JSON.parse, s));
+  function parseJson(pred, s) {
+    return filter(pred, encase(JSON.parse, s));
   }
   S.parseJson =
-  def('parseJson', {}, [TypeRep, $.String, $Maybe(a)], parseJson);
+  def('parseJson', {}, [$.Function, $.String, $Maybe(b)], parseJson);
 
   //. ### RegExp
 
--- a/test/get.js
+++ b/test/get.js
@@ -17,44 +17,45 @@ test('get', function() {
          TypeError,
          'Invalid value\n' +
          '\n' +
-         'get :: Accessible a => TypeRep -> String -> a -> Maybe b\n' +
-         '                       ^^^^^^^\n' +
+         'get :: Accessible a => Function -> String -> a -> Maybe c\n' +
+         '                       ^^^^^^^^\n' +
          '                          1\n' +
          '\n' +
          '1)  [1, 2, 3] :: Array Number, Array FiniteNumber, Array NonZeroFiniteNumber, Array Integer, Array ValidNumber\n' +
          '\n' +
-         'The value at position 1 is not a member of ‘TypeRep’.\n');
+         'The value at position 1 is not a member of ‘Function’.\n');
 
-  throws(function() { S.get(Number, [1, 2, 3]); },
+  throws(function() { S.get(S.is(Number), [1, 2, 3]); },
          TypeError,
          'Invalid value\n' +
          '\n' +
-         'get :: Accessible a => TypeRep -> String -> a -> Maybe b\n' +
-         '                                  ^^^^^^\n' +
-         '                                    1\n' +
+         'get :: Accessible a => Function -> String -> a -> Maybe c\n' +
+         '                                   ^^^^^^\n' +
+         '                                     1\n' +
          '\n' +
          '1)  [1, 2, 3] :: Array Number, Array FiniteNumber, Array NonZeroFiniteNumber, Array Integer, Array ValidNumber\n' +
          '\n' +
          'The value at position 1 is not a member of ‘String’.\n');
 
-  throws(function() { S.get(Number, 'x', null); },
+  throws(function() { S.get(S.is(Number), 'x', null); },
          TypeError,
          'Type-class constraint violation\n' +
          '\n' +
-         'get :: Accessible a => TypeRep -> String -> a -> Maybe b\n' +
-         '       ^^^^^^^^^^^^                         ^\n' +
-         '                                            1\n' +
+         'get :: Accessible a => Function -> String -> a -> Maybe c\n' +
+         '       ^^^^^^^^^^^^                          ^\n' +
+         '                                             1\n' +
          '\n' +
          '1)  null :: Null\n' +
          '\n' +
          '‘get’ requires ‘a’ to satisfy the Accessible type-class constraint; the value at position 1 does not.\n');
 
-  eq(S.get(Number, 'x', {x: 0, y: 42}), S.Just(0));
-  eq(S.get(Number, 'y', {x: 0, y: 42}), S.Just(42));
-  eq(S.get(Number, 'z', {x: 0, y: 42}), S.Nothing);
-  eq(S.get(String, 'x', {x: 0, y: 42}), S.Nothing);
+  eq(S.get(S.is(Number), 'x', {x: 0, y: 42}), S.Just(0));
+  eq(S.get(S.is(Number), 'y', {x: 0, y: 42}), S.Just(42));
+  eq(S.get(S.is(Number), 'z', {x: 0, y: 42}), S.Nothing);
+  eq(S.get(S.is(String), 'z', {x: 0, y: 42}), S.Nothing);
+  eq(S.get(S.is(String), 'x', {x: 0, y: 42}), S.Nothing);
 
-  eq(S.get(RegExp, 'x', {x: vm.runInNewContext('/.*/')}), S.Just(/.*/));
-  eq(S.get(vm.runInNewContext('RegExp'), 'x', {x: /.*/}), S.Just(/.*/));
+  eq(S.get(S.is(RegExp), 'x', {x: vm.runInNewContext('/.*/')}), S.Just(/.*/));
+  eq(S.get(S.is(vm.runInNewContext('RegExp')), 'x', {x: /.*/}), S.Just(/.*/));
 
 });
--- a/test/gets.js
+++ b/test/gets.js
@@ -17,47 +17,47 @@ test('gets', function() {
          TypeError,
          'Invalid value\n' +
          '\n' +
-         'gets :: Accessible a => TypeRep -> Array String -> a -> Maybe b\n' +
-         '                        ^^^^^^^\n' +
+         'gets :: Accessible a => Function -> Array String -> a -> Maybe c\n' +
+         '                        ^^^^^^^^\n' +
          '                           1\n' +
          '\n' +
          '1)  [1, 2, 3] :: Array Number, Array FiniteNumber, Array NonZeroFiniteNumber, Array Integer, Array ValidNumber\n' +
          '\n' +
-         'The value at position 1 is not a member of ‘TypeRep’.\n');
+         'The value at position 1 is not a member of ‘Function’.\n');
 
-  throws(function() { S.gets(Number, null); },
+  throws(function() { S.gets(S.is(Number), null); },
          TypeError,
          'Invalid value\n' +
          '\n' +
-         'gets :: Accessible a => TypeRep -> Array String -> a -> Maybe b\n' +
-         '                                   ^^^^^^^^^^^^\n' +
-         '                                        1\n' +
+         'gets :: Accessible a => Function -> Array String -> a -> Maybe c\n' +
+         '                                    ^^^^^^^^^^^^\n' +
+         '                                         1\n' +
          '\n' +
          '1)  null :: Null\n' +
          '\n' +
          'The value at position 1 is not a member of ‘Array String’.\n');
 
-  throws(function() { S.gets(Number, [], null); },
+  throws(function() { S.gets(S.is(Number), [], null); },
          TypeError,
          'Type-class constraint violation\n' +
          '\n' +
-         'gets :: Accessible a => TypeRep -> Array String -> a -> Maybe b\n' +
-         '        ^^^^^^^^^^^^                               ^\n' +
-         '                                                   1\n' +
+         'gets :: Accessible a => Function -> Array String -> a -> Maybe c\n' +
+         '        ^^^^^^^^^^^^                                ^\n' +
+         '                                                    1\n' +
          '\n' +
          '1)  null :: Null\n' +
          '\n' +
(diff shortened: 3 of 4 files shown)


index.js
```diff
@@ -2341,25 +2341,24 @@
   //. ```
   S.lastIndexOf = sanctifyIndexOf('lastIndexOf');
 
-  //# pluck :: Accessible a => TypeRep b -> String -> [a] -> [Maybe b]
+  //# pluck :: Accessible a => (b -> Boolean) -> String -> [a] -> [Maybe c]
```
Avaq: I'm not super happy with the type signatures. Both `b` and `c` are "unique type variables", so they may as well be `Any`. I wish there was some way to indicate that `c` is the set of values that pass the predicate inside the type signature. Perhaps a `Predicate a` pseudotype?

Avaq:
A major concern I have about this change is the ease of upgrading. First, `get` and `gets` are some of my most used functions in Sanctuary, but since many TypeReps are actually also Functions, the type checker will not complain. Take the following example:

```js
S.get(Array, 'x', {x: 1})
```

Before this commit, this would return `Nothing()`, which is what the developer intended to have happen. *After* this commit, this will in fact return `Just(1)`, because `Array` is being used as a predicate to filter and `Boolean(Array(1)) === true`.

A half-solution would be to change filter to be more strict about the output of predicates:

```diff
  //  filter :: (Monad m, Monoid m) => ((a -> Boolean), m a) -> m a
  var filter = function(pred, m) {
    return m.chain(function(x) {
-      return pred(x) ? m.of(x) : m.empty();
+      return pred(x) === true ? m.of(x) : m.empty();
    });
  };
```

But this would just make the function always return `Nothing()` when given an "invalid predicate" like that. Still quite difficult for a developer to figure out what's going wrong (especially since it will probably happen in lots of places).

It would be somewhat better if we asserted the output of the predicate to be a `Boolean`, at least. How is that function checking feature coming along? :)

index.js
davidchambers: > Perhaps a `Predicate a` pseudotype?

I'm not sure how this would work. Could you provide the type signature you have in mind?

Avaq: The idea would be that `Predicate a` somehow constrains `a` to be the set of values for which the predicate function of type `Predicate a` returned true.. It doesn't quite work like that I think, but I wish it could.

davidchambers:
@Avaq, could you point to the __dc-higher-order-functions__ branch of sanctuary-def and try the `$._UnaryFunction` type constructor? It doesn't yet support type variables, but you could use `$._UnaryFunction($.Any, $.Boolean)` for now. :)

Avaq:
I rebased this on `sanctuary#master`, but `sanctuary-def#dc-higher-order-functions` is still at 0.5, whereas `master` wants `0.6.x`. Would it be a good idea to rebase `sanctuary-def#dc-higher-order-functions` as well? Otherwise I feel like I'm working in the past. ^^

davidchambers:
I just rebased that branch for you. :)

index.js
```diff
@@ -279,6 +279,9 @@
   var l = $.TypeVariable('l');
   var r = $.TypeVariable('r');
 
+  //  Predicate :: Type
+  var Predicate = $._UnaryFunction($.Any, $.Boolean);
```
Avaq: Not sure where to put this, nor how to signature it.

test/gets.js
```diff
@@ -20,59 +21,76 @@ describe('gets', function() {
...
-                   'gets :: Accessible a => TypeRep -> Array String -> a -> Maybe b\n' +
-                   '                                   ^^^^^^^^^^^^\n' +
-                   '                                        1\n' +
+                   'gets :: Accessible a => (Any -> Boolean) -> Array String -> a -> Maybe c\n' +
+                   '                                ^^^^^^^\n' +
+                   '                                   1\n' +
+                   '\n' +
+                   '1)  "a" :: String\n' +
+                   '\n' +
+                   'The value at position 1 is not a member of ‘Boolean’.\n'));
```
Avaq: I added these additional assertions in the unit tests to make sure the cases of `S.get(Array, 'x', {x: 1}) -> Just(1)` are resolved once and for all.

index.js
davidchambers: Sounds good to me, since in each case the predicate must accept any value. For a function such as `filter`, though, we'd want `a -> Boolean` rather than `Any -> Boolean`.

index.js
```diff
@@ -2571,25 +2574,24 @@
         return Nothing();
       });
 
-  //# pluck :: Accessible a => TypeRep b -> String -> Array a -> Array (Maybe b)
+  //# pluck :: Accessible a => (b -> Boolean) -> String -> [a] -> [Maybe c]
```
davidchambers: Let's stick with `Array a` rather than `[a]`, in accordance with #180.

index.js
```diff
@@ -2571,25 +2574,24 @@
...
   //.
-  //. Takes a [type representative](#type-representatives), a property name,
-  //. and an array of objects and returns an array of equal length. Each
-  //. element of the output array is Just the value of the specified property
-  //. of the corresponding object if the value is of the specified type
-  //. (according to [`is`](#is)); Nothing otherwise.
+  //. Takes a predicate, a property name, and a list of objects and returns a
+  //. list of equal length. Each element of the output list is Just the value
```
davidchambers: `%s/a list/an array/g`

index.js
```diff
@@ -2686,66 +2688,62 @@
...
   var get = S.get =
   def('get',
       {a: [Accessible]},
-      [TypeRep, $.String, a, $Maybe(b)],
-      function(type, key, obj) { return filter(is(type), Just(obj[key])); });
+      [Predicate, $.String, a, $Maybe(c)],
+      function(pred, key, obj) { return filter(pred, Just(obj[key])); });
 
-  //# gets :: Accessible a => TypeRep b -> Array String -> a -> Maybe b
+  //# gets :: Accessible a => (b -> Boolean) -> [String] -> a -> Maybe c
```
davidchambers: Let's stick with `Array String` here.

test/get.js
```diff
@@ -1,10 +1,11 @@
 'use strict';
 var throws = require('assert').throws;
-var vm = require('vm');
 
 var eq = require('./utils').eq;
 var errorEq = require('./utils').errorEq;
 var S = require('..');
+var T = function() { return true; };
+var F = function() { return false; };
```
davidchambers: These functions should be unary rather than nullary, to match the expected type. Perhaps `S.K(true)` is short enough?

test/gets.js
davidchambers: :thumbsup:

index.js
Avaq: Whoops! I missed that while resolving the merge conflicts. Silly me. ;/

test/get.js
Avaq: Ah. Shouldn't `$._UnaryFunction` have warned me though?

davidchambers: > Shouldn't `$._UnaryFunction` have warned me though?

I'm not sure about `length` checking. `S.add` is a valid unary function even though it reports its `length` as `2`.

Avaq: Right. I guess you could `length >= 1` but then `(x = 1) => ...` wouldn't work.

Avaq:
Seen that `Error: Cannot find module 'to-iso-string'` before?

davidchambers:
> Seen that `Error: Cannot find module 'to-iso-string'` before?

No. That's new to me.

davidchambers:
It was an intermittent (network?) error, apparently. Tests are now passing (after rebuilding). :)

Avaq:
I'm in the process of rebasing this, because it's on the to-do list for the next release, but it may take a while. A lot has changed.

davidchambers:
That's exciting news, @Avaq! Thank you. :)

Avaq:
I've made the rebase. Several things have changed.

* `pluck` is no longer affected
* I dropped the changes which depended on a specific commit hash of `sanctuary-def`. I think it's better to merge this first and change it once sanctuary-def is updated. Otherwise we keep running in circles and stuff will never be merged.

Avaq:
Doc tests didn't run in 7.0. Fixing docs.

Avaq:
> I think it's better to merge this first 

Let me correct that; The reason we didn't merge just came back to me. It was [this concern](https://github.com/sanctuary-js/sanctuary/pull/191#issuecomment-213743955) I had about `S.get(Array, 'x', {x: 1})` returning `Just(1)`, making upgrading a pain.

Perhaps we should hold off until #232 is merged. :neutral_face:

test/get.js
```diff
@@ -17,44 +16,44 @@ test('get', function() {
...
          '\n' +
          '‘get’ requires ‘a’ to satisfy the Accessible type-class constraint; the value at position 1 does not.\n');
 
-  eq(S.get(Number, 'x', {x: 0, y: 42}), S.Just(0));
-  eq(S.get(Number, 'y', {x: 0, y: 42}), S.Just(42));
-  eq(S.get(Number, 'z', {x: 0, y: 42}), S.Nothing);
-  eq(S.get(String, 'x', {x: 0, y: 42}), S.Nothing);
+  eq(S.get(S.K(true), 'x', {x: 0, y: 42}), S.Just(0));
+  eq(S.get(S.K(true), 'y', {x: 0, y: 42}), S.Just(42));
+  eq(S.get(S.K(true), 'z', {x: 0, y: 42}), S.Just(undefined));
```
Avaq: This behavior was discussed with @davidchambers on Gitter, and he noted that he would prefer this to return `Nothing` (by determining beforehand whether the object property even exists). This has several implications:

* The behavior will change from how it currently works. Currently; if we were to provide a `TypeRep Nil`, we would correctly get `Just(undefined)`. The only reason we don't come across this is because `TypeRep Nil` does not exist.
* We will have to make a decision on what constitutes existence of an object property, eg. do we walk the prototype chain? Currently we delegate all of the decision making to the type identifier.
* I'm concerned that behavior might be unexpected. If I provide a type identifier which includes `undefined` as one of its members eg. `union([Nil, String])` then it would behave the same as `String` for JSON parsed data.

What are your thoughts?

safareli: What about this?
```js
function get(pred, key, obj) {
  var x = obj[key];
  return key in myObj && pred(x) ? Just(x) : Nothing;
}
```

davidchambers: I draw a distinction between `{x: undefined}` and `{}`. In the Sanctuary world we should never think _the result of `({}).x` is `undefined`_. Instead, we should think _the result of indiscriminately getting the value of the `x` property of `{}` is `Nothing`_.

Avaq: Understood. Your reasoning makes sense, yet my concern still stands. I will make the change, and I'll have to see how it plays out in practice once it's released. It's a minor concern anyway. :)

index.js
```diff
@@ -2787,64 +2787,62 @@
...
-  //. > S.gets(Number, ['a', 'b', 'c'], {})
+  //. > S.gets(S.is(Number), ['a', 'b', 'c'], {})
   //. Nothing
   //. ```
-  function gets(typeRep, keys, obj) {
+  function gets(pred, keys, obj) {
     var maybe = keys.reduce(function(m, k) {
-      return m.chain(function(x) { return x == null ? Nothing : Just(x[k]); });
+      return m.chain(function(x) {
+        return x != null && k in x ? Just(x[k]) : Nothing;
```
Avaq: Decided to negate the `null` check rather than negating the `k in x` check for aesthetic reasons.

test/get.js
```diff
@@ -1,9 +1,8 @@
 'use strict';
 
-var vm = require('vm');
-
 var S = require('..');
 
+var vm = require('vm');
```
davidchambers: Revert this change, please. :)

Avaq: :zap:

test/gets.js
```diff
@@ -17,47 +17,46 @@ test('gets', function() {
...
+  eq(S.gets(S.is(Number), ['y'], {x: {z: 0}, y: 42}), S.Just(42));
+  eq(S.gets(S.is(Number), ['z'], {x: {z: 0}, y: 42}), S.Nothing);
+  eq(S.gets(S.is(Number), ['x', 'z'], {x: {z: 0}, y: 42}), S.Just(0));
+  eq(S.gets(S.is(Number), ['a', 'b', 'c'], {x: {z: 0}, y: 42}), S.Nothing);
+  eq(S.gets(S.is(Number), [], {x: {z: 0}, y: 42}), S.Nothing);
+  eq(S.gets(S.is(Object), [], {x: {z: 0}, y: 42}), S.Just({x: {z: 0}, y: 42}));
 
-  eq(S.gets(RegExp, ['x'], {x: vm.runInNewContext('/.*/')}), S.Just(/.*/));
-  eq(S.gets(vm.runInNewContext('RegExp'), ['x'], {x: /.*/}), S.Just(/.*/));
+  eq(S.gets(S.K(true), ['x'], {x: vm.runInNewContext('/.*/')}), S.Just(/.*/));
```
davidchambers: Could we use `S.is(RegExp)` here rather than `S.K(true)`, and reinstate the second assertion?

Avaq: I figured `S` would be undefined in a new context. But I haven't asserted that. Let me check. :)

test/gets.js
```diff
@@ -17,47 +17,46 @@ test('gets', function() {
...
+         '                        ^^^^^^^^\n' +
          '                           1\n' +
          '\n' +
          '1)  [1, 2, 3] :: Array Number, Array FiniteNumber, Array NonZeroFiniteNumber, Array Integer, Array ValidNumber\n' +
          '\n' +
-         'The value at position 1 is not a member of ‘TypeRep’.\n');
+         'The value at position 1 is not a member of ‘Function’.\n');
 
-  throws(function() { S.gets(Number, null); },
+  throws(function() { S.gets(S.K(true), null); },
```
davidchambers: Let's use `S.is(Number)` here (and below).

index.js
```diff
@@ -3255,29 +3253,27 @@
...
   //.
-  //. > S.parseJson(Object, '["foo","bar","baz"]')
+  //. > S.parseJson(S.is(Object), '["foo","bar","baz"]')
   //. Nothing
   //. ```
-  function parseJson(typeRep, s) {
-    return filter(function(x) { return is(typeRep, x); },
-                  encase(JSON.parse, s));
+  function parseJson(pred, s) {
+    return filter(function(x) { return pred(x); }, encase(JSON.parse, s));
```
davidchambers: `function(x) { return pred(x); }` is better written `pred`. :stuck_out_tongue_winking_eye:

test/gets.js
Avaq: Indeed; `ReferenceError: S is not defined`.

index.js
Avaq: Very good point. I'll change it in both places.

test/gets.js
Avaq: Same for the `get` tests? Should I change all `S.K(Bool)` functions to appropriate `S.is(TypeRep)` functions?

index.js
```diff
@@ -2787,64 +2787,62 @@
...
   //.
-  //. > S.get(Number, 'x', {})
+  //. > S.get(S.is(Number), 'x', {})
   //. Nothing
   //. ```
-  function get(typeRep, key, obj) {
+  function get(pred, key, obj) {
     var x = obj[key];
-    return is(typeRep, x) ? Just(x) : Nothing;
+    return key in obj && pred(x) ? Just(x) : Nothing;
```
davidchambers: We shouldn't attempt `obj[key]` before the `key in obj` check. Let's replace the two lines above with the following:

```javascript
return key in obj && pred(obj[key]) ? Just(obj[key]) : Nothing;
```

safareli: if we want to optimise property access we can do this:
```js
var x = null
return key in obj && (x = obj[key], pred(x) ? Just(x) : Nothing)
```

test/gets.js
davidchambers: This is what I had in mind:

```diff
-  eq(S.gets(RegExp, ['x'], {x: vm.runInNewContext('/.*/')}), S.Just(/.*/));
-  eq(S.gets(vm.runInNewContext('RegExp'), ['x'], {x: /.*/}), S.Just(/.*/));
+  eq(S.gets(S.is(RegExp), ['x'], {x: vm.runInNewContext('/.*/')}), S.Just(/.*/));
+  eq(S.gets(S.is(vm.runInNewContext('RegExp')), ['x'], {x: /.*/}), S.Just(/.*/));
```

This way we'd continue to assert that a regexp from one environment is a member of the type represented by `RegExp` from another environment. Admittedly this is really testing `is` rather than `gets`, but that's okay.

test/gets.js
davidchambers: Yes, please. It's nice for the diff to act as an upgrade guide as much as possible. :)

index.js
Avaq: I suppose it's cleaner that way. Though perhaps it would then be best to wrap the whole statement in an `if`. When [implementing Promise/A+](https://github.com/Avaq/Fluture/pull/35) I had to access the `.then` property *once and only once*, because property access in JavaScript is impure due to custom getters.

test/gets.js
Avaq: Ahhh. Right. :zap:

index.js
davidchambers: I don't mind too much. I'll leave it to you, Aldwin. If you go with Irakli's approach, we could avoid using the comma operator:

```javascript
var x;
return key in obj && pred(x = obj[key]) ? Just(x) : Nothing;
```

This approach seems okay too:

```javascript
if (key in obj) {
  var x = obj[key];
  if (pred(x)) return Just(x);
}
return Nothing;
```

Avaq: @safareli Your statement returns `false` rather than `Nothing` for non-existent properties.

Avaq: Can't make it better than `key in obj ? (x = obj[key], pred(x)) ? Just(x) : Nothing : Nothing` :\

test/parseJson.js
```diff
@@ -15,28 +15,28 @@ test('parseJson', function() {
...
          '1)  [1, 2, 3] :: Array Number, Array FiniteNumber, Array NonZeroFiniteNumber, Array Integer, Array ValidNumber\n' +
          '\n' +
          'The value at position 1 is not a member of ‘String’.\n');
 
-  eq(S.parseJson(Object, '[Invalid JSON]'), S.Nothing);
-  eq(S.parseJson(Array, '{"foo":"bar"}'), S.Nothing);
-  eq(S.parseJson(Array, '["foo","bar"]'), S.Just(['foo', 'bar']));
+  eq(S.parseJson(S.K(true), '[Invalid JSON]'), S.Nothing);
+  eq(S.parseJson(S.K(false), '{"foo":"bar"}'), S.Nothing);
+  eq(S.parseJson(S.K(true), '["foo","bar"]'), S.Just(['foo', 'bar']));
```
Avaq: Fixing these.

index.js
safareli: @Avaq you are right.
```
// this
key in obj && (x = obj[key], true) && pred(x) ? Just(x) : Nothing
// and this will work
key in obj && (x = obj[key], 1) && pred(x) ? Just(x) : Nothing
```
but it's a bit cryptic

test/parseJson.js
Avaq: :zap:

test/parseJson.js
```diff
@@ -15,28 +15,28 @@ test('parseJson', function() {
...
          '1)  [1, 2, 3] :: Array Number, Array FiniteNumber, Array NonZeroFiniteNumber, Array Integer, Array ValidNumber\n' +
          '\n' +
          'The value at position 1 is not a member of ‘String’.\n');
 
-  eq(S.parseJson(Object, '[Invalid JSON]'), S.Nothing);
-  eq(S.parseJson(Array, '{"foo":"bar"}'), S.Nothing);
-  eq(S.parseJson(Array, '["foo","bar"]'), S.Just(['foo', 'bar']));
+  eq(S.parseJson(S.is(Object), '[Invalid JSON]'), S.Nothing);
+  eq(S.parseJson(S.is(Array), '{"foo":"bar"}'), S.Nothing);
+  eq(S.parseJson(S.is(Array), '["foo","bar"]'), S.Just(['foo', 'bar']));
```
davidchambers: I'd love to replace the two assertions above with the following:

```javascript
eq(S.parseJson($.test([], $.Array($.String)), '{"foo":"bar"}'), S.Nothing);
eq(S.parseJson($.test([], $.Array($.String)), '[1,2,3,4,5,6]'), S.Nothing);
eq(S.parseJson($.test([], $.Array($.String)), '["foo","bar"]'), S.Just(['foo', 'bar']));
```

Very [aeson][1]-like!

This requires a recent version of sanctuary-def, though. I'm currently working on the pull request to upgrade our sanctuary-def dependency. :)


[1]: https://hackage.haskell.org/package/aeson

index.js
```diff
@@ -2787,64 +2787,62 @@
...
   //.
-  //. > S.get(Number, 'x', {})
+  //. > S.get(S.is(Number), 'x', {})
   //. Nothing
   //. ```
-  function get(typeRep, key, obj) {
-    var x = obj[key];
-    return is(typeRep, x) ? Just(x) : Nothing;
+  function get(pred, key, obj) {
+    var x = null;
```
davidchambers: Thoughts on `var x = null;` versus `var x;` here?

index.js
```diff
@@ -2787,64 +2787,62 @@
...
 
-  //# get :: Accessible a => TypeRep b -> String -> a -> Maybe b
+  //# get :: Accessible a => (b -> Boolean) -> String -> a -> Maybe c
   //.
-  //. Takes a [type representative](#type-representatives), a property
-  //. name, and an object and returns Just the value of the specified object
-  //. property if it is of the specified type (according to [`is`](#is));
+  //. Takes a predicate, a property name, and an object and returns Just the
+  //. value of the specified object property if it passes the given predicate;
   //. Nothing otherwise.
```
davidchambers: We should state that `Nothing` is returned if the property does not exist:

_Takes a predicate, a property name, and an object and returns Just the value of the specified object property if <del>it passes</del> <ins>the property exists and the value satisfies</ins> the given predicate; Nothing otherwise._

index.js
```diff
@@ -2787,64 +2787,62 @@
...
-  //# gets :: Accessible a => TypeRep b -> Array String -> a -> Maybe b
+  //# gets :: Accessible a => (b -> Boolean) -> Array String -> a -> Maybe c
   //.
-  //. Takes a [type representative](#type-representatives), an array of
-  //. property names, and an object and returns Just the value at the path
-  //. specified by the array of property names if such a path exists and
-  //. the value is of the specified type; Nothing otherwise.
+  //. Takes a predicate, an array of property names, and an object and returns
+  //. Just the value at the path specified by the array of property names if
+  //. such a path exists and the value passes the given predicate; Nothing
```
davidchambers: I'd prefer "satisfies" in place of "passes".

index.js
Avaq: Let's make it consistent with similar code.

index.js
davidchambers: Scroll up for a more elegant solution, @safareli. ;)

index.js
Avaq: I realized there is no similar case in the entire code base.

index.js
```diff
@@ -2787,64 +2787,62 @@
...
   //.
-  //. Takes a [type representative](#type-representatives), a property
-  //. name, and an object and returns Just the value of the specified object
-  //. property if it is of the specified type (according to [`is`](#is));
-  //. Nothing otherwise.
-  //.
-  //. The `Object` type representative may be used as a catch-all since most
-  //. values have `Object.prototype` in their prototype chains.
+  //. Takes a predicate, a property name, and an object and returns Just the
+  //. value of the specified object property if it exists and the value
```
Avaq: Went for "*it*" over your suggested "*the property*".

davidchambers: This seems slightly ambiguous to me. I see two valid interpretations:

  - _return Just (the value of the specified object property) if it exists_; or
  - _return Just the value of (the specified object property) if it exists_.

"It" can refer to either parenthesized term, depending on where one's brain decides the parens should go. For some reason my brain wants to include "the value of" in the parens.

Avaq: Hm, I found that it read more fluently. Referring to the existence of the subject clears the ambiguity to me - it doesn't matter whether we talk about the existence of the property or it's value. Non-existent applies to the whole pair. ;)
I guess you should decide in the end. What will it be? :)

davidchambers: I find your argument persuasive. Let's stick with "it". :grinning:

Avaq:
So what do you think? Merge this before or after #232?

davidchambers:
Go ahead and squash, rebase on `master`, and update this pull request. Once the tests pass I will merge it. :)

Avaq:
:zap:

Avaq:
Seems like a test-runner related error. Has to do with running code in iframes.

Shard:
I re-ran the tests and they worked 😕

davidchambers:
I've hit that error once too. It seems to be nondeterministic.

davidchambers:
:deciduous_tree:

Issue #119, opened 2015-11-17
either: implement Foldable


davidchambers:
[Foldable](https://github.com/fantasyland/fantasy-land#foldable)

PR #192, opened 2016-04-23
Add S.sum and S.product

svozza:
I had to move a few functions to the top of the file so that we could replace the Ramda versions used throughout the code.

### DIFF
--- a/index.js
+++ b/index.js
@@ -228,7 +228,7 @@
   var Foldable = $.TypeClass(
     'sanctuary/Foldable',
     function(x) {
-      return _type(x) === 'Array' || hasMethod('reduce');
+      return _type(x) === 'Array' || hasMethod('reduce')(x);
     }
   );
 
@@ -266,6 +266,7 @@
   var b = $.TypeVariable('b');
   var c = $.TypeVariable('c');
   var d = $.TypeVariable('d');
+  var f = $.TypeVariable('f');
   var l = $.TypeVariable('l');
   var r = $.TypeVariable('r');
 
@@ -678,12 +679,12 @@
   //. > S.meld([Math.pow, S.sub])(3)(4)(5)
   //. 76
   //. ```
-  var meld = S.meld =
+  S.meld =
   def('meld',
       {},
       [$.Array($.Function), $.Function],
       function(fs) {
-        var n = 1 + R.sum(R.map(R.length, fs)) - fs.length;
+        var n = 1 + sum(R.map(R.length, fs)) - fs.length;
         return R.curryN(n, function() {
           var args = Array.prototype.slice.call(arguments);
           for (var idx = 0; idx < fs.length; idx += 1) {
@@ -948,7 +949,7 @@
   //. > S.Nothing().map(S.inc)
   //. Nothing()
   //.
-  //. > S.Just([1, 2, 3]).map(R.sum)
+  //. > S.Just([1, 2, 3]).map(S.sum)
   //. Just(6)
   //. ```
   Maybe.prototype.map =
@@ -1253,7 +1254,7 @@
   def('mapMaybe',
       {},
       [$.Function, $.Array(a), $.Array(b)],
-      meld([R.map, justs]));
+      function(f, xs) { return justs(R.map(f, xs)); });
 
   //# encase :: (a -> b) -> a -> Maybe b
   //.
@@ -1566,7 +1567,7 @@
   //. > S.Left('Cannot divide by zero').map(S.inc)
   //. Left('Cannot divide by zero')
   //.
-  //. > S.Right([1, 2, 3]).map(R.sum)
+  //. > S.Right([1, 2, 3]).map(S.sum)
   //. Right(6)
   //. ```
   Either.prototype.map =
@@ -2528,7 +2529,7 @@
   //. > S.reduce(xs => x => [x].concat(xs), [], [1, 2, 3, 4, 5])
   //. [5, 4, 3, 2, 1]
   //. ```
-  S.reduce =
+  var reduce = S.reduce =
   def('reduce',
       {b: [Foldable]},
       [$.Function, a, b, a],
@@ -2699,6 +2700,29 @@
       [$.FiniteNumber, $.FiniteNumber, $.FiniteNumber],
       function(a, b) { return a + b; });
 
+  //# sum :: Foldable f => f FiniteNumber -> FiniteNumber
+  //.
+  //. Returns the sum of the given array of (finite) numbers.
+  //.
+  //. ```javascript
+  //. > S.sum([1, 2, 3, 4, 5])
+  //. 15
+  //.
+  //. > S.sum([])
+  //. 0
+  //.
+  //. > S.sum(S.Just(42))
+  //. 42
+  //.
+  //. > S.sum(S.Nothing())
+  //. 0
+  //. ```
+  var sum = S.sum =
+  def('sum',
+      {f: [Foldable]},
+      [f, $.FiniteNumber],
+      reduce(function(a) { return function(b) { return a + b; }; }, 0));
+
   //# sub :: FiniteNumber -> FiniteNumber -> FiniteNumber
   //.
   //. Returns the difference between two (finite) numbers.
@@ -2754,6 +2778,29 @@
       {},
       [$.FiniteNumber, $.FiniteNumber, $.FiniteNumber],
       function(a, b) { return a * b; });
+
+  //# product :: Foldable f => f FiniteNumber -> FiniteNumber
+  //.
+  //. Returns the product of the given array of (finite) numbers.
+  //.
+  //. ```javascript
+  //. > S.product([1, 2, 3, 4, 5])
+  //. 120
+  //.
+  //. > S.product([])
+  //. 1
+  //.
+  //. > S.product(S.Just(42))
+  //. 42
+  //.
+  //. > S.product(S.Nothing())
+  //. 1
+  //. ```
+  S.product =
+  def('product',
+      {f: [Foldable]},
+      [f, $.FiniteNumber],
+      reduce(function(a) { return function(b) { return a * b; }; }, 1));
 
   //# div :: FiniteNumber -> NonZeroFiniteNumber -> FiniteNumber
   //.
--- a/test/product.js
+++ b/test/product.js
@@ -1,0 +1,85 @@
+'use strict';
+
+var throws = require('assert').throws;
+
+var eq = require('./utils').eq;
+var errorEq = require('./utils').errorEq;
+var S = require('..');
+
+
+describe('product', function() {
+
+  it('is a unary function', function() {
+    eq(typeof S.product, 'function');
+    eq(S.product.length, 1);
+  });
+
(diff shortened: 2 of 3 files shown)


index.js
```diff
@@ -2529,11 +2543,17 @@
...
   //. 2
   //. ```
-  S.add =
-  def('add',
-      {},
-      [$.FiniteNumber, $.FiniteNumber, $.FiniteNumber],
-      function(a, b) { return a + b; });
+  S.add = add;
+
+  //# sum :: [FiniteNumber] -> FiniteNumber
```
davidchambers: Should the type be as follows?

```haskell
sum :: Foldable f => f FiniteNumber -> FiniteNumber
```

index.js
```diff
@@ -2591,6 +2611,20 @@
...
+  //.
+  //. ```javascript
+  //. > S.product([1, 2, 3, 4, 5])
+  //. 120
+  //. ```
+  S.product =
+  def('product',
+      {},
+      [$.Array($.FiniteNumber), $.FiniteNumber],
+      reduce(S.mult, 1));
```
davidchambers: `s/S[.]//`

index.js
```diff
@@ -369,6 +369,34 @@
...
+  def('add',
+      {},
+      [$.FiniteNumber, $.FiniteNumber, $.FiniteNumber],
+      function(a, b) { return a + b; });
+
+  var sum =
+  def('sum',
+      {},
+      [$.Array($.FiniteNumber), $.FiniteNumber],
+      reduce(add, 0));
```
davidchambers: I'd be totally okay with replacing `add` with `function(a, b) { return a + b; }` here. I'm not sure it's worth reusing a trivial function if it forces us to separate a function's implementation from its documentation. What do you think?

index.js
```diff
@@ -369,6 +369,34 @@
         }
       });
 
+  var reduce =
```
davidchambers: Let's include a type signature.

index.js
svozza: Yep, I agree.

svozza:
Just realised I forgot to do `git add` with the tests!

index.js
```diff
@@ -369,6 +369,30 @@
...
+        }
+      });
+
+  //  sum :: Foldable f => f FiniteNumber -> FiniteNumber
+  var sum =
+  def('sum',
+      {},
+      [$.Array($.FiniteNumber), $.FiniteNumber],
+      reduce(function(a, b) { return a + b; }, 0));
+
```
davidchambers: It looks as though each of these functions can now be moved alongside its description. `sum` is referenced in the definition of `meld`, but will have been defined by the time the body of that function is executed.

index.js
```diff
@@ -2535,6 +2545,16 @@
       [$.FiniteNumber, $.FiniteNumber, $.FiniteNumber],
       function(a, b) { return a + b; });
 
+  //#  sum :: Foldable f => f FiniteNumber -> FiniteNumber
+  //.
+  //. Returns the sum of a list of (finite) numbers.
```
davidchambers: In light of #180, let's `s/a list/an array/`.

Also, `sum = K(0)` would fit with this description, as it always returns the sum of `[]`, an array of finite numbers. ;)

Let's write *the array* rather than *an array*. I suggest:

_Returns the sum of the given array of (finite) numbers._

index.js
```diff
@@ -2535,6 +2545,16 @@
       [$.FiniteNumber, $.FiniteNumber, $.FiniteNumber],
       function(a, b) { return a + b; });
 
+  //#  sum :: Foldable f => f FiniteNumber -> FiniteNumber
```
davidchambers: `s/# /#/`

test/product.js
```diff
@@ -0,0 +1,72 @@
...
+                   'product :: Array FiniteNumber -> FiniteNumber\n' +
+                   '                 ^^^^^^^^^^^^\n' +
+                   '                      1\n' +
+                   '\n' +
+                   '1)  -Infinity :: Number, ValidNumber\n' +
+                   '\n' +
+                   'The value at position 1 is not a member of ‘FiniteNumber’.\n'));
+  });
+
+  it('returns the product of a list of numbers', function() {
```
davidchambers: `s/a list/an array/`

test/product.js
```diff
@@ -0,0 +1,72 @@
...
+    eq(S.product([1, 2, 3, 4, 5]), 120);
+    eq(S.product([1, 2, 3, 4, -5]), -120);
+  });
```
davidchambers: Let's also test `[]`, an array containing `0`, and an array containing `-0`.

test/sum.js
```diff
@@ -0,0 +1,72 @@
...
+                   '\n' +
+                   '1)  -Infinity :: Number, ValidNumber\n' +
+                   '\n' +
+                   'The value at position 1 is not a member of ‘FiniteNumber’.\n'));
+  });
+
+  it('sums a list of numbers together', function() {
+    eq(S.sum([1, 2, 3, 4, 5]), 15);
+    eq(S.sum([1, 2, 3, 4, -5]), 5);
+  });
```
davidchambers: Let's also test `[]`.

index.js
```diff
@@ -2585,12 +2605,26 @@
...
+  //. Returns the product of a list (finite) numbers.
+  //.
+  //. ```javascript
+  //. > S.product([1, 2, 3, 4, 5])
+  //. 120
+  //. ```
+  S.product =
+  def('product',
+      {},
+      [$.Array($.FiniteNumber), $.FiniteNumber],
```
davidchambers: I think the best we can currently do here is `{f: [Foldable]}, [f, $.FiniteNumber]`.

index.js
svozza: No, the tests fail if I move them and do something like 

```js
var sum = S.sum =
```

davidchambers:
Let's also test these:

```javascript
S.sum(S.Nothing())
S.sum(S.Just(42))
S.sum(S.Left('XXX'))
S.sum(S.Right(42))
```

The last two won't work until we resolve #119, so we could put these assertions in `it.skip` blocks.

index.js
davidchambers: Ah, because `meld` is applied in the definition of `mapMaybe`. Let's change that definition from `meld([R.map, justs])` to something along these lines:

```javascript
function(f, xs) { return justs(R.map(f, xs)); }
```

index.js
svozza: Interesting, I found a bug in the `Foldable` type check function. I think we never saw it before because it always short-circuited on the array check.

svozza:
I've updated it to use the Foldable type constraint. I'm not happy with the error messages for the `Infinity` cases. The `product` message now leaks that it is implemented in terms of `mult` and `sum` gives an error based on the return value so it actually allows the disallowed computation occur.

davidchambers:
> The `product` message now leaks that it is implemented in terms of `mult` and `sum` gives an error based on the return value so it actually allows the disallowed computation occur.

I suggest replacing `mult` with `function(a, b) { return a * b; }` so we refer to the right function.

Complaining about the return value isn't ideal, but it's the best we can do until sanctuary-def permits us to express `Foldable f => f FiniteNumber`.

index.js

svozza:
Yeah, it's not great but I guess there's nothing we can do.

index.js
```diff
@@ -2378,21 +2396,7 @@
...
-          var acc = initial;
-          for (var idx = 0; idx < foldable.length; idx += 1) {
-            acc = f(acc, foldable[idx]);
-          }
-          return acc;
-        } else {
-          return foldable.reduce(f, initial);
-        }
-      });
+  S.reduce = reduce;
```
davidchambers: We can revert this change, I believe.

index.js
```diff
@@ -2535,6 +2539,20 @@
...
       function(a, b) { return a + b; });
 
+  //# sum :: Foldable f => f FiniteNumber -> FiniteNumber
+  //.
+  //. Returns the sum of the given array of (finite) numbers.
+  //.
+  //. ```javascript
+  //. > S.sum([1, 2, 3, 4, 5])
+  //. 15
+  //. ```
```
davidchambers: Let's include `S.sum([])`.

index.js
```diff
@@ -2591,6 +2609,20 @@
...
       function(a, b) { return a * b; });
 
+  //# product :: Foldable f => f FiniteNumber -> FiniteNumber
+  //.
+  //. Returns the product of the given array of (finite) numbers.
+  //.
+  //. ```javascript
+  //. > S.product([1, 2, 3, 4, 5])
+  //. 120
+  //. ```
```
davidchambers: Let's include `S.product([])`.

index.js
davidchambers: Let's also include `S.sum(S.Nothing())` and `S.sum(S.Just(42))`.

index.js
davidchambers: Let's also include `S.product(S.Nothing())` and `S.product(S.Just(42))`.

index.js

index.js

index.js

index.js
```diff
@@ -2378,7 +2379,8 @@
   //. > S.reduce((xs, x) => [x].concat(xs), [], [1, 2, 3, 4, 5])
   //. [5, 4, 3, 2, 1]
   //. ```
-  S.reduce =
+  //  reduce :: Foldable f => (a -> b -> a) -> a -> f b -> a
```
davidchambers: This line is now redundant.

svozza: Oops!

svozza:
Anything else left on this?

davidchambers:
> Anything else left on this?

This looks great, Stefano. This has motivated me to resolve #119. I plan to do so tomorrow. Once that pull request is merged, you'll be able to rebase this one and remove the `.skip`s. :)

test/product.js
```diff
@@ -0,0 +1,85 @@
...
+
+  it('can be applied to Maybes', function() {
+    eq(S.product(S.Nothing()), 1);
+    eq(S.product(S.Just(42)), 42);
+  });
+
+  it.skip('can be applied to Eithers', function() {
```
davidchambers: You can remove the `.skip` now thanks to #204. :)

svozza: :zap:

index.js
```diff
@@ -2699,6 +2700,29 @@
...
+  //. 42
+  //.
+  //. > S.sum(S.Nothing())
+  //. 0
+  //. ```
+  var sum = S.sum =
+  def('sum',
+      {f: [Foldable]},
+      [f, $.FiniteNumber],
+      reduce_(function(a, b) { return a + b; }, 0));
```
davidchambers: Might this be better?

```javascript
reduce(function(a) { return function(b) { return a + b; }; }, 0)
```

`reduce_` will curry the provided function, but we could remove a step by defining a curried function.

We needn't worry about this, though. Once we can express `f FiniteNumber` in the type system we'll be able to use `S.add` in this definition without references to that function appearing in error messages. :)

davidchambers:
:deciduous_tree: Lovely patch, Stefano!

PR #212, opened 2016-05-07
either: implement Traversable

### LINKED ISSUE #120: either: implement Traversable
[Traversable](https://github.com/fantasyland/fantasy-land#traversable)

davidchambers:
Closes #120

This is similar to #121. You're the best person to review this patch, @scott-christopher (though others are of course welcome to do so).

### DIFF
--- a/index.js
+++ b/index.js
@@ -212,6 +212,17 @@
     function(x) { return x != null; }
   );
 
+  //  Applicative :: TypeClass
+  var Applicative = $.TypeClass(
+    'sanctuary/Applicative',
+    function(x) {
+      return (
+        _type(x) === 'Array' ||
+        Apply.test(x) && (hasMethod('of')(x) || hasMethod('of')(x.constructor))
+      );
+    }
+  );
+
   //  Apply :: TypeClass
   var Apply = $.TypeClass(
     'sanctuary/Apply',
@@ -1011,10 +1022,13 @@
   //.
   //. > S.Just(S.Right(42)).sequence(S.Either.of)
   //. Right(Just(42))
+  //.
+  //. > S.Just(S.Left('Cannot divide by zero')).sequence(S.Either.of)
+  //. Left('Cannot divide by zero')
   //. ```
   Maybe.prototype.sequence =
   method('Maybe#sequence',
-         {a: [Apply], b: [Apply]},
+         {a: [Applicative], b: [Applicative]},
          [$Maybe(a), $.Function, b],
          function(maybe, of) {
            return maybe.isJust ? R.map(Just, maybe.value) : of(maybe);
@@ -1357,8 +1371,8 @@
   //. `Either a b` is either a Left whose value is of type `a` or a Right whose
   //. value is of type `b`.
   //.
-  //. The Either type satisfies the [Semigroup][], [Monad][], [Foldable][], and
-  //. [Extend][] specifications.
+  //. The Either type satisfies the [Semigroup][], [Monad][], [Traversable][],
+  //. and [Extend][] specifications.
 
   //# EitherType :: Type -> Type -> Type
   //.
@@ -1613,6 +1627,33 @@
          [$Either(a, b), $.Function, c, c],
          function(either, f, x) {
            return either.isRight ? f(x, either.value) : x;
+         });
+
+  //# Either#sequence :: Applicative f => Either a (f b) ~> (b -> f b) -> f (Either a b)
+  //.
+  //. Evaluates an applicative action contained within the Either,
+  //. resulting in:
+  //.
+  //.   - a pure applicative of a Left if `this` is a Left; otherwise
+  //.
+  //.   - an applicative of a Right of the evaluated action.
+  //.
+  //. ```javascript
+  //. > S.Left('Cannot divide by zero').sequence(S.Maybe.of)
+  //. Just(Left('Cannot divide by zero'))
+  //.
+  //. > S.Right(S.Just(42)).sequence(S.Maybe.of)
+  //. Just(Right(42))
+  //.
+  //. > S.Right(S.Nothing()).sequence(S.Maybe.of)
+  //. Nothing()
+  //. ```
+  Either.prototype.sequence =
+  method('Either#sequence',
+         {b: [Applicative], c: [Applicative]},
+         [$Either(a, b), $.Function, c],
+         function(either, of) {
+           return either.isRight ? R.map(Right, either.value) : of(either);
          });
 
   //# Either#toBoolean :: Either a b ~> Boolean
--- a/test/Either/Either.js
+++ b/test/Either/Either.js
@@ -2,15 +2,118 @@
 
 var throws = require('assert').throws;
 
+var jsc = require('jsverify');
+var R = require('ramda');
+
 var errorEq = require('../utils').errorEq;
 var S = require('../..');
 
+
+//  Identity :: a -> Identity a
+var Identity = function Identity(x) {
+  return {
+    of: Identity,
+    map: function(fn) {
+      return Identity(fn(x));
+    },
+    ap: function(y) {
+      return Identity(x(y));
+    },
+    equals: function(other) {
+      return R.equals(x, other.value);
+    },
+    value: x
+  };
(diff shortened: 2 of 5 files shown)


index.js
```diff
@@ -212,6 +212,17 @@
...
+  var Applicative = $.TypeClass(
+    'sanctuary/Applicative',
+    function(x) {
+      return (
+        _type(x) === 'Array' ||
+        Apply.test(x) && (hasMethod('of')(x) || hasMethod('of')(x.constructor))
+      );
+    }
+  );
+
```
davidchambers: Is this definition correct according to your interpretation of the [spec][1]?


[1]: https://github.com/fantasyland/fantasy-land#applicative

index.js
```diff
@@ -1011,10 +1022,13 @@
...
   //. > S.Just(S.Right(42)).sequence(S.Either.of)
   //. Right(Just(42))
+  //.
+  //. > S.Just(S.Left('Cannot divide by zero')).sequence(S.Either.of)
+  //. Left('Cannot divide by zero')
   //. ```
   Maybe.prototype.sequence =
   method('Maybe#sequence',
-         {a: [Apply], b: [Apply]},
+         {a: [Applicative], b: [Applicative]},
```
davidchambers: Is this accurate?

index.js
```diff
@@ -1615,6 +1629,33 @@
...
 
+  //# Either#sequence :: Applicative f => Either a (f b) ~> (b -> f b) -> f (Either a b)
+  //.
+  //. Evaluates an applicative action contained within the Either,
+  //. resulting in:
+  //.
+  //.   - a pure applicative of a Left if `this` is a Left; otherwise
+  //.
+  //.   - an applicative of a Right of the evaluated action.
+  //.
```
davidchambers: Please confirm that the type signature is correct.

Let me know if you have suggestions as to how to improve the description.

test/Either/Either.js
```diff
@@ -2,15 +2,124 @@
...
+      },
+      value: x
+    };
+  };
+  _Compose.of = function(x) {
+    return _Compose(F.of(G.of(x)));
+  };
+  return _Compose;
+};
+
```
davidchambers: I copied this code from __test/Maybe/Maybe.js__. We could move it to a shared file (or multiple shared files), but my intuition is telling me not to do so.

test/Either/Either.js
```diff
@@ -2,15 +2,124 @@
...
+        var lhs = u.map(C).sequence(C.of);
+        var rhs = C(u.sequence(Identity.of).map(function(x) {
+          return x.sequence(S.Either.of);
+        }));
+        return lhs.equals(rhs);
+      }));
+    });
+
+  });
+
```
davidchambers: Please check these tests. I'll freely admit to not having a complete understanding of Traversable. I more or less copied the tests added in #121 then switched `Maybe` and `Either`, `Nothing` and `Left`, and `Just` and `Right`.

index.js
scott-christopher: Yep, that looks fine. It might be worthwhile adding some functionality to `$.TypeClass` to help extending type classes like in this example.

index.js
scott-christopher: It's arguably correct either way, given that the function to make the `Apply` and `Applicative` is explicitly provided by the caller. That said, I think it think it probably makes more said using `Applicative` here.

test/Maybe/Maybe.js
```diff
@@ -65,6 +67,7 @@ var RightArb = function(arb) {
 var Compose = function(F, G) {
   var _Compose = function _Compose(x) {
     return {
+      constructor: _Compose,
```
davidchambers: In order to test both sides of `hasMethod('of')(x) || hasMethod('of')(x.constructor)` I defined `of` on "instances" for Identity and on the "constructor" for Compose.

index.js
davidchambers: > It might be worthwhile adding some functionality to `$.TypeClass` to help extending type classes like in this example.

Good idea. Do you have an API in mind? Here are two options:

```javascript
//    Applicative :: TypeClass
const Applicative = $.TypeClass(
  'sanctuary/Applicative',
  $.extends(
    [Apply],
    x => _type(x) === 'Array' || hasMethod('of')(x) || hasMethod('of')(x.constructor)
  )
);
```

```javascript
//    Applicative :: TypeClass
const Applicative = $.TypeClass(
  'sanctuary/Applicative',
  [Apply],
  x => _type(x) === 'Array' || hasMethod('of')(x) || hasMethod('of')(x.constructor)
);
```

index.js
davidchambers: > It's arguably correct either way, given that the function to make the `Apply` and `Applicative` is explicitly provided by the caller.

The caller could provide her own `of`, certainly. Presumably, though, the caller will simply provide `Future.of` or whatever.

If it's correct either way, let's go with the option which matches our type signature. :)

index.js
scott-christopher: I think I prefer the brevity of the second example, though that would break the existing API.

test/Maybe/Maybe.js
scott-christopher: 👍 It'd be nice to only have to support that method on the constructor, but such is life.

index.js
davidchambers: I prefer the second approach too. It would make it possible to create a graph of the relationships, as dependencies could be exposed via a field on a TypeClass value rather than being hidden in its predicate.

index.js
scott-christopher: Both the description and signature look good to me.

test/Maybe/Maybe.js
davidchambers: Agreed. I have several complaints about the current state of Fantasy Land, but I'm grateful to have a spec at all and I'm willing to implement it as faithfully as possible.

index.js
davidchambers: Okay. :)

index.js
scott-christopher: Agreed.

test/Either/Either.js
scott-christopher: These should be fine to share internally if you'd prefer. In fact, it might be worthwhile investigating reusing the laws defined in the [FantasyLand repo](https://github.com/fantasyland/fantasy-land/blob/master/laws/traversable.js) at some point in the future (though there are some changes I'd want included to allow for passing in arbitrary functions to the laws rather than use `identity` everywhere first)

It might also be a good idea to have the JSVerify arbitrary instances for each type exported somewhere to encourage reuse.

davidchambers: I'm certainly in favour of sharing code where possible. If we're to move the definitions of `S.Maybe` and `S.Either` to a separate repository, as Kevin has suggested, we'll no doubt reorganize the tests at that point, so I'm happy to live with this duplication in the short term.

test/Either/Either.js
scott-christopher: These look fine too. It gets a little confusing because we're using `Either` as both a `Traversable` and one of the `Applicative`s in the naturality test, which we might be able to tidy up at some point by creating a more general `Applicative` arbitrary instance.

scott-christopher:
🍄

test/Either/Either.js
davidchambers: Could we use Identity (rather than Either) as either the traversable or the applicative to avoid confusion? If so, how would this look?

scott-christopher: Sure. The following should do the trick.

```js
function identityToMaybe(i) {
  return S.Just(i.value);
}

it('satisfies naturality', function() {
  jsc.assert(jsc.forall(EitherArb(jsc.integer, IdentityArb(jsc.string)), function(either) {
    var lhs = identityToMaybe(either.sequence(Identity.of));
    var rhs = either.map(identityToMaybe).sequence(S.Maybe.of);
    return lhs.equals(rhs);
  }));
});
```

davidchambers: Thanks! That's clearer.

:zap:

scott-christopher: 🍄

PR #213, opened 2016-05-08
license: update copyright information and automate future updates

davidchambers:
I updated the Plaid copyright year to 2016, where it will remain.

The Sanctuary copyright year will be updated—to reflect the most recently authored commit—each time we run `make release-$LEVEL`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,13 @@ XYZ = node_modules/.bin/xyz --repo git@github.com:sanctuary-js/sanctuary.git --s
 
 
 .PHONY: all
-all: README.md
+all: LICENSE README.md
+
+.PHONY: LICENSE
+LICENSE:
+	cp -- '$@' '$@.orig'
+	sed 's/Copyright (c) .* Sanctuary/Copyright (c) $(shell git log --date=format:%Y --pretty=format:%ad | sort -r | head -n 1) Sanctuary/' '$@.orig' >'$@'
+	rm -- '$@.orig'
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -7,7 +7,13 @@ XYZ = node_modules/.bin/xyz --repo git@github.com:sanctuary-js/sanctuary.git --s
 
 
 .PHONY: all
-all: README.md
+all: LICENSE README.md
+
+.PHONY: LICENSE
+LICENSE:
+	cp -- '$@' '$@.orig'
+	sed 's/Copyright (c) .* Sanctuary/Copyright (c) $(shell git log --date=format:%Y --pretty=format:%ad | sort -r | head -n 1) Sanctuary/' '$@.orig' >'$@'
```
scott-christopher: You can remove the need for `sort` and `head` if you just want the year of the latest commit by passing `-1` as an argument to `git log`.
```
git log -1 --date=format:%Y --pretty=format:%ad
```

davidchambers: I did consider `-1`. It's hypothetically possible, though, that the HEAD commit was authored a year ago but only recently pushed to `master`. This is an unlikely scenario, but since I'm about to open similar pull requests in several repositories I thought it worth writing a pipeline which handles this edge case.

`git log` provides an `--author-date-order` flag which would obviate the need for `sort`, but I find `sort` clearer.

scott-christopher:
🍄

PR #214, opened 2016-05-08
Add S.keys, S.values and S.pairs

svozza:
Creating a type-safe version of `R.keys` was mentioned in #142 and now that `StrMap` is available I thought I'd road test it a bit. The implementation of this will probably provoke discussion, firstly because I'm using an ES5 only implementation: I find the Ramda implementation far too complicated for what should be a simple function. Secondly, I've chosen not to implement `S.values` in terms of `S.keys` because I don't think it makes sense to pay the penalty of type checking the object twice.

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,6 +27,8 @@
 //.
 //.     R.map(S.toUpper, S.head(words))
 //.
+//. Sanctuary is designed to work in Node.js and in ES5-compatible browsers.
+//.
 //. ## Types
 //.
 //. Sanctuary uses Haskell-like type signatures to describe the types of
@@ -2744,6 +2746,52 @@
           x = x[keys[idx]];
         }
         return filter(is(type), Just(x));
+      });
+
+  //# keys :: StrMap a -> Array String
+  //.
+  //. Returns the keys of the given string map, in arbitrary order.
+  //.
+  //. ```javascript
+  //. > S.keys({b: 2, c: 3, a: 1}).sort()
+  //. ['a', 'b', 'c']
+  //. ```
+  S.keys =
+  def('keys',
+      {},
+      [$.StrMap(a), $.Array($.String)],
+      Object.keys);
+
+  //# values :: StrMap a -> Array a
+  //.
+  //. Returns the values of the given string map, in arbitrary order.
+  //.
+  //. ```javascript
+  //. > S.values({a: 1, c: 3, b: 2}).sort()
+  //. [1, 2, 3]
+  //. ```
+  S.values =
+  def('values',
+      {},
+      [$.StrMap(a), $.Array(a)],
+      function(strMap) {
+        return Object.keys(strMap).map(function(key) { return strMap[key]; });
+      });
+
+  //# toPairs :: StrMap a -> Array (Pair String a)
+  //.
+  //. Returns the key–value pairs of the given string map, in arbitrary order.
+  //.
+  //. ```javascript
+  //. > S.pairs({b: 2, a: 1, c: 3}).sort()
+  //. [['a', 1], ['b', 2], ['c', 3]]
+  //. ```
+  S.pairs =
+  def('pairs',
+      {},
+      [$.StrMap(a), $.Array($.Pair($.String, a))],
+      function(strMap) {
+        return Object.keys(strMap).map(function(k) { return [k, strMap[k]]; });
       });
 
   //. ### Number
--- a/test/keys.js
+++ b/test/keys.js
@@ -1,0 +1,59 @@
+'use strict';
+
+var throws = require('assert').throws;
+
+var eq = require('./utils').eq;
+var errorEq = require('./utils').errorEq;
+var S = require('..');
+
+
+describe('keys', function() {
+
+  it('is a unary function', function() {
+    eq(typeof S.keys, 'function');
+    eq(S.keys.length, 1);
+  });
+
+  it('type checks its arguments', function() {
+    throws(function() { S.keys('xxx'); },
+           errorEq(TypeError,
+                   'Invalid value\n' +
+                   '\n' +
+                   'keys :: StrMap a -> Array String\n' +
+                   '        ^^^^^^^^\n' +
+                   '           1\n' +
+                   '\n' +
+                   '1)  "xxx" :: String\n' +
+                   '\n' +
+                   'The value at position 1 is not a member of ‘StrMap a’.\n'));
+
+    throws(function() { S.keys({a: '1', b: 2, c: '3'}); },
+           errorEq(TypeError,
+                   'Type-variable constraint violation\n' +
+                   '\n' +
+                   'keys :: StrMap a -> Array String\n' +
+                   '               ^\n' +
+                   '               1\n' +
+                   '\n' +
+                   '1)  "1" :: String\n' +
+                   '    2 :: Number, FiniteNumber, NonZeroFiniteNumber, Integer, ValidNumber\n' +
+                   '    "3" :: String\n' +
+                   '\n' +
+                   'Since there is no type of which all the above values are members, the type-variable constraint has been violated.\n'));
+  });
+
+  it("returns an array of the given object's own keys", function() {
+    eq(S.keys({}), []);
+    eq(S.keys({a: 1, b: 2, c: 3}).sort(), ['a', 'b', 'c']);
+  });
+
+  it('does not include prototype properties', function() {
+    var proto = {a: 1, b: 2};
+    var obj = Object.create(proto);
+    obj.c = 3;
+    obj.d = 4;
+
+    eq(S.keys(obj).sort(), ['c', 'd']);
+  });
+
+});
--- a/test/values.js
+++ b/test/values.js
@@ -1,0 +1,59 @@
+'use strict';
+
+var throws = require('assert').throws;
+
+var eq = require('./utils').eq;
+var errorEq = require('./utils').errorEq;
+var S = require('..');
+
+
+describe('values', function() {
+
+  it('is a unary function', function() {
+    eq(typeof S.values, 'function');
+    eq(S.values.length, 1);
+  });
+
+  it('type checks its arguments', function() {
+    throws(function() { S.values('xxx'); },
+           errorEq(TypeError,
+                   'Invalid value\n' +
+                   '\n' +
+                   'values :: StrMap a -> Array a\n' +
+                   '          ^^^^^^^^\n' +
+                   '             1\n' +
+                   '\n' +
+                   '1)  "xxx" :: String\n' +
+                   '\n' +
+                   'The value at position 1 is not a member of ‘StrMap a’.\n'));
+
+    throws(function() { S.values({a: '1', b: 2, c: '3'}); },
+           errorEq(TypeError,
+                   'Type-variable constraint violation\n' +
+                   '\n' +
+                   'values :: StrMap a -> Array a\n' +
+                   '                 ^\n' +
+                   '                 1\n' +
+                   '\n' +
+                   '1)  "1" :: String\n' +
+                   '    2 :: Number, FiniteNumber, NonZeroFiniteNumber, Integer, ValidNumber\n' +
+                   '    "3" :: String\n' +
+                   '\n' +
+                   'Since there is no type of which all the above values are members, the type-variable constraint has been violated.\n'));
+  });
+
+  it("returns an array of the given object's own values", function() {
+    eq(S.values({}), []);
+    eq(S.values({a: 1, b: 2, c: 3}).sort(), [1, 2, 3]);
+  });
+
+  it('does not include prototype values', function() {
+    var proto = {a: 1, b: 2};
+    var obj = Object.create(proto);
+    obj.c = 3;
+    obj.d = 4;
+
+    eq(S.values(obj).sort(), [3, 4]);
+  });
+
+});
--- a/test/pairs.js
+++ b/test/pairs.js
@@ -1,0 +1,63 @@
+'use strict';
+
+var throws = require('assert').throws;
+
+var eq = require('./utils').eq;
+var errorEq = require('./utils').errorEq;
+var S = require('..');
+
+
+describe('toPairs', function() {
+
+  var comparePairsAsc = function(a, b) {
+    return a[0] < b[0] ? -1 : a[0] > b[0] ? 1 : 0;
+  };
+
+  it('is a unary function', function() {
+    eq(typeof S.pairs, 'function');
+    eq(S.pairs.length, 1);
+  });
+
+  it('type checks its arguments', function() {
+    throws(function() { S.pairs('xxx'); },
+           errorEq(TypeError,
+                   'Invalid value\n' +
+                   '\n' +
+                   'pairs :: StrMap a -> Array (Pair String a)\n' +
+                   '         ^^^^^^^^\n' +
+                   '            1\n' +
+                   '\n' +
+                   '1)  "xxx" :: String\n' +
+                   '\n' +
+                   'The value at position 1 is not a member of ‘StrMap a’.\n'));
+
+    throws(function() { S.pairs({a: '1', b: 2, c: '3'}); },
+           errorEq(TypeError,
+                   'Type-variable constraint violation\n' +
+                   '\n' +
+                   'pairs :: StrMap a -> Array (Pair String a)\n' +
+                   '                ^\n' +
+                   '                1\n' +
+                   '\n' +
+                   '1)  "1" :: String\n' +
+                   '    2 :: Number, FiniteNumber, NonZeroFiniteNumber, Integer, ValidNumber\n' +
+                   '    "3" :: String\n' +
+                   '\n' +
+                   'Since there is no type of which all the above values are members, the type-variable constraint has been violated.\n'));
+  });
+
+  it('returns an array with the key value pairs of each property of the object', function() {
+    eq(S.pairs({}), []);
+    eq(S.pairs({a: 1, b: 2, c: 3}).sort(comparePairsAsc), [['a', 1], ['b', 2], ['c', 3]]);
+  });
+
+  it('does not include prototype properties', function() {
+    var proto = {a: 1, b: 2};
+    var obj = Object.create(proto);
+    obj.c = 3;
+    obj.d = 4;
+
+    eq(S.pairs(obj).sort(comparePairsAsc), [['c', 3], ['d', 4]]);
+  });
+
+});


davidchambers:
I'm excited to see these functions; they'll make great additions to the library!

I'd also like to provide `pairs :: StrMap a -> Array (Pair String a)`. I'd like `pairs`, `keys`, and `values` to guarantee consistent ordering. This would mean one could apply `keys` and `values` to a value independently and later zip the results, confident that the pairs will align correctly.

We could rely on the default behaviour of `Array#sort`.

> The implementation of this will probably provoke discussion, firstly because I'm using an ES5 only implementation: I find the Ramda implementation far too complicated for what should be a simple function.

The Ramda implementation is complicated because it works around bugs in certain old browser versions. We needn't do this. The ES3-compatible implementations needn't be complex:

```javascript
//  keys :: StrMap a -> Array String
var keys = function(strMap) {
  var ks = [];
  for (var k in strMap) if (has(k, strMap)) ks.push(k);
  return ks.sort();
}

//  values :: StrMap a -> Array a
var values = function(strMap) {
  var ks = keys(strMap);
  var vs = [];
  for (var idx = 0; idx < ks.length; idx += 1) {
    vs.push(strMap[ks[idx]]);
  }
  return vs;
};

//  pairs :: StrMap a -> Array (Pair String a)
var pairs = function(strMap) {
  var ks = keys(strMap);
  var kvs = [];
  for (var idx = 0; idx < ks.length; idx += 1) {
    kvs.push([ks[idx], strMap[ks[idx]]]);
  }
  return kvs;
}
```

> I don't think it makes sense to pay the penalty of type checking the object twice.

Type checking is currently very slow. I'm confident we can reduce the cost by at least 90%. Once one of us spends a few days making the type checking less naive we'll hopefully be able to stop worrying about paying the cost twice.

kedashoe:
> I'd like pairs, keys, and values to guarantee consistent ordering.

What about using @svozza's implementations for these names, and also adding `sortedKeys`, `sortedValues`, `sortedPairs` for when consistent ordering is needed? Iterating over a collection 3-ish times to get `values` feels like a bit much to me.

And what about a simple `for (key in obj)` to get the values in `values`?

davidchambers:
I'm open to providing "fast" versions of these functions as well, but I'd like to name them `keys_`, `values_`, and `pairs_`. I'd like the "nice" versions of the functions to have the nice names.

The other option is to sidestep the ordering problem by defining Set and Bag types. We could then define:

```haskell
keys   :: StrMap a -> Set String
values :: StrMap a -> Bag a
pairs  :: StrMap a -> Set (Pair String a)
```

I'd like to stick with arrays, though, so we could guarantee this property:

    forall m :: StrMap a. R.zip(S.keys(m), S.values(m)) = S.pairs(m)

kedashoe:
For me the faster versions make sense with the plain names because

1. I almost never care about the order of these functions
2. It seems more natural to me to think "do I care about order here" than "do I care about performance here"

But I see the appeal of ordered results to these functions.. I'm okay either way :shrug:

davidchambers:
I might be on the wrong side of the fence here. :)

What do others think? @Avaq? @benperez? @scott-christopher? I'm also interested in your thoughts on the matter, @CrossEye, since we've discussed this in Ramda threads several times.

svozza:
I've ES3-ified the implementation. Do you want me to add `pairs` to this PR?

Regarding sorting the keys, it seems odd to impose an the order of keys in a map, if the user really cares she should just sort the result herself.

davidchambers:
> I've ES3-ified the implementation.

Thanks. We're still relying on `Array#map` in `values`, though.

> Do you want me to add `pairs` to this PR?

If you'd like to do so, sure! It's up to you. :)

> Regarding sorting the keys, it seems odd to impose an order of keys in a map, if the user really cares she should just sort the result herself.

Sorting the array returned by `keys` is not so important: as you rightly point out the caller can do this after the fact. She cannot, though, sort the values *by their keys* after the fact, as the keys are gone.

The question is whether to support diverging/converging pipelines such as this:

```javascript
                          {x: 0, y: 42}
                               / \
                              /   \
                             /     \
                            /       \
                           /         \
                          /           \
         ['x', 'y'] <=  keys        values  => [0, 42]
                         |             |
                         |             |
                         |             |
                         |             |
                         |             |
                         |             |
     ['X', 'Y'] <=  map(toUpper)   map(inc)  => [1, 43]
                          \           /
                           \         /
                            \       /
                             \     /
                              \   /
                               \ /
                               zip  => [['X', 1], ['Y', 43]]
                                |
                                |
                                |
                                |
                                |
                                |
                            fromPairs  => {X: 1, Y: 43}
```

If we decide not to guarantee consistent ordering, the user can of course use this approach instead:

```javascript
                          {x: 0, y: 42}
                                |
                                |
                                |
                                |
                              pairs  => [['x', 0], ['y', 42]]
                                |
                                |
                                |
                                |
                 map(over(lensIndex(0), toUpper))  => [['X', 0], ['Y', 42]]
                                |
                                |
                                |
                                |
                   map(over(lensIndex(1), inc))  => [['X', 1], ['Y', 43]]
                                |
                                |
                                |
                                |
                            fromPairs  => {X: 1, Y: 43}
```

Both approaches seem fine to me. If we support both, users are free to use whichever feels more natural. Forcing users to *think* about such flows in a certain way doesn't appeal to me (though I realize Sanctuary is an opinionated library which forces users to think a certain way in many other contexts).

Does this clarify my position, @svozza? As I say, I might be on the wrong side of this disagreement. ;)

index.js
```diff
@@ -362,6 +364,14 @@
         }
       });
 
+  //  keys :: StrMap a -> Array String
+  var keys = function(strMap) {
+    if (typeof Object.keys === 'function') return Object.keys(strMap);
```
davidchambers: Is this for performance? My default position is to minimize branching, but I'm happy to introduce branching if there's a significant performance difference.

index.js
```diff
@@ -362,6 +364,14 @@
...
+    var ks = [];
+    for (var k in strMap) if (hasOwnProperty(k, strMap)) ks.push(k);
```
davidchambers: I think you mean `hasOwnProperty.call(k, strMap)`.

index.js
```diff
@@ -2683,6 +2693,38 @@
         return filter(is(type), Just(x));
       });
 
+  //# keys :: StrMap a -> Array String
+  //.
+  //. Returns a list containing the keys of the supplied StrMap.
+  //.
+  //. ```javascript
+  //. > S.keys({a: 1, b: 2, c:3})
```
davidchambers: `s/c:3/c: 3/`

index.js
```diff
@@ -2683,6 +2693,38 @@
...
+      {},
+      [$.StrMap(a), $.Array($.String)],
+      keys);
+
+  //# values :: StrMap a -> Array a
+  //.
+  //. Returns a list containing the values of the supplied StrMap.
+  //.
+  //. ```javascript
+  //. > S.values({a: 1, b: 2, c:3})
```
davidchambers: `s/c:3/c: 3/`

index.js
```diff
@@ -2683,6 +2693,38 @@
         return filter(is(type), Just(x));
       });
 
+  //# keys :: StrMap a -> Array String
+  //.
+  //. Returns a list containing the keys of the supplied StrMap.
```
davidchambers: `s/a list containing/an array of/`

index.js
```diff
@@ -2683,6 +2693,38 @@
...
+  //. ```
+  S.keys =
+  def('keys',
+      {},
+      [$.StrMap(a), $.Array($.String)],
+      keys);
+
+  //# values :: StrMap a -> Array a
+  //.
+  //. Returns a list containing the values of the supplied StrMap.
```
davidchambers: `s/a list containing/an array of/`

test/keys.js
```diff
@@ -0,0 +1,58 @@
...
+                   '\n' +
+                   '1)  "1" :: String\n' +
+                   '    2 :: Number, FiniteNumber, NonZeroFiniteNumber, Integer, ValidNumber\n' +
+                   '    "3" :: String\n' +
+                   '\n' +
+                   'Since there is no type of which all the above values are members, the type-variable constraint has been violated.\n'));
+  });
+
+  it("returns an array of the given object's own keys", function() {
+    eq(S.keys({a: 1, b: 2, c: 3}).sort(), ['a', 'b', 'c']);
```
davidchambers: It's worth noting that sorted keys would be useful here. ;)

index.js
svozza: I remember reading before that `Object.keys` is heavily optimised but I can't for the life of me remember where.

index.js

index.js

index.js

index.js

index.js

test/keys.js
svozza: I'm not sure how it would have made it any easier to copy and paste this test from Ramda's. ;)

svozza:
I can see where you're coming from with the sorted keys but to me I think if you care about how the elements line up your far more likely to work on them as pairs, as per your second example and probably more like this:

```js
const f = (pair) => [R.toUpper(pair[0]), R.inc(pair[1])];

map(f, toPairs({x: 1, y: 2}));
```

I have a slight preference for leaving the resulting arrays unsorted as iterating over the list three times (if you include the `hasOwnProperty` check) seems like overkill but I'm happy to go with the consensus.

rjmk:
On the one hand, I think that it would be nice if Sanctuary avoided non-deterministic functions. On the other hand, it seems odd for Sanctuary to impose arbitrary structure on values. I think I like the `Set` / `Bag` solution most. I think if people want the values ordered by the keys, but don't want to `map

index.js
Avaq: I would advice against optimizing performance before

1. It's established that there is a performance issue
2. It's established that the optimization fixes it
3. The optimized code [behaves exactly the same](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys#Polyfill) as the non-optimized code

That said, a quick performance benchmark on my local machine shows that `Object.keys` is about *5 times as fast* as `for ... in hasOwn`, so that certainly crosses out `2` from my list. ;)

I'm a little worried about the behavior differences between `Object.keys` and `for ... in` as illustrated by the MDN suggested polyfill. Most of it seems to be related to some bug where `for ... in` misses some properties that *should* have been enumerated though. It probably won't affect us.

My vote goes in favor of the `Object.keys` branch.

davidchambers: If we're to branch, would it make sense to do the check once only?

```javascript
var keys = typeof Object.keys === 'function' ? Object.keys : function(strMap) {…};
```

svozza: :zap: Although I'm not sure if I've got the formatting right...

davidchambers: > I'm not sure if I've got the formatting right

LGTM!

svozza: Now, what do I about the code coverage? None of the test environments we run against lack `Object.keys` but we really ought to test this branch of code. I tried just overwriting the `Object` prototype version of `keys` with `null` in a test but it caused all manner of other failures.

index.js
```diff
@@ -2683,6 +2694,41 @@
...
+  def('values',
+      {},
+      [$.StrMap(a), $.Array(a)],
+      function(strMap) {
+        var ks = keys(strMap);
+        var vs = [];
+        for (var idx = 0; idx < ks.length; idx += 1) {
+          vs.push(strMap[ks[idx]]);
+        }
+        return vs;
```
davidchambers: Since we're thinking about performance, I wonder whether reusing the keys array would help:

```javascript
var xs = keys(strMap);
for (var idx = 0; idx < xs.length; idx += 1) xs[idx] = strMap[xs[idx]];
return xs;
```

index.js
davidchambers: I'd try this approach:

```javascript
if (typeof Object.keys === 'function') {
  var Object$keys = Object.keys;
  delete Object.keys;
  // [Test fallback code...]
  Object.keys = Object$keys;
}
```

svozza: Yeah, that's what I tried but the problem is that Mocha runs the tests in parallel so when you mess with the `Object` prototype it affects all the other tests.

davidchambers: > Mocha runs the tests in parallel

I don't think that's the case. The problem is that Node itself uses `Object.keys`, so deleting it can cause problems. For example:

```javascript
> delete Object.keys
true
> console.log({})
TypeError: Object.keys is not a function
```

I don't have a good solution to this problem. The *simplest* option is to remove the "fast" path, but I'm open to other suggestions. :)

svozza: I'd rather remove the slow path tbh.

scott-christopher: If you really want to test this branch and don't mind restricting your test suite to run in Node only, you could potentially execute the test in a new VM context.

```js
vm.runInNewContext('console.log(Object.keys)', R.dissocPath(['Object', 'keys'], this))
```

That said, how desirable is ES3 support?

davidchambers: > That said, how desirable is ES3 support?

Dropping support for ES3 seems popular. I'm happy to do so, though I'd like to add a section to the readme to document which environments we target. The description could be as brief as _Sanctuary is designed to work in Node and in ES5-compatible browsers._

index.js
svozza: Good idea! :zap:

index.js

svozza:
`S.toPairs` added!

davidchambers:
> `S.toPairs` added!

Very exciting! I'm not sure about the name, though. In [Data.HashMap][1] the three functions are named [`keys`][2], [`elems`][3], and [`assocs`][4]/[`toList`][5] (I don't know why there are two names for this one).

Since all the Sanctuary functions live in a single namespace `S.toArray` is not an option. I like `S.pairs`, though, rather than than `S.toPairs`. It matches `S.keys` and `S.values` (which are not `S.toKeys` and `S.toValues` after all). What do you think?


[1]: https://hackage.haskell.org/package/hashmap-1.3.0.1/docs/Data-HashMap.html
[2]: https://hackage.haskell.org/package/hashmap-1.3.0.1/docs/Data-HashMap.html#v:keys
[3]: https://hackage.haskell.org/package/hashmap-1.3.0.1/docs/Data-HashMap.html#v:elems
[4]: https://hackage.haskell.org/package/hashmap-1.3.0.1/docs/Data-HashMap.html#v:assocs
[5]: https://hackage.haskell.org/package/hashmap-1.3.0.1/docs/Data-HashMap.html#v:toList

svozza:
Yep, I prefer `S.pairs`, the only reason I had it the other way was because I was following Ramda.

index.js
```diff
@@ -2746,6 +2748,59 @@
         return filter(is(type), Just(x));
       });
 
+  //# keys :: StrMap a -> Array String
+  //.
+  //. Returns an array of the keys of the supplied StrMap.
```
davidchambers: We could omit "an array of" as this is indicated by the type signature.

Let's append "in arbitrary order".

index.js
```diff
@@ -2746,6 +2748,59 @@
...
+  //.
+  //. ```javascript
+  //. > S.keys({a: 1, b: 2, c: 3})
+  //. ['a', 'b', 'c']
```
davidchambers: Let's add `.sort()` to the input, both to ensure determinism and to reinforce the comment about arbitrary ordering.

index.js
```diff
@@ -2746,6 +2748,59 @@
...
+  //. ```
+  S.keys =
+  def('keys',
+      {},
+      [$.StrMap(a), $.Array($.String)],
+      function(strMap) { return Object.keys(strMap); });
```
davidchambers: Is this anonymous function necessary? It seems we could use `Object.keys` directly.

index.js
```diff
@@ -2746,6 +2748,59 @@
...
+
+  //# values :: StrMap a -> Array a
+  //.
+  //. Returns an array of the values of the supplied StrMap.
```
davidchambers: My comments on the `keys` description apply here as well.

index.js
```diff
@@ -2746,6 +2748,59 @@
...
+  //.
+  //. ```javascript
+  //. > S.values({a: 1, b: 2, c: 3})
+  //. [1, 2, 3]
```
davidchambers: Let's `.sort()` here too.

index.js
```diff
@@ -2746,6 +2748,59 @@
...
+  def('values',
+      {},
+      [$.StrMap(a), $.Array(a)],
+      function(strMap) {
+        var xs = Object.keys(strMap);
+        for (var idx = 0; idx < xs.length; idx += 1) xs[idx] = strMap[xs[idx]];
+        return xs;
+      });
+
+  //# toPairs :: StrMap a -> Array Pair String a
```
davidchambers: We need parens: `Array (Pair String a)`. As written Array appears to be a ternary type. ;)

index.js
```diff
@@ -2746,6 +2748,59 @@
...
+  //.
+  //. Returns an array of key, value pairs corresponding to an object's
+  //. own properties.
```
davidchambers: I like to write "key–value pairs" with an [en dash][1]. I suggest:

_Returns the key–value pairs of the given string map, in arbitrary order._


[1]: http://www.thepunctuationguide.com/en-dash.html

index.js
```diff
@@ -2746,6 +2748,59 @@
...
+  //.
+  //. ```javascript
+  //. > S.toPairs({a: 1, b: 2, c: 3})
+  //. [['a', 1], ['b', 2], ['c', 3]]
```
davidchambers: We should probably `.sort()` here too, though relying on type coercion feels a little dirty.

index.js
```diff
@@ -2746,6 +2748,59 @@
...
+  S.toPairs =
+  def('toPairs',
+      {},
+      [$.StrMap(a), $.Array($.Pair($.String, a))],
+      function(strMap) {
+        var xs = Object.keys(strMap);
+        for (var idx = 0; idx < xs.length; idx += 1) {
+          xs[idx] = [xs[idx], strMap[xs[idx]]];
+        }
+        return xs;
```
davidchambers: How about using `Array#map`?

```javascript
return Object.keys(strMap).map(function(k) { return [k, strMap[k]]; });
```

test/toPairs.js
```diff
@@ -0,0 +1,64 @@
...
+var errorEq = require('./utils').errorEq;
+var S = require('..');
+
+
+describe('toPairs', function() {
+
+  var comparePairsAsc = function(a, b) {
+    if (a[0] > b[0]) return 1;
+    if (a[0] < b[0]) return -1;
+    return 0;
```
davidchambers: I'd prefer `a[0] < b[0] ? -1 : a[0] > b[0] ? 1 : 0` but I'm aware that not everyone is as fond of ternary expressions as I am. :)

test/keys.js
davidchambers: Let's also test `S.keys({})`.

test/values.js
```diff
@@ -0,0 +1,58 @@
...
+                   '\n' +
+                   '1)  "1" :: String\n' +
+                   '    2 :: Number, FiniteNumber, NonZeroFiniteNumber, Integer, ValidNumber\n' +
+                   '    "3" :: String\n' +
+                   '\n' +
+                   'Since there is no type of which all the above values are members, the type-variable constraint has been violated.\n'));
+  });
+
+  it("returns an array of the given object's own values", function() {
+    eq(S.values({a: 1, b: 2, c: 3}).sort(), [1, 2, 3]);
```
davidchambers: Let's also test `S.values({})`.

test/toPairs.js
```diff
@@ -0,0 +1,64 @@
...
+                   '\n' +
+                   '1)  "1" :: String\n' +
+                   '    2 :: Number, FiniteNumber, NonZeroFiniteNumber, Integer, ValidNumber\n' +
+                   '    "3" :: String\n' +
+                   '\n' +
+                   'Since there is no type of which all the above values are members, the type-variable constraint has been violated.\n'));
+  });
+
+  it('returns an array with the key value pairs of each property of the object', function() {
+    eq(S.toPairs({a: 1, b: 2, c: 3}).sort(comparePairsAsc), [['a', 1], ['b', 2], ['c', 3]]);
```
davidchambers: Let's also test `S.toPairs({})`.

index.js
svozza: Ah yes, I wasn't sure about that one but I knew it didn't look right.

index.js
svozza: Another way to do this that would be to list the elements in each of the functions' examples in an obviously out of order sequence. That way we don't clutter the docs with things that really aren't related to the functions:

```js
S.keys({a: 1, b: 2, c: 3}) // => [c, b, a]
S.values({a: 1, b: 2, c: 3}) // => [3, 2, 1]
S.toPairs({a: 1, b: 2, c: 3}) //=> [['b', 2], ['a', 1], ['c', 3]]

File: index.js (hunk 17 above)
Comment (svozza) [inline index.js]: Yeah, that's definitely easier to read.

File: test/toPairs.js (hunk 18 above)
Comment (svozza) [inline test/toPairs.js]: No, I'm good with that.

File: index.js (hunk 11 above)
Comment (svozza) [inline index.js]: Of course, that was a hangover from the previous branching solution.

**Comment by svozza** [comment]
:zap: Think I've everything except adding `.sort()` to the docs.

File: index.js (hunk 16 above)
Comment (davidchambers) [inline index.js]: The root problem is that we're using an ordered collection for something with no inherent order. When comparing two arrays for equality, order is of course significant. To ensure the doctests pass, we have two options:

  - rely on property enumeration order; or
  - use `Array#sort`.

Thinking about this again, I believe the best approach is to jumble the keys in the *input*:

```javascript
> S.keys({c: 3, a: 1, b: 2}).sort()
['a', 'b', 'c']

> S.values({c: 3, a: 1, b: 2}).sort()
[1, 2, 3]

> S.pairs({c: 3, a: 1, b: 2}).sort()
[['a', 1], ['b', 2], ['c', 3]]
```

index.js
```diff
@@ -2746,6 +2748,54 @@
...
+  //. [1, 2, 3]
+  //. ```
+  S.values =
+  def('values',
+      {},
+      [$.StrMap(a), $.Array(a)],
+      function(strMap) {
+        var xs = Object.keys(strMap);
+        for (var idx = 0; idx < xs.length; idx += 1) xs[idx] = strMap[xs[idx]];
+        return xs;
```
davidchambers: How about using `Array#map` here too?

```javascript
return Object.keys(strMap).map(function(k) { return strMap[k]; });
```

svozza: Ha. We've come full circle to my original implementation!

index.js
svozza: Of course, I'd forgotten about the doc tests.

:zap:

index.js
davidchambers: Yep! Now that we're embracing ES5 we can take advantage of its conveniences. :)

svozza: Yup, all good!

davidchambers:
:deciduous_tree: Very nice pull request, Stefano!

index.js
```diff
@@ -2746,6 +2748,52 @@
...
+  //. ```
+  S.values =
+  def('values',
+      {},
+      [$.StrMap(a), $.Array(a)],
+      function(strMap) {
+        return Object.keys(strMap).map(function(key) { return strMap[key]; });
+      });
+
+  //# toPairs :: StrMap a -> Array (Pair String a)
```
davidchambers: `s/toPairs/pairs/`

I'll fix this shortly.

davidchambers: #226

PR #215, opened 2016-05-09
either: add S.eitherToMaybe

### LINKED ISSUE #203: eitherToMaybe :: Either a b -> Maybe b
@kedashoe added [`maybeToEither`][1] in #76. It would be nice to provide a function for going in the other direction. The implementation is not complex, and already appears in the [test suite][2]:

> ```javascript
> //  eitherToMaybe :: Either a b -> Maybe b
> var eitherToMaybe = S.either(S.K(S.Nothing()), S.Just);
> ```

[1]: http://sanctuary.js.org/#maybeToEither
[2]: https://github.com/sanctuary-js/sanctuary/blob/v0.10.0/test/index.js#L117-L118

davidchambers:
Closes #203

### DIFF
--- a/index.js
+++ b/index.js
@@ -1302,7 +1302,7 @@
   //. Binary version of [`encase`](#encase).
   //.
   //. See also [`encase2_`](#encase2_).
-  S.encase2 =
+  var encase2 = S.encase2 =
   def('encase2',
       {},
       [$.Function, a, b, $Maybe(c)],
@@ -1327,7 +1327,7 @@
             return f_(x, y);
           };
         };
-        return S.encase2(f, x, y);
+        return encase2(f, x, y);
       });
 
   //# encase3 :: (a -> b -> c -> d) -> a -> b -> c -> Maybe d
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -1362,7 +1362,29 @@
...
           };
         };
-        return S.encase3(f, x, y, z);
+        return encase3(f, x, y, z);
+      });
+
+  //# maybeToEither :: a -> Maybe b -> Either a b
+  //.
+  //. Takes a value of any type and a Maybe, and returns an Either.
+  //. If the second argument is a Nothing, a Left containing the first
```
kedashoe: Might not be worth it for the sake of consistency throughout the docs, but when the types of arguments are different, it is easier for me to follow if the type is referenced rather than the argument position, ie

`If the Maybe is a Nothing, a Left containing the first argument is returned.`

Obviously since `is a Nothing` immediately follows `second argument`, I should know it is a Maybe I'm dealing with, but when I hit "second argument" I always pause.. think "second argument..??" jump back to the signature.. then continue on :)

davidchambers: "The Maybe" is *slightly* less precise than "the second argument", since the first argument could also be a Maybe, but I'm happy to make your suggested change.

davidchambers: Another possibility:

_Converts a Maybe to an Either. A Nothing becomes a Left (containing the first argument); a Just becomes a Right._

What do you think?

kedashoe: 👍

PR #216, opened 2016-05-11
fantasy land: add Fantasy Land functions

### LINKED ISSUE #178: (pointfree) chain: Chain f => (a -> f a) -> f a -> f a
I am missing the pointfree `chain` function in this library.

I have several functions of signature `Functor f => a -> f a`, which I would like to compose using `pipe`.
I have written the following conversion to pointfree of `chain` to allow this.

``` javascript
// Functor f => (a -> f a) -> f a -> f a
const chain = func => x => x.chain(func);
```

Now I can write

``` javascript
// Functor f => a -> f a
const composed = pipe([func1, chain(func2), chain(func3)]);
```

Maybe there is a different intended way of doing this with this library? If not, I think the simple `chain` function would be a good addition.

### LINKED ISSUE #252: mention Fantasy Land in readme
We link to the [Fantasy Land spec](https://github.com/fantasyland/fantasy-land) in several places. It would be nice to include a section introducing Fantasy Land, explaining Sanctuary's relationship with it, and mentioning that the methods Sanctuary defines are for Fantasy Land compatibility and are not intended for direct use.

davidchambers:
Closes #178; closes #252

This pull request adds the following functions:

  - `alt :: Alt f => f a -> f a -> f a`
  - `ap :: Apply f => f (a -> b) -> f a -> f b` (`ap` replaces `S`)
  - `apFirst :: Apply f => f a -> f b -> f a`
  - `apSecond :: Apply f => f a -> f b -> f b`
  - `bimap :: Bifunctor f => (a -> b) -> (c -> d) -> f a c -> f b d`
  - `chain :: Chain m => (a -> m b) -> m a -> m b`
  - `chainRec :: ChainRec m => TypeRep m -> (a -> Either (m b) (m b)) -> a -> m b`
  - `empty :: Monoid a => TypeRep a -> a`
  - `equals :: Setoid a => a -> a -> Boolean`
  - `extend :: Extend w => (w a -> b) -> w a -> w b`
  - `extract :: Comonad w => w a -> a`
  - `filter :: (Applicative f, Foldable f, Monoid (f a)) => (a -> Boolean) -> f a -> f a`
  - `filterM :: (Monad m, Monoid (m a)) => (a -> Boolean) -> m a -> m a`
  - `join :: Chain m => m (m a) -> m a`
  - `map :: Functor f => (a -> b) -> f a -> f b` (`map` replaces `B` and `lift`)
  - `of :: Applicative f => TypeRep f -> a -> f a`
  - `promap :: Profunctor p => (a -> b) -> (c -> d) -> p b c -> p a d`
  - `sequence :: (Applicative f, Traversable t) => TypeRep f -> t (f a) -> f (t a)`
  - `toString :: Any -> String`
  - `traverse :: (Applicative f, Traversable t) => TypeRep f -> (a -> f b) -> t a -> f (t b)`
  - `zero :: Plus f => TypeRep f -> f a`

Original description:

> sanctuary-js/sanctuary-type-classes#2 hasn't yet generated the excitement I expected, so I thought I'd open a pull request to add all the Fantasy Land functions except `equals` and `of` while _removing_ code.
>
> :rainbow:


svozza:
Ah so this is where old `map` functions go to die. ;) Looks great. I wasn't sure how to handle those test failures in `S.match` either, I presume it's because of the weird object/array that you get back from using the native regex function in JS?

index.js
```diff
@@ -151,14 +151,16 @@
...
   } else if (typeof define === 'function' && define.amd != null) {
-    define(['ramda', 'sanctuary-def'], f);
+    define(['ramda', 'sanctuary-type-classes', 'sanctuary-def'], f);
   } else {
-    self.sanctuary = f(self.R, self.sanctuaryDef);
+    self.sanctuary = f(self.R, self.sanctuaryTypeClasses, self.sanctuaryDef);
   }
 
-}(function(R, $) {
+}(function(R, λ, $) {
```
svozza: What's the quickest way to type this symbol without having to resort to copy and pate?

davidchambers:
> I wasn't sure how to handle those test failures in `S.match` either, I presume it's because of the weird object/array that you get back from using the native regex function in JS?

The problem actually lies with optional non-capturing groups. Since there's no type of which both `'foo'` and `undefined` are members, `['foo', undefined]` is heterogeneous and therefore considered invalid.

```javascript
S.map(S.I, /(foo)?(bar)?/.exec('foobar'));
// => ['foobar', 'foo', 'bar']

S.map(S.I, /(foo)?(bar)?/.exec('foo'));
//
//   map :: (Functor a, Functor b) => Function -> a -> b
//                                                ^
//                                                1
//
//   1)  ["foo", "foo", undefined, "index": 0, "input": "foo"] :: Array ???
//
```

The problematic expression is `map(toMaybe, match)`. The solution is to replace this with a `for` loop (`match.map(toMaybe)` wouldn't work since `toMaybe` would complain about receiving three arguments).
The fact that we must work around the type system when bringing inconsistent values into the consistent world doesn't worry me – in fact, it's a sign that the type system is doing its job. :)

index.js
davidchambers: I'm not sure. I placed it in the `l` register in Vim so I was able to type <kbd>"</kbd> <kbd>l</kbd> <kbd>p</kbd> to insert it.

We should probably switch to an identifier that's easier to type. `_`, perhaps? I like one-character identifiers for dependencies we reference dozens of times. :)

kedashoe: `L` for lambda?

svozza:
> The fact that we must work around the type system when bringing inconsistent values into the consistent world doesn't worry me – in fact, it's a sign that the type system is doing its job. :)

I agree completely, it's better to deal with these sort of oddities in the safety of a functions scope rather than having them leak out, even if it's at the cost of the implementations elegance.

index.js
svozza: Yeah, `L` sounds good.

index.js
```diff
@@ -502,6 +506,448 @@
...
+  //.
+  //. > S.join([[[1, 2, 3]]])
+  //. [[1, 2, 3]]
+  //.
+  //. > S.join(S.Just(S.Just(1)))
+  //. S.Just(1)
+  //. ```
+  S.join = def('join', {m: [Z.Chain]}, [m(m(a)), m(a)], Z.join);
+
+  //# chainRec :: ChainRec m => TypeRep m -> (a -> Either (m b) (m b)) -> a -> m b
```
davidchambers: @safareli, can you explain to me why the type of the third argument is `a` rather than `m a`? It seems to me that `m a` would be more consistent with the type of `chain`. Furthermore, `m a` would obviate the need for a type representative. Could we have this instead?

```purescript
chain    :: Chain m    => (a -> m b)                -> m a -> m b
chainRec :: ChainRec m => (a -> Either (m b) (m b)) -> m a -> m b
```

index.js
```diff
@@ -502,6 +506,448 @@
...
+  //. {x: 1, y: 3, z: 4}
+  //.
+  //. > S.concat(S.Just([1, 2, 3]), S.Just([4, 5, 6]))
+  //. Just([1, 2, 3, 4, 5, 6])
+  //. ```
+  S.concat = def('concat', {a: [Z.Semigroup]}, [a, a, a], Z.concat);
+
+  //# empty :: Monoid a => TypeRep a -> a
+  //.
+  //. Alias of [`Z.empty`][].
```
Avaq: Might be misleading to call it an alias, since `Z` does no type-checking.

davidchambers: Good point. Perhaps something like this would be better:

_[`def`][1]'d version of [`Z.empty`][2]._


[1]: https://github.com/sanctuary-js/sanctuary-def/tree/v0.9.0
[2]: https://github.com/sanctuary-js/sanctuary-type-classes/tree/v2.0.1#empty

Avaq: I' not sure about the word `def'd`. Maybe use something like "_[`Type safe`][1] version of [`Z.empty`][2]._"?

[1]: https://github.com/sanctuary-js/sanctuary-def/tree/v0.9.0#sanctuary-def
[2]: https://github.com/sanctuary-js/sanctuary-type-classes/tree/v2.0.1#empty

davidchambers: That sounds better. Thanks for the suggestion.

index.js
syaiful6: looking ```chainRec :: ChainRec m => (a -> Either (m b) (m b)) -> m a -> m b``` it look like the function only allowed to return same ```m b``` both Left and Right of Either, is this really the case?
if so, this not allowed?
```chainRec(i => i ? 1000 : Identity(Left(i + 1)) : Identity(Right('Boom')), 0)``` and @safareli https://github.com/safareli/free/blob/master/src/free.js#L121 also not allowed since that value inside done and next contains different type, that's it the next contains free monad and done not.

Comment (davidchambers) [inline index.js]: > it look like [`chainRec` is] only allowed to return same `m b` both Left and Right of Either, is this really the case?

I don't know, to be honest. I find the ChainRec type confusing. I'll let @safareli clarify this for us. :)

File: index.js [hunk 4]
```diff
@@ -502,6 +506,455 @@
...
+  //. Curried version of [`Z.bimap`][].
+  //.
+  //. ```javascript
+  //. > S.bimap(S.toUpper, Math.sqrt, S.Left('foo'))
+  //. Left('FOO')
+  //.
+  //. > S.bimap(S.toUpper, Math.sqrt, S.Right(64))
+  //. Right(8)
+  //. ```
+  S.bimap =
```
svozza: Does that mean we don't need `S.either` anymore?

index.js
safareli: type from sanctuary-type-classes is:
```hs
chainRec :: ChainRec m => (TypeRep m, (a -> c, b -> c, a) -> m c, a) -> m b
```

If we want to use `Either`, then it should be:
```hs
chainRec :: (TypeRep m, a -> m (Either a b), a) -> m b
```
in recent version of [purescript-tailRec](https://github.com/purescript/purescript-tailrec/blob/master/src/Control/Monad/Rec/Class.purs#L56) `Step` is used instead of `Either`, like this:
```hs
data Step a b = Loop a | Done b
tailRecM :: forall a b. (a -> m (Step a b)) -> a -> m b
```
---

> ... would be more consistent with the type of `chain` ...

type of `unfold` is:
```hs
:: (b -> Maybe (Pair a b)) ->       b -> Array a
```
instead of
```hs
:: (b -> Maybe (Pair a b)) -> Array b -> Array a
```

the `chainRec` looks more like unfold rather than `chain`:
```hs
unfold   :: (a -> Maybe (Pair b a)) -> a -> Array b
chainRec :: (a ->     m (Step a b)) -> a -> m     b
```

also if seed value must be of type `m a` then it means that in order for `chainRec` to be useful for arbitrary some seed `a` then the structure should also have a way to to lift  the  `a` to  `m a` i.e. `of` method, which makes the Chain a Monad. by not asking for `m a` the `chainRec` could be useful for `a`(any) value rather than `m a`. and why ask for more when it is not needed. if someone has m a  one could call chain and get `a` and in there do `chainRec`

so if we use Either it type should be
```hs
chainRec :: ChainRec m => TypeRep m -> (a -> m (Either a b)) -> a -> m b
```
and if i'm correct
```js
def('chainRec',
      {m: [Z.ChainRec]},
      [TypeRep($.TypeVariable('m')), Fn(a, m($Either(a, b))), a, m(b)],
...
```

index.js
Avaq: @svozza `S.either` is for unpacking an `Either a b`, it returns the return value of the mapper. `S.bimap` returns a new `Either` of the value returned by the mapper.

svozza: Oh yes, of course, `either` is similar to `fromMaybe`.

davidchambers: Indeed, although it's most similar to `maybe` and `maybe_`:

```purescript
maybe  ::        b  -> (a -> b) -> Maybe a -> b
maybe_ :: (() -> b) -> (a -> b) -> Maybe a -> b
either ::  (a -> c) -> (b -> c) -> Either a b -> c
```

`fromMaybe` and `fromEither` are also closely related:

```purescript
fromMaybe  :: a -> Maybe a -> a
```

test/ap.js
```diff
@@ -0,0 +1,21 @@
...
+  eq(typeof S.ap, 'function');
+  eq(S.ap.length, 2);
+  eq(S.ap.toString(), 'ap :: Apply f => f (a -> b) -> f a -> f b');
+
+  eq(S.ap([], []), []);
+  eq(S.ap([], [1, 2, 3]), []);
+  eq(S.ap([S.inc], []), []);
+  eq(S.ap([S.inc], [1, 2, 3]), [2, 3, 4]);
+  eq(S.ap([S.dec, Math.sqrt], [1, 4, 9]), [0, 3, 8, 1, 2, 3]);
+  eq(S.ap(S.Just(S.inc), S.Just(42)), S.Just(43));
```
safareli: maybe also add a failing case?
```diff
+  eq(S.ap(S.Nothing, S.Just(42)), S.Nothing);
```

test/filterM.js
```diff
@@ -0,0 +1,14 @@
...
+
+var eq = require('./internal/eq');
+
+
+test('filterM', function() {
+
+  eq(typeof S.filterM, 'function');
+  eq(S.filterM.length, 2);
+  eq(S.filterM.toString(), 'filterM :: (Monad m, Monoid m) => (a -> Boolean) -> m a -> m a');
+
```
safareli: I think `filter` tests could be reused here as well as both Array  and Maybe are Monads

test/internal/Compose.js
```diff
@@ -23,11 +25,11 @@ module.exports = function Compose(F) {
     };
 
     ComposeFG.prototype[FL.map] = function(f) {
-      return ComposeFG(Z.map(function(y) { return Z.map(f, y); }, this.value));
+      return ComposeFG(Z.map(S.map(f), this.value));
     };
 
     ComposeFG.prototype[FL.ap] = function(other) {
-      return ComposeFG(Z.ap(Z.map(function(u) { return function(y) { return Z.ap(u, y); }; }, other.value), this.value));
+      return ComposeFG(Z.ap(Z.map(S.ap, other.value), this.value));
```
safareli: Is the reason to mix `Z` and `S` the fact that Z has uncurried functions?

test/sequence.js
```diff
@@ -0,0 +1,14 @@
...
+
+var eq = require('./internal/eq');
+
+
+test('sequence', function() {
+
+  eq(typeof S.sequence, 'function');
+  eq(S.sequence.length, 2);
+  eq(S.sequence.toString(), 'sequence :: (Applicative f, Traversable t) => TypeRep f -> t (f b) -> f (t b)');
+
```
safareli: TK?

test/promap.js
```diff
@@ -0,0 +1,14 @@
...
+
+var eq = require('./internal/eq');
+
+
+test('promap', function() {
+
+  eq(typeof S.promap, 'function');
+  eq(S.promap.length, 3);
+  eq(S.promap.toString(), 'promap :: Profunctor p => (a -> b) -> (c -> d) -> p b c -> p a d');
+
```
safareli: TK?

test/of.js
```diff
@@ -0,0 +1,14 @@
...
+
+var eq = require('./internal/eq');
+
+
+test('of', function() {
+
+  eq(typeof S.of, 'function');
+  eq(S.of.length, 2);
+  eq(S.of.toString(), 'of :: Applicative f => TypeRep f -> a -> f a');
+
```
safareli: TK?

test/join.js
```diff
@@ -0,0 +1,22 @@
...
+  eq(S.join.toString(), 'join :: Chain m => m (m a) -> m a');
+
+  eq(S.join([]), []);
+  eq(S.join([[]]), []);
+  eq(S.join([[[]]]), [[]]);
+  eq(S.join([[1], [2], [3]]), [1, 2, 3]);
+  eq(S.join([[[1, 2, 3]]]), [[1, 2, 3]]);
+  eq(S.join(S.Just(S.Just(1))), S.Just(1));
+  eq(S.join(S.Just(S.Just(S.Just(1)))), S.Just(S.Just(1)));
+
```
safareli: maybe also add failing case?
```diff
+  eq(S.join(S.Nothing)), S.Nothing);
```

index.js
```diff
@@ -502,6 +506,455 @@
...
+  //. > S.equals(NaN, NaN)
+  //. true
+  //.
+  //. > S.equals(S.Just([1, 2, 3]), S.Just([1, 2, 3]))
+  //. true
+  //.
+  //. > S.equals(S.Just([1, 2, 3]), S.Just(['1', '2', '3']))
+  //. false
+  //. ```
+  S.equals = def('equals', {}, [a, b, $.Boolean], Z.equals);
```
davidchambers: That depends on whether we think `S.equals('foo', 42)` should evaluate to `false` or throw a type error. I'm in favour of the former, since the latter would make `S.equals` cumbersome to use in test suites. What do you think?

index.js
```diff
@@ -502,6 +506,455 @@
...
+  //. []
+  //.
+  //. > S.zero(Object)
+  //. {}
+  //.
+  //. > S.zero(S.Maybe)
+  //. Nothing
+  //. ```
+  S.zero =
+  def('zero', {f: [Z.Plus]}, [TypeRep($.TypeVariable('f')), f(a)], Z.zero);
```
davidchambers: `f :: Type -> Type` is not a member of sanctuary-def's internal Type type. Should it be? I don't know. For now, at least, we need to provide a value of type `Type`.

index.js
```diff
@@ -502,6 +506,455 @@
...
+  //.
+  //. > S.filter(S.odd, S.Just(9))
+  //. Just(9)
+  //.
+  //. > S.filter(S.odd, S.Just(4))
+  //. Nothing
+  //. ```
+  S.filter =
+  def('filter',
+      {f: [Z.Applicative, Z.Foldable, Z.Monoid]},
```
davidchambers: This is a limitation of sanctuary-def. I'd love to work out how to properly capture this constraint.

index.js
```diff
@@ -735,27 +1188,14 @@
...
-  //. > Z.map(S.flip(Math.pow)(2), [1, 2, 3, 4, 5])
+  //. > S.map(S.flip(Math.pow)(2), [1, 2, 3, 4, 5])
   //. [1, 4, 9, 16, 25]
   //. ```
   function flip(f, x, y) {
     return f(y, x);
   }
   S.flip = def('flip', {}, [$.Function([a, b, c]), b, a, c], flip);
 
-  //# lift :: Functor f => (a -> b) -> f a -> f b
```
davidchambers: Feel free to open an issue for discussion. :)

index.js
safareli: Returning `false` is more practical indeed. What if we add `equalsStrict` or something which has `a -> a -> bool` and throws type error? but it could be done in other PR, after we are we really want it, personally I think it would be good, as it will prevent some type errors in programs.

index.js
safareli: not sure, just looked strange

index.js
davidchambers: I see value in providing a second function. We might want to name it something bland like `equals_`, though, to avoid confusion with strict (`===`) equality.

test/alt.js
```diff
@@ -0,0 +1,27 @@
...
+  eq(S.alt.toString(), 'alt :: Alt f => f a -> f a -> f a');
+
+  eq(S.alt([], []), []);
+  eq(S.alt([], [1, 2, 3]), [1, 2, 3]);
+  eq(S.alt([1, 2, 3], []), [1, 2, 3]);
+  eq(S.alt([1, 2, 3], [4, 5, 6]), [1, 2, 3, 4, 5, 6]);
+  eq(S.alt({}, {}), {});
+  eq(S.alt({}, {a: 1, b: 2, c: 3}), {a: 1, b: 2, c: 3});
+  eq(S.alt({a: 1, b: 2, c: 3}, {}), {a: 1, b: 2, c: 3});
+  eq(S.alt({a: 1, b: 2, c: 3}, {d: 4, e: 5, f: 6}), {a: 1, b: 2, c: 3, d: 4, e: 5, f: 6});
```
davidchambers: :zap:

test/ap.js
davidchambers: :zap:

test/filterM.js
davidchambers: :zap:

test/internal/Compose.js
davidchambers: That's right. Perhaps it would be better to use our internal [`ap`][1] function instead (in both cases). What do you think?


[1]: https://github.com/sanctuary-js/sanctuary/blob/0a704a0b72f806b4a3acd1306b7e1de899e82af2/test/internal/laws/ap.js

test/join.js
davidchambers: :zap:

test/internal/Compose.js
safareli: I think that would a good approach, as when we move this laws in a separate module, we might not want it to be depend on sanctuary, because it has curried functions .

index.js
safareli: nice ⚡️

davidchambers: I've added `equals_ :: Setoid a => a -> a -> Boolean`. I'm now wondering whether `equals` and `equals_` should be switched. The `_` suffix is used elsewhere in the library as a subtle warning to users, and I believe we should discourage users from comparing values of different types. Do you agree?

safareli: I think in "your" codebase where you know types of values, you will only use equals when two values have same type, as if you know that one value has different type then the other, it's useless to use equals as it will return false for every value of those types.

But one might use equals in library like code, when someone outside of the library could pass some value of any type. But for cases like this one could just use S.unchecked.equals.

So after writing/thinking on it i think `equals:: a -> a -> Boolena` would be nice if one could also do this `S.unchecked.equals` to access unchecked version of `equals`

test/internal/Compose.js
davidchambers: #326

index.js
davidchambers: > So after writing/thinking on it i think `equals:: a -> a -> Boolean` would be nice if one could also do this `S.unchecked.equals` to access unchecked version of `equals`

Great idea! Using the `checkTypes` switch to avoid having two versions of the same function is very elegant. I've pushed a commit to replace `S.equals` with the stricter variant.

In #262 Aldwin and I have been talking about bringing back `S.unchecked`, which would allow `S.unchecked.equals` as you suggest.

test/of.js
davidchambers: :zap:

test/promap.js
davidchambers: :zap:

test/sequence.js
davidchambers: :zap:

index.js
```diff
@@ -502,6 +506,452 @@
...
+  //.
+  //. ```javascript
+  //. > S.join([[1], [2], [3]])
+  //. [1, 2, 3]
+  //.
+  //. > S.join([[[1, 2, 3]]])
+  //. [[1, 2, 3]]
+  //.
+  //. > S.join(S.concat)('abc')
+  //. 'abcabc'
```
davidchambers: Yesterday when I asked about `\x -> f x x` @brendanhay told me that `join` is the W combinator! Very cool! See [Stack Overflow][1] if you're interested in the substitutions which show that `S.join(f)` is equivalent to `x => f(x)(x)`.


[1]: http://stackoverflow.com/q/38401606/312785

Avaq: I was just looking at it. Quite something. So we have:

```
map = B
ap = S
join = W
```

I'll add them to [my table](https://gist.github.com/Avaq/1f0636ec5c8d6aed2e45#file-combinators-md) ;).

index.js
```diff
@@ -774,12 +774,25 @@
...
+  //. Note what happens when `Apply f => f` is replaced with `Function x`:
+  //.
+  //.     Apply f => f (a -> b) -> f a -> f b
+  //.     Function x (a -> b) -> Function x a -> Function x b
+  //.     Function x (a -> c) -> Function x a -> Function x c
+  //.     Function x (b -> c) -> Function x b -> Function x c
+  //.     Function a (b -> c) -> Function a b -> Function a c
+  //.     (a -> b -> c) -> (a -> b) -> (a -> c)
+  //.
+  //. This is the S combinator from combinatory logic.
```
Avaq: Maybe move the conclusion above the explanation, so a reader knows why he's reading it, and may choose beforehand to skim over it.

index.js
```diff
@@ -598,15 +598,28 @@
...
+  //. Note what happens when `Functor f => f` is replaced with `Function x`:
+  //.
+  //.     Functor f => (a -> b) -> f a -> f b
+  //.     (a -> b) -> Function x a -> Function x b
+  //.     (a -> c) -> Function x a -> Function x c
+  //.     (b -> c) -> Function x b -> Function x c
+  //.     (b -> c) -> Function a b -> Function a c
+  //.     (b -> c) -> (a -> b) -> (a -> c)
+  //.
+  //. [`compose`](#compose) is thus a specialized form of `map`.
```
Avaq: Same remark as with `ap`

index.js
```diff
@@ -887,12 +887,24 @@
...
+  //.
+  //. Note what happens when `Chain m => m` is replaced with `Function x`:
+  //.
+  //.     Chain m => m (m a) -> m a
+  //.     Function x (Function x a) -> Function x a
+  //.     Function x (Function x b) -> Function x b
+  //.     Function a (Function a b) -> Function a b
+  //.     (a -> a -> b) -> (a -> b)
+  //.
+  //. This is the W combinator from combinatory logic.
```
Avaq: Same remark as with `ap`.

index.js
davidchambers: Good idea. "This" in the sentence above refers to `(a -> b -> c) -> (a -> b) -> (a -> c)` rather than to `ap`, which is more general. We could, though, combine this sentence with the sentence preceding the substitutions:

_Replacing `Apply f => f` with `Function x` produces the S combinator from combinatory logic:_

test/chainRec.js
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  eq(S.chainRec(Array, permute, ''), ['oo!', 'oo?', 'on!', 'on?', 'no!', 'no?', 'nn!', 'nn?']);
+
+  function stepper(n) {
+    return n === 3000 ? map(S.Right)(function(env) { return n + env.inc; })
+                      : map(S.Left)(function(env) { return n + env.step; });
+  }
+
+  eq(S.chainRec(Function, stepper, 0)({step: 2, inc: 100}), 3100);
```
rjmk: If I'm correct that this test is as much to examine stack behaviour as output, perhaps that warrants a comment?

test/alt.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
```
rjmk: It strikes me that it would be nice to test these typeclass functions on as minimal a case as possible i.e. the `FreeFoo`.

Certainly out of scope for this PR, and I'm not sure where you get your functors from, but what do you think about it in principle?

index.js
```diff
@@ -511,6 +515,490 @@
...
+  //. > S.join(S.Just(S.Just(1)))
+  //. S.Just(1)
+  //. ```
+  //.
+  //. Replacing `Chain m => m` with `Function x` produces the W combinator
+  //. from combinatory logic:
+  //.
+  //.     Chain m => m (m a) -> m a
+  //.     Function x (Function x a) -> Function x a
+  //.     Function x (Function x b) -> Function x b
```
rjmk: I don't know how much the variable renaming helps here

test/chainRec.js
davidchambers: Good idea.

:zap:

test/alt.js
davidchambers: Are you suggesting defining a type which provides *only* the methods and/or TypeRep functions necessary to satisfy the requirements of the type class?

index.js
davidchambers: Are you suggesting this instead?

```
Chain m => m (m a) -> m a
Function x (Function x a) -> Function x a
(x -> x -> a) -> (x -> a)
```

I'm happy to make this change.

Are you suggesting removing the alpha-renaming from the `map` and `ap` documentation as well, or were you commenting on `join` specifically?

I find the alpha-renaming useful for `map` and `ap` as I find the signatures more recognizable when the conventional identifiers are used. Also, it's a good idea to perform alpha-renaming one variable at a time if one is reusing identifiers. Otherwise variables may appear to *move*, which is not the correct interpretation.

rjmk: Only `chain` really stuck out to me, but perhaps it's better to be consistent

davidchambers: Consistency led me to the longer version, but I like your shorter version. When there are only two type variables I'm happy to deal with `x` and `a`. :)

davidchambers:
:unicorn:

test/alt.js
rjmk: Pretty much. I was thinking we could use the free structures (free monoids, applicatives, monads ...). You do need a base functor for those, though, and I'm not sure whether the natural choice (`Identity`) actually gives you much assurance.

PR #217, opened 2016-05-16
sanctuary-def@0.6.x

davidchambers:
sanctuary-js/sanctuary-def#72

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ setup:
 
 .PHONY: test
 test:
-	$(ISTANBUL) cover node_modules/.bin/_mocha -- --recursive
+	$(ISTANBUL) cover node_modules/.bin/_mocha -- --recursive --timeout 10000
 	$(ISTANBUL) check-coverage --branches 100
 ifneq ($(shell node --version | sed 's/[.][^.]*$$//'),v0.10)
 	$(DOCTEST) -- index.js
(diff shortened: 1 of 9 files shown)


svozza:
Any particular reason to favour `_test` rather than the new `$.test` function?

Makefile
```diff
@@ -58,7 +58,7 @@ setup:
 
 .PHONY: test
 test:
-	$(ISTANBUL) cover node_modules/.bin/_mocha -- --recursive
+	$(ISTANBUL) cover node_modules/.bin/_mocha -- --recursive --timeout 10000
```
davidchambers: Two tests timed out in <https://circleci.com/gh/sanctuary-js/sanctuary/999>.

test/I.js
```diff
@@ -13,6 +13,7 @@ describe('I', function() {
 
   it('returns its argument', function() {
     eq(S.I([1, 2, 3]), [1, 2, 3]);
+    eq(S.I(['foo', 42]), ['foo', 42]);
```
davidchambers: This works now because this patch adds [`$.Pair`][1] to Sanctuary's environment. #206 will fix the problem of "invalid" values in a general manner.


[1]: https://github.com/sanctuary-js/sanctuary-def#pair

davidchambers:
> Any particular reason to favour `_test` rather than the new `$.test` function?

Nope. They work equally well in these cases. I decided to make the easy change as many occurrences of `_test` will disappear in #216. :)

svozza:
Cool. I just thought I might be missing something!

Makefile
svozza: Interesting, I've seen these failures locally and just thought it was to do with the terrible laptop I'm using.

svozza:
LGTM.

PR #222, opened 2016-05-27
array: add S.zipWith

davidchambers:
This'll check one more box in #145.

### DIFF
--- a/test/zipWith.js
+++ b/test/zipWith.js
@@ -1,0 +1,83 @@
+'use strict';
+
+var throws = require('assert').throws;
+
+var eq = require('./utils').eq;
+var errorEq = require('./utils').errorEq;
+var S = require('..');
+
+
+//  pow :: Number -> Number -> Number
+var pow = function(exp) {
+  return function(n) {
+    return Math.pow(n, exp);
+  };
+};
+
+
+describe('zipWith', function() {
+
+  it('is a ternary function', function() {
+    eq(typeof S.zipWith, 'function');
+    eq(S.zipWith.length, 3);
+  });
+
+  it('type checks its arguments', function() {
+    throws(function() { S.zipWith(['foo', 'bar', 'baz']); },
+           errorEq(TypeError,
+                   'Invalid value\n' +
+                   '\n' +
+                   'zipWith :: Function -> Array a -> Array b -> Array c\n' +
+                   '           ^^^^^^^^\n' +
+                   '              1\n' +
+                   '\n' +
+                   '1)  ["foo", "bar", "baz"] :: Array String\n' +
+                   '\n' +
+                   'The value at position 1 is not a member of ‘Function’.\n'));
+
+    throws(function() { S.zipWith(pow, 42); },
+           errorEq(TypeError,
+                   'Invalid value\n' +
+                   '\n' +
+                   'zipWith :: Function -> Array a -> Array b -> Array c\n' +
+                   '                       ^^^^^^^\n' +
+                   '                          1\n' +
+                   '\n' +
+                   '1)  42 :: Number, FiniteNumber, NonZeroFiniteNumber, Integer, ValidNumber\n' +
+                   '\n' +
+                   'The value at position 1 is not a member of ‘Array a’.\n'));
+
+    throws(function() { S.zipWith(pow, [], 42); },
+           errorEq(TypeError,
+                   'Invalid value\n' +
+                   '\n' +
+                   'zipWith :: Function -> Array a -> Array b -> Array c\n' +
+                   '                                  ^^^^^^^\n' +
+                   '                                     1\n' +
+                   '\n' +
+                   '1)  42 :: Number, FiniteNumber, NonZeroFiniteNumber, Integer, ValidNumber\n' +
+                   '\n' +
+                   'The value at position 1 is not a member of ‘Array b’.\n'));
+  });
+
+  it('zips two arrays with the specified binary function', function() {
+    eq(S.zipWith(pow, [], []), []);
+    eq(S.zipWith(pow, [1, 2, 3], []), []);
+    eq(S.zipWith(pow, [], [4, 5, 6]), []);
+    eq(S.zipWith(pow, [1, 2, 3], [7, 6, 5]), [7, 36, 125]);
+    eq(S.zipWith(pow, [1, 2, 3], [7, 6, 5, 4]), [7, 36, 125]);
+    eq(S.zipWith(pow, [1, 2, 3, 4], [7, 6, 5]), [7, 36, 125]);
+
+    var expected = [S.Just('f'), S.Just('ba'), S.Just('baz')];
+    eq(S.zipWith(S.take, [1, 2, 3], ['foo', 'bar', 'baz']), expected);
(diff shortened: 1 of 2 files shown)


test/zipWith.js
```diff
@@ -0,0 +1,82 @@
...
+                   '\n' +
+                   'The value at position 1 is not a member of ‘Array b’.\n'));
+  });
+
+  it('zips two arrays with the specified binary function', function() {
+    eq(S.zipWith(pow, [], []), []);
+    eq(S.zipWith(pow, [1, 2, 3], []), []);
+    eq(S.zipWith(pow, [], [4, 5, 6]), []);
+    eq(S.zipWith(pow, [1, 2, 3], [7, 6, 5, 4]), [7, 36, 125]);
+    eq(S.zipWith(pow, [1, 2, 3, 4], [7, 6, 5]), [7, 36, 125]);
```
svozza: Do we want a test with two arrays of the same size?

davidchambers: :zap:

joneshf:
Another option is to not provide this, and instead point people to a semantically correct<sup>1</sup> `ZipArray a` data type. Then the `zipWith` we know is just `lift2`!

Side benefit of pushing the right data structure is that `transpose` comes for free as well. It is `sequence` for a `ZipArray (ZipArray a)`.

---

<sup>1</sup> It's semantically correct in the same way that pointing someone to a "Set" data type is correct when you actually want an array to have the behavior of a set, rather than pointing them to functions that operates on arrays as though they are sets.

davidchambers:
Great idea, @joneshf! In response to your comment I read a little about Haskell's [ZipList][1]. I'll implement it in JavaScript to make sure I understand it. :)


[1]: https://en.wikibooks.org/wiki/Haskell/Applicative_functors#ZipList

rjmk:
@davidchambers Good idea to implement it for understanding.

If we want to include it, it might be worth picking up [joneshf/ziparray](https://github.com/joneshf/ziparray)

davidchambers:
Thanks for the link, @rjmk. You should do more self-promotion, @joneshf! :)

I sat in a park today and implemented the following functions:

  - `ZipArray :: Array a -> ZipArray a`
  - `ZipArray#map :: ZipArray a ~> (a -> b) -> ZipArray b`
  - `ZipArray#ap :: ZipArray (a -> b) ~> ZipArray a -> ZipArray b`

I now know that Hardy has written a library which contains these functions, but it was a useful learning exercise nonetheless.

I appreciate the elegance of encoding behaviour in data types and leveraging generic functions such as `map`, `ap`, and `lift2`. The resulting user code, though, is less elegant:

```javascript
S.lift2(dist, ZipArray(xs), ZipArray(ys))
```

The equivalent code written in terms of `zipWith`:

```javascript
S.zipWith(dist, xs, ys)
```

`zipWith` strikes me as useful even if we decide to provide `S.ZipArray`.

I'll leave this pull request open for some time as the idea of ZipArray is new to me, I'd appreciate input from others, and I'd like to see how Hardy's new library develops. :)

scott-christopher:
> @joneshf: Side benefit of pushing the right data structure is that transpose comes for free as well. It is sequence for a `ZipArray (ZipArray a)`

I was halfway through writing up that this wouldn't be possible in JS because we can't represent an infinite list to implement `ZipList.of`, meaning we couldn't make use of `sequence`/`traverse` ... and then I wondered if it were possible with some fancy new iterators, just for giggles.

It's not pretty, but behold:
```js
class ZipList {
  constructor(xs) { this.xs = xs }
  map(f) {
    return ZipList.of(f).ap(this)
  }
  ap(other) {  
    const fsIter = this.xs[Symbol.iterator](),
          xsIter = other.xs[Symbol.iterator](),
          zipFX = () => {
      const fsNext = fsIter.next(),
            xsNext = xsIter.next()
      return (fsNext.done || xsNext.done)
        ? []
        : [fsNext.value(xsNext.value), ...zipFX()]
    }
    return new ZipList(zipFX())
  }
  static of(a) {
    const repeatA = {
      [Symbol.iterator]: () => ({
        next: () => ({ value: a, done: false })
      })
    }
    return new ZipList(repeatA)
  }
  [Symbol.iterator]() { return this.xs[Symbol.iterator]() }
}
```
and sure enough:
```js
const _zipWith = (f, xs, ys) => [...ZipList.of(f).ap(new ZipList(xs)).ap(new ZipList(ys))]
_zipWith(add, [1, 2, 3], [4, 5, 6, 7]) // [5, 7, 9]

const _transpose = xss => [...sequence(ZipList.of, map(xs => new ZipList(xs), xss))]
_transpose([[1, 2, 3], [4, 5, 6], [7, 8, 9]]) // [[1, 4, 7], [2, 5, 8], [3, 6, 9]]
```
😄 

Regarding the topic at hand, I agree that `zipWith` is still useful to expose, whether implemented via `ZipList` or otherwise (the full applicative instance for `ZipList` along with the above iterators aren't necessary for implementing `zipWith`, as we can get by with just `ZipList.ap`)

joneshf:
> The resulting user code, though, is less elegant:

The recommended solution is that you should push the constructor as far to the edge as possible. The `Functor`, `Foldable` and `Traversable` implementations are the exact same between `ZipArray a` and `[a]`. So any code you have that uses `map`, `reduce`, `traverse` and friends doesn't need to change at all!

I.e. you would write

```js
S.lift2(dist, xs, ys)
```
rather than 
```js
S.lift2(dist, ZipArray(xs), ZipArray(ys))
```

because you would have defined it as

```js
const xs = ZipArray([...]);
```

> It's not pretty, but behold:

Nice!

> `zipWith` strikes me as useful even if we decide to provide `S.ZipArray`.

 
> Regarding the topic at hand, I agree that `zipWith` is still useful to expose, whether implemented via `ZipList` or otherwise

Please don't take my suggestion to mean that you shouldn't merge this PR. I don't want to stop progress. I want to offer alternative ideas.

davidchambers:
> The recommended solution is that you should push the constructor as far to the edge as possible.

Ah, yes. I get a sense now of how you imagine this being used.

> Please don't take my suggestion to mean that you shouldn't merge this PR. I don't want to stop progress. I want to offer alternative ideas.

I :heart: alternative ideas. The reason I'd like to leave this pull request open for a while is I'm *almost* ready to publish v0.11.0 and I'd like to do so before merging this. If we decide to include `zipWith`, it can certainly wait until the following release. :)

davidchambers:
How do people feel about this function? I'm leaning towards including it, because I can imagine situations in which I'd use it if it were available rather than introduce a dependency on [ziparray][1].

By the way, the Static Land support in ziparray is very cool, @joneshf! /cc @rpominov


[1]: https://www.npmjs.com/package/ziparray

rjmk:
@davidchambers My thought is that I'd rather it out (as in existing, and Sanctuary-compatible, but elsewhere) if possible, but I'm not sure.

a) Sanctuary's story for having an ecosystem around it is that great at the moment. (Hopefully wrong on this)
b) Whether others feel the same as I do about keeping core Sanctuary minimal

Even if people do feel the same, I think given the functions we currently have in, it probably belongs equally well

davidchambers:
I'm going to close this pull request for now. We may choose to revisit this function in the future.

PR #223, opened 2016-05-28
Add mergeStrMaps

svozza:
I've purposely used `for` loops here rather than the native `map` or `reduce` functions because internally they do null checks and other 'type-checking' activities so there's no point in doing that twice given we have `sanctuary-def`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2915,6 +2915,35 @@
         return Object.keys(strMap).map(function(k) { return [k, strMap[k]]; });
       });
 
+  //# mergeStrMaps :: Array (StrMap a) -> StrMap a
+  //.
+  //. Merges an array of string maps into one string map. Precedence is given
+  //. to later string maps in the array.
+  //.
+  //. ```javascript
+  //. > S.mergeStrMaps([{a: 1}, {b: 2}])
+  //. {a: 1, b: 2}
+  //.
+  //. > S.mergeStrMaps([{a: 1}, {b: 2}, {b: 3}])
+  //. {a: 1, b: 3}
+  //. ```
+  S.mergeStrMaps =
+  def('mergeStrMaps',
+      {},
+      [$.Array($.StrMap(a)), $.StrMap(a)],
+      function(xs) {
+        var strMap = {};
+        for (var idx = 0; idx < xs.length; idx += 1) {
+          var obj = xs[idx];
+          var keys = Object.keys(obj);
+          for (var idx1 = 0; idx1 < keys.length; idx1 += 1) {
+            var key = keys[idx1];
+            strMap[key] = obj[key];
+          }
+        }
+        return strMap;
+      });
+
   //. ### Number
 
   //# negate :: ValidNumber -> ValidNumber
--- a/test/mergeStrMaps.js
+++ b/test/mergeStrMaps.js
@@ -1,0 +1,55 @@
+'use strict';
+
+var throws = require('assert').throws;
+
+var eq = require('./utils').eq;
+var errorEq = require('./utils').errorEq;
+var S = require('..');
+
+
+describe('mergeStrMaps', function() {
+
+  it('is a unnary function', function() {
+    eq(typeof S.mergeStrMaps, 'function');
+    eq(S.mergeStrMaps.length, 1);
+  });
+
+  it('type checks its arguments', function() {
+    throws(function() { S.mergeStrMaps([1, 2, 3]); },
+           errorEq(TypeError,
+                   'Invalid value\n' +
+                   '\n' +
+                   'mergeStrMaps :: Array (StrMap a) -> StrMap a\n' +
+                   '                      ^^^^^^^^^^\n' +
+                   '                          1\n' +
+                   '\n' +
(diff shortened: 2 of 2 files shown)


kedashoe:
Maybe it would be helpful to wait for sanctuary-js/sanctuary-type-classes#2 for this? Then we could do something like

`concatAll :: Semigroup a => [a] -> a`

and [S.concat](https://github.com/sanctuary-js/sanctuary-type-classes/pull/2/files#diff-168726dbe96b3ce427e7fedce31bb0bcR366) everything in the loop :)

davidchambers:
This function seems overly specific. We might want to concatenate arrays of things other than objects (and the [Foldable][1] in question needn't be Array).

```javascript
> S.reduce(S.concat, {}, [{a: 1}, {b: 1}, {b: 2, c: 2}])
{a: 1, b: 2, c: 2}

> S.reduce(S.concat, '', ['a', 'b', 'c'])
'abc'

> S.reduce(S.concat, '', Cons('a', Cons('b', Cons('c', Nil()))))
'abc'
```

We don't have `S.concat` quite yet, of course, but when we do I'm not sure giving a name to `S.reduce(S.concat, {})` makes its behaviour clearer.


kedashoe:
> and the Foldable in question needn't be Array

nice! What would the signature for that be, out of curiosity? We can't use `[a]` to denote a `Foldable`, can we?

I am +1 for a generic `concatAll` fwiw, I use it fairly often in front end stuff (with arrays) and the object instance also comes up every now and then.

And I was wrong of course, we need a monoid as implied with your `reduce`. Maybe this is similar to http://hackage.haskell.org/package/base-4.9.0.0/docs/Data-Monoid.html#v:mconcat?

davidchambers:
> What would the signature for that be, out of curiosity?

This, I think:

```haskell
(Foldable f, Monoid m) => m a -> f (m a) -> m a
```

> I am +1 for a generic `concatAll`

I don't see how we can achieve this. Given `['a', 'b', 'c']` we can use `empty('a')` to determine the empty value, but given `[]` we're out of luck. So we require an initial value. At that point all we can do is provide a name for `S.reduce(S.concat)`, but I can't imagine a name which would increase clarity.

kedashoe:
> So we require an initial value. At that point all we can do is provide a name for S.reduce(S.concat), but I can't imagine a name which would increase clarity.

Good point. Since I use these often enough I wouldn't mind if they make it in but maybe it is best to not include.

svozza:
The reason I didn't want to use `concat` is because then we'll be cloning every intermediate object as we go through the list.

davidchambers:
That's a compelling argument, @svozza. Can we come up with a different name, though? `mergeAll` doesn't seem right for a "variadic" form of `concat` (and `concatAll` doesn't seem appropriate for a function which is not polymorphic). `mergeStrMaps`?

svozza:
Apologies, I'm only back from holidays and much as I love Sanctuary I was too busy drinking and laying by the beach to write any code. ;) I  was thinking though, can we not have our cake and eat it here: we're going to have to special case arrays for performance reasons anyway so the problem of specifying an inital value for a `reduce` is not issue so we could still do the more general `concatAll` function.

davidchambers:
I'm on board with this function, but `mergeAll` doesn't sit well with me. It suggests the existence of a binary `merge` function, which is misleading. `mergeStrMaps` doesn't roll off the tongue, so I'm open to alternatives, but I do prefer it to `mergeAll`.

index.js
```diff
@@ -2922,6 +2922,35 @@
...
+  //.
+  //. ```javascript
+  //. > S.mergeAll([{a: 1}, {b: 2}])
+  //. {a: 1, b: 2}
+  //.
+  //. > S.mergeAll([{a: 1}, {b: 2}, {b: 3}])
+  //. {a: 1, b: 3}
+  //. ```
+  S.mergeAll =
+    def('mergeAll',
```
davidchambers: :arrow_left: :arrow_left:

svozza:
Sorry, I misunderstood why we can't create the more general `(Foldable f, Monoid m) => m a -> f (m a) -> m a` function, I thought you were talking about arrays specifically but I see that you mean if someone gave `S.concatAll([])` then we'd have no idea what to return. 

I agree that `mergeAll` isn't suitable a suitable name, `mergeStrMaps` is a bit of a mouthful but at least it's unambiguous.

svozza:
:zap:

index.js
```diff
@@ -2915,6 +2915,35 @@
         return Object.keys(strMap).map(function(k) { return [k, strMap[k]]; });
       });
 
+  //# mergeStrMaps :: Array (StrMap a) => StrMap a
```
davidchambers: This should be a regular arrow (`->`) rather than a fat arrow (`=>`).

svozza:
I'm going to close this now that we have `Object.concat`.

davidchambers:
Thanks for closing this pull request, but thanks very much for opening it in the first place! Much good came of it, despite the fact it was not merged. :)

PR #225, opened 2016-05-30
maybe: change type of S.Nothing from ‘-> Maybe a’ to ‘Maybe a’

davidchambers:
Initially, `S.Nothing` existed both to create the `Nothing` value and to support `R.is(S.Nothing)`. `S.Nothing` is no longer a constructor function, and we now have [`S.isNothing`][1] to answer the question of whether a given value of type `Maybe a` is `Nothing`. This means `S.Nothing` can now be the `Nothing` value, improving Haskell correspondence:

|                   | Nothing           | Just              |
| ----------------- | ----------------- | ----------------- |
| Sanctuary, before | `S.Nothing()`     | `S.Just(42)`      |
| Sanctuary, after  | `S.Nothing`       | `S.Just(42)`      |
| Haskell           | `Nothing`         | `Just 42`         |

Can anyone think of a problem with this change?

---

_This evening I realized for the first time that “nothing” is comprised of “no” and “thing”. I now think Haskell should have gone with `data Maybe a = Thing a | Nothing`. ;)_

[1]: http://sanctuary.js.org/#isNothing

### DIFF
--- a/index.js
+++ b/index.js
@@ -46,8 +46,8 @@
 //.
 //. Sanctuary embraces types. JavaScript doesn't support algebraic data types,
 //. but these can be simulated by providing a group of data constructors which
-//. return values with the same set of methods. A value of the Maybe type, for
-//. example, is created via the Nothing constructor or the Just constructor.
+//. return values with the same set of methods. A value of the Either type, for
+//. example, is created via the Left constructor or the Right constructor.
 //.
 //. It's necessary to extend Haskell's notation to describe implicit arguments
 //. to the *methods* provided by Sanctuary's types. In `x.map(y)`, for example,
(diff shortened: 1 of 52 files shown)


index.js
```diff
@@ -801,25 +801,27 @@
...
+    var isJust = box.length > 0;
+    if (isJust) this.value = box[0];
+    this.isNothing = !isJust;
+    this.isJust = isJust;
   };
 
   //# Maybe.empty :: -> Maybe a
   //.
-  //. Returns a Nothing.
+  //. Returns Nothing.
```
davidchambers: I :heart: getting to remove “a” here: conceptually there's a single `Nothing` value.

svozza:
LGTM.

Issue #130, opened 2016-01-15
fromEither :: b -> Either a b -> b


davidchambers:
This was suggested by @Bpless as a way to extract the `b` from a value of type `Either a b`. It's analogous to [`fromMaybe :: a -> Maybe a -> a`][1].

Usage:

```javascript
> S.fromEither(NaN, S.Left('XXX'))
NaN

> S.fromEither(NaN, S.Right(42))
42
```

[1]: https://github.com/plaid/sanctuary#fromMaybe

svozza:
Oh nice, I'd really like this function.

PR #229, opened 2016-06-07
Add S.fromEither to pull out values from either

wennergr:
S.fromEither :: b -> Either a b -> b

Closes: #130

### DIFF
--- a/test/fromEither.js
+++ b/test/fromEither.js
@@ -1,0 +1,41 @@
+'use strict';
+
+var throws = require('assert').throws;
+
+var eq = require('./utils').eq;
+var S = require('..');
+
+describe('fromEither', function() {
+
+  it('is a binary function', function() {
+    eq(typeof S.fromEither, 'function');
+    eq(S.fromEither.length, 2);
+  });
+
+  it('type checks its arguments', function() {
+    throws(function() { S.fromEither(0, [1, 2, 3]); },
+           'Invalid value\n' +
+           '\n' +
+           'fromEither :: b -> Either a b -> b\n' +
+           '                   ^^^^^^^^^^n' +
+           '                        1\n' +
+           '\n' +
+           '1)  [1, 2, 3] :: Array Number, Array FiniteNumber, Array NonZeroFiniteNumber, Array Integer, Array ValidNumber\n' +
+           '\n' +
+           'The value at position 1 is not a member of ‘Either a b’.\n');
+  });
+
+  it('can be applied to a Right', function() {
+    eq(S.fromEither(0, S.Right(42)), 42);
+  });
+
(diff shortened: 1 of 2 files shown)


davidchambers:
Thank you, @wennergr! This was a nice surprise. :)

davidchambers:
The order in which the functions appear in the source file determines the order in which the functions are listed in the documentation, so it's nice to group related functions. In this case, let's follow the (somewhat arbitrary) precedent set by `fromMaybe`—which appears after `isJust` and before `maybe`—by inserting `fromEither` between `isRight` and `either`.

test/fromEither.js
```diff
@@ -0,0 +1,41 @@
...
+describe('fromEither', function() {
+
+  it('is a binary function', function() {
+    eq(typeof S.fromEither, 'function');
+    eq(S.fromEither.length, 2);
+  });
+
+  it('type checks its arguments', function() {
+    throws(function() { S.fromEither(0, [1, 2, 3]); },
+                   'Invalid value\n' +
```
davidchambers: Let's adjust the indentation like so:

```javascript
throws(function() { S.fromEither(0, [1, 2, 3]); },
       'Invalid value\n' +
```

test/fromEither.js
```diff
@@ -0,0 +1,41 @@
...
+                   'The value at position 1 is not a member of ‘Either a b’.\n');
+  });
+
+  it('can be applied to a Left', function() {
+    eq(S.fromEither(0, S.Right(42)), 42);
+  });
+
+  it('can be applied to a Right', function() {
+    eq(S.fromEither(0, S.Left(42)), 0);
+  });
```
davidchambers: The expressions and the accompanying descriptions should be switched. ;)

wennergr:
Thanks for your feedback

I did not do property based tests because the fromMaybe tests didn't. Is it only used to test laws? Performance reasons? More then happy to change ;)

Also more then happy to rebase before the merge.

davidchambers:
> I did not do property based tests because the fromMaybe tests didn't. Is it only used to test laws?

The truthful answer is that property-based testing is still relatively new to me, so I don't think to use it. While it would be great to add more property-based tests, that's a larger change orthogonal to this one.

> Also more then happy to rebase before the merge.

That would be marvellous. :)

I'd like to add that I :heart: the way in which you went about this pull request. You matched the existing code and documentation style, and referenced the issue this pull request will close. Very nice!

wennergr:
Done, and thanks!

davidchambers:
:deciduous_tree:

davidchambers:
```

$ node
undefined
> S.fromEither
```

PR #231, opened 2016-06-11
Added Either#swap to flip left/right values

wennergr:
Either#swap :: Either a b ~> Either b a

Flips the left and the right values. (2-complement)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1774,6 +1774,25 @@
          [$Either(a, b), $.Boolean],
          prop('isRight'));
 
+  //# Either#swap :: Either a b ~> Either b a
+  //.
+  //. Flips the left and the right values. (2-complement)
+  //.
+  //. ```javascript
+  //. > S.Left(42).swap()
+  //. Right(42)
+  //.
+  //. > S.Right(42).swap()
+  //. Left(42)
+  //. ```
+  Either.prototype.swap =
+  method('Either#swap',
+         {},
+         [$Either(a, b), $Either(b, a)],
+         function(either) {
+           return either.isLeft ? Right(either.value) : Left(either.value);
+         });
+
   //# Either#toString :: Either a b ~> String
   //.
   //. Returns the string representation of the Either.
--- a/test/Either/Either.js
+++ b/test/Either/Either.js
@@ -80,6 +80,14 @@ var Compose = function(F, G) {
 
 describe('Either', function() {
 
+  describe('Properties of the "swap" method', function() {
+    it('it should be it\'s own inverse', function() {
+      jsc.assert(jsc.forall(EitherArb(jsc.integer, jsc.integer), function(either) {
+        return either.swap().swap().equals(either);
+      }));
+    });
+  });
+
   it('throws if called', function() {
     throws(function() { S.Either(); },
            errorEq(Error, 'Cannot instantiate Either'));
--- a/test/Either/Left.js
+++ b/test/Either/Left.js
@@ -195,6 +195,10 @@ describe('Left', function() {
     eq(S.Left('abc').toBoolean(), false);
   });
 
+  it('provides a "swap" method', function() {
+    eq(S.Left(42).swap(), S.Right(42));
+  });
+
   it('provides a "toString" method', function() {
     eq(S.Left('abc').toString.length, 0);
     eq(S.Left('abc').toString(), 'Left("abc")');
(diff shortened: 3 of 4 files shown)


davidchambers:
Thanks for the pull request. :)

I'd like to see this as a regular function rather than a method. I consider `this` to be one of the (many) bad parts of JavaScript, for several reasons:

  - “Explicit is better than implicit” (to quote _The Zen of Python_).

  - I don't want to think about context when writing JavaScript – it's incidental complexity. I like to refactor code through reasoning alone, but I cannot know whether replacing `x => a.b(x)` with `a.b` is valid without reading source code to see whether `b` references `this`.

  - I like to avoid special cases if possible. Treating one argument differently from the others strikes me as a special case.

  - Treating one argument differently restricts our freedom to choose argument order, leading to inconsistencies. Despite the symmetry which exists between

    ```haskell
    map :: Functor f => (a -> b) -> f a -> f b
    ```

    and

    ```haskell
    ap :: Apply f => f (a -> b) -> f a -> f b
    ```

    the symmetry is broken in the (object-oriented) Fantasy Land interface.

  - For some reason I find methods particularly difficult to reason about in conjunction with currying. With a curried function it's crucial to be aware of how many arguments are still to be provided, but it's never obvious to me how to adjust my thinking in response to implicit arguments.

I'm aware that one could present a list of problems with functions which do not apply to methods. One problem which comes to mind is the need to explicitly import each function one wishes to use in a module, whereas methods are imported implicitly. This is (I believe) the main reason @DrBoolean prefers method-based interfaces.

It's becoming apparent to me that we need a Sanctuary equivalent of [_The Philosophy of Ramda_][1]. I'm sorry to have drafted a section of this document in a comment on a pull request. ;)

How would you feel about `S.swap :: Either a b -> Either b a`, @wennergr? Might there be other things we decide should be "swappable" at some point? If so, perhaps `swapEither` is a better name.


[1]: http://fr.umio.us/the-philosophy-of-ramda/

wennergr:
Sorry for my late response and thanks for you long explanation.

`S.swapEither` is better then `S.swap`. (There are many things that can be `swapped`)

A design philosophy document would be great ;) 

I agree with you that instance methods are less great. Two benefits that I can see (however) with them are:

 1. Natural namespace
 2. Natural left to right flow when chaining

Less worried about (2) since there is an easy fix (such as reverse composition)

Avaq:
Wouldn't `Bifunctor | Profunctor` be equal to `Swappable`? They are working on a Bi/Pro-functor spec (https://github.com/fantasyland/fantasy-land/pull/137). It would be cool if `swap` can somehow be derived from `bimap`, but I don't think that's possible without some pretty awful practices.

wennergr:
Yeah, I like the idea of going the BiFunctor route. Maybe even call it `flip` instead of `swap`. 
We could then support to `flip` (potential) future type classes such as `BiTraversable`, `BiApplicative`, and `BiApply`

What are the awful practices that you are talking about?

Would we need a separate flip function for `ProFunctor` (even thou they would be very similar)? ending up with something like: `flipBi`, `flipPro`?

Not as familiar with the ProFunctor but it looks like a bi functor that is contravarient in it's first argument. Never used it thou.

joneshf:
How would you write `swap`?

DrBoolean:
Can't one just do it with fold?

'''js
e.fold(Right, Left)
'''

joneshf:
[`S.either`](http://sanctuary.js.org/#either), but yes.

DrBoolean:
I just watched https://www.reddit.com/r/haskell/comments/4poiq5/phil_freeman_fun_with_profunctors/

At 32:00, he swaps a Tuple with dimap

Fun stuff

davidchambers:
Thanks again for the pull request, @wennergr. The ensuing discussion was very much worthwhile.

I'll close this pull request since, as the [Magic Read Along][1] hosts pointed out, defining `swapEither` in terms of [`S.either`][2] is straightforward:

```javascript
//    swapEither :: Either a b -> Either b a
const swapEither = S.either(S.Right, S.Left);

swapEither(Left(42));
// => Right(42)

swapEither(Right(42));
// => Left(42)
```


[1]: http://www.magicreadalong.com/
[2]: https://sanctuary.js.org/#either

rjmk:
Nice 🔌 , David! :laugh:

PR #232, opened 2016-06-13
sanctuary-def@0.8.0

davidchambers:
This pull request:

  - makes the Either and Maybe types compatible with Fantasy Land v2.2.x;
  - removes type checking from Either and Maybe methods;
  - removes the Ramda dependency;
  - defines `S.__`;
  - adds [sanctuary-type-classes][1] as a dependency; and
  - updates the sanctuary-def dependency to v0.8.0 (as described below).

---

Original description:

> This pull request is not yet ready to merge since `sanctuary-def@0.7.0` is not yet available, but I couldn't wait any longer to share this. We now have the ability to describe higher-order functions via sanctuary-def!
>
> ``` javascript
> S.find(x => x, ['foo', 'bar', 'baz']);
> // ! TypeError: Invalid value
> //
> //   find :: (a -> Boolean) -> Array a -> Maybe a
> //                 ^^^^^^^
> //                    1
> //
> //   1)  "foo" :: String
> //
> //   The value at position 1 is not a member of ‘Boolean’.
> ```
>
> [HOLY SMOKES, BATMAN!](https://wall.alphacoders.com/big.php?i=274701)
>
> Consider the updated type of `find`:
>
> ``` diff
> -[$.Function, $.Array(a), $Maybe(a)]
> +[Fn(a, $.Boolean), $.Array(a), $Maybe(a)]
> ```
>
> What we currently know as `$.Function` is to be renamed `$.AnyFunction`. `$.Function` will be a type constructor of type `Array Type -> Type` – `$.Function([$.String, $.Integer])`, for example, will represent `String -> Integer`. Since curried functions are so common in the Sanctuary codebase I defined `Fn :: (Type, Type) -> Type` so we can write `Fn(a, Fn(b, c))` rather than the noisy `$.Function([a, $.Function([b, c])])`.
>
> @Avaq, you'll be able to update #191 once this pull request has been merged. :)
>
> The one other sanctuary-def feature I would like to add before publishing the next version is support for parameterized type variables. That will mean we can accurately describe the type of a function such as `lift :: Functor f => (a -> b) -> f a -> f b`!

[1]: https://github.com/sanctuary-js/sanctuary-type-classes


davidchambers:
I added `$.UnaryTypeVariable` to sanctuary-def, so the type of `lift` is now `[Fn(a, b), f(a), f(b)]` with the constraint `{f: [Functor]}`. :tada:

test/on.js
```diff
@@ -11,6 +11,7 @@ describe('on', function() {
   it('is a quaternary function', function() {
     eq(typeof S.on, 'function');
     eq(S.on.length, 4);
+    eq(S.on.toString(), 'on :: (b -> b -> c) -> (a -> b) -> a -> a -> c');
```
davidchambers: It's exciting to be able to express the types of functions such as this one, @Bradcomp. :)

Shard:
I rebased this branch on master and fixed up the conflicts at [mb-sanctuary-def](/sanctuary-js/sanctuary/tree/mb-sanctuary-def). They were all "mine-side" merges, no changes needed, tests all passing ok.

davidchambers:
More than six months later, this pull request is ready for review!

davidchambers:
This pull request is the next one we must merge en route to v0.12.0. I would love a review from @safareli or anyone else. :)

karma.conf.js
```diff
@@ -8,7 +11,10 @@ function depMain(name) {
 }
 
 //  dependencies :: Array String
-var dependencies = Object.keys(require('./package.json').dependencies);
+var dependencies = ['sanctuary-type-classes', 'sanctuary-def'];
+
+eq(dependencies.slice().sort(),
+   Object.keys(require('./package.json').dependencies).sort());
```
davidchambers: @Shard, this is my low-tech fix for #309. How does it look to you?

test/Either/Left.js
```diff
@@ -18,186 +19,60 @@ suite('Left', function() {
...
-           'The value at position 1 is not a member of ‘Function’.\n');
+  test('"fantasy-land/map" method', function() {
+    eq(S.Left('abc')[FL.map].length, 1);
+    eq(S.Left('abc')[FL.map](Math.sqrt), S.Left('abc'));
   });
 
-  test('"sequence" method', function() {
-    eq(S.Left('abc').sequence.length, 1);
-    eq(S.Left('abc').sequence(S.Maybe.of), S.Just(S.Left('abc')));
+  test('"fantasy-land/reduce" method', function() {
```
svozza: What happened to `sequence`?

test/Maybe/Nothing.js
```diff
@@ -1,174 +1,63 @@
...
-           'Invalid value\n' +
-           '\n' +
-           'Maybe#filter :: Maybe a -> Function -> Maybe a\n' +
-           '                           ^^^^^^^^\n' +
-           '                              1\n' +
-           '\n' +
-           '1)  null :: Null\n' +
-           '\n' +
-           'The value at position 1 is not a member of ‘Function’.\n');
+  test('"fantasy-land/extend" method', function() {
```
svozza: What about `filter`?

index.js
```diff
@@ -14,9 +14,9 @@
 //. [![CircleCI](https://img.shields.io/circleci/project/github/sanctuary-js/sanctuary/master.svg)](https://circleci.com/gh/sanctuary-js/sanctuary/tree/master)
 //. [![Gitter](https://img.shields.io/gitter/room/badges/shields.svg)](https://gitter.im/sanctuary-js/sanctuary)
 //.
-//. Sanctuary is a functional programming library inspired by Haskell and
-//. PureScript. It depends on and works nicely with [Ramda][]. Sanctuary
-//. makes it possible to write safe code without null checks.
+//. Sanctuary is a functional programming library inspired by Haskell
+//. and PureScript. It's stricter and more opinionated than [Ramda][].
+//. Sanctuary makes it possible to write safe code without null checks.
```
davidchambers: @Jacse, let's collaborate. I like your changes in #317, but since this pull request removes our Ramda dependency we must update the introduction accordingly. Feel free to suggest changes and/or additions to the paragraph above. :)

test/indexOf.js
```diff
@@ -3,25 +3,13 @@
...
-         'Invalid value\n' +
-         '\n' +
-         'indexOf :: a -> List a -> Maybe Integer\n' +
-         '                ^^^^^^\n' +
-         '                  1\n' +
-         '\n' +
-         '1)  null :: Null\n' +
-         '\n' +
-         'The value at position 1 is not a member of ‘List a’.\n');
+  eq(S.indexOf.toString(), 'indexOf :: a -> List a -> Maybe Integer');
```
svozza: I'm a bit confused as to what the criteria are for using Array over List in various type signatures. `indexOf` seems to me to be more an array function than a list one given how expensive it is for the latter data structure.

test/Either/Left.js
davidchambers: See #257 for the discussion which led to its removal.

test/Maybe/Nothing.js
davidchambers: This was also discussed in #257.

test/indexOf.js
Avaq: I believe the List type in Sanctuary is just a union of Array, String, Arguments and other types with a length. The type signature has `Array` when the function would not work on all of these other types as well; https://sanctuary.js.org/#list

davidchambers: The List type attempts to unify [`$.Array`][1] and [`$.String`][2]. It doesn't have anything to do with linked lists. It's a confusing type, certainly!

> ### List
>
> The List type constructor enables type signatures to describe ad hoc polymorphic functions which operate on either [`Array`][1] or [`String`][2] values.
>
> Mental gymnastics are required to treat arrays and strings similarly. `[1, 2, 3]` is a list containing `1`, `2`, and `3`. `'abc'` is a list containing `'a'`, `'b'`, and `'c'`. But what is the type of `'a'`? `String`, since JavaScript has no Char type! Thus:
>
>     'abc' :: String, List String, List (List String), ...
>
> Every member of `String` is also a member of `List String`! This affects the interpretation of type signatures. Consider the type of `indexOf`:
>
>     a -> List a -> Maybe Integer
>
> Assume the second argument is `'hello' :: List String`. `a` must then be replaced with `String`:
>
>     String -> List String -> Maybe Integer
>
> Since `List String` and `String` are interchangeable, the former can be replaced with the latter:
>
>     String -> String -> Maybe Integer
>
> It's then apparent that the first argument needn't be a single-character string; the correspondence between arrays and strings does not hold.

As for `S.indexOf` and `S.lastIndexOf`, I'd really like to remove these functions. Addressing array elements by index is not a practice we should encourage.


[1]: https://github.com/sanctuary-js/sanctuary-def#Array
[2]: https://github.com/sanctuary-js/sanctuary-def#String

index.js
Jacse: @davidchambers looks good, but maybe we should mention that Sanctuary can/should be used as an alternative to Ramda. On the top of my head: "It includes much the same functionality as Ramda, but is stricter and more opinionated."

I think this would make it more clear what functionality people can expect from Sanctuary ("write safe code without null checks" can be a bit vague). Most likely developers diving into functional programming in javascript will take a look at both Ramda and Sanctuary and I think a wording along these lines would make the relationship between them clear.

Just my 2 cents.

test/indexOf.js
Avaq: > Addressing array elements by index is not a practice we should encourage

:medal_sports:

index.js
davidchambers: I like it. I'll make the change you have proposed.

When Sanctuary first came into existence it was focused on avoiding the null checks which several Ramda functions require. Sanctuary's scope is now much broader.

test/indexOf.js
svozza: Ah yes, I vaguely remember the discussions we had about `List` a few months ago now.

test/Either/Left.js
svozza: Ah, I missed that one.

karma.conf.js
Shard: Yea this is probably the better way now that I think about it, just having a test check if things are out of sync.

👍

index.js
davidchambers: I've updated the introduction. I went with "provides a similar suite of functions" rather than "includes much the same functionality".

Jacse: @davidchambers yeah that's better.

index.js
```diff
@@ -199,36 +200,25 @@
...
-      return x != null && typeof x[name] === 'function';
-    };
-  }
+  //  Fn :: (Type, Type) -> Type
+  function Fn(x, y) { return $.Function([x, y]); }
 
-  //  negativeZero :: a -> Boolean
-  var negativeZero = R.either(R.equals(-0), R.equals(new Number(-0)));
+  //  Thunk :: Type -> Type
+  function Thunk(x) { return $.Function([x]); }
```
davidchambers: @safareli, we can replace `$.Function([x])` with `$.Thunk(x)` once sanctuary-def supports this. :)

Issue #241, opened 2016-08-06
encase return Left/Nothing when null / undefined / NaN


JAForbes:
I'm finding that despite using encase, I'm still encountering nulls (they are pervasive in JS libraries).  Would it to possible to make encase even stricter, and return Left/Nothing if the result is null, undefined or NaN?

joneshf:
This suggest sounds sane from a consistency standpoint. We have `toMaybe` which does something like this null stuff. And `toMaybe` looks very similar to what an `encase0` would look like:

```js
encase3 :: (a -> b -> c -> d) -> (a -> b -> c -> Maybe d)
encase2 :: (a -> b -> c)      -> (a -> b -> Maybe c)
encase  :: (a -> b)           -> (a -> Maybe b)
toMaybe :: (a?)               -> (Maybe a)
```

I think this sounds like a good idea, but I think others probably have a better idea of whether or not it actually is.

davidchambers:
I'm of the opinion that exceptions and "bogus" values should be handled separately for maximum flexibility and ease of understanding. Consider the following functions:

```haskell
-- ! indicates a function which may raise an exception
f :: String -> Integer !
g :: String -> Nullable Integer
h :: String -> Nullable Integer !
```

We can create safe versions of these functions using [`encase`][1], [`toMaybe`][2], or both:

```javascript
//    fʹ :: String -> Maybe Integer
const fʹ = S.encase(f);

//    gʹ :: String -> Maybe Integer
const gʹ = S.compose(S.toMaybe, g);

//    hʹ :: String -> Maybe Integer
const hʹ = S.compose(R.chain(S.toMaybe), S.encase(h));
```

From the definition of `hʹ` we can infer that `h` may raise an exception or return `null`/`undefined`. Were we to provide a function which handled both failure conditions, the definition of `hʹ` could become `S.<name>(h)`, but the reader would then wonder whether `h` throws, returns `null`, or both.

Additionally, some functions can legitimately return `null`. The result of `JSON.parse('null')`, for example, is `null`, so `S.encase(JSON.parse, 'null')` should evaluate to `Just(null)`.

While `encase` and `toMaybe` have a common goal, they are solving different problems. I suggest we keep them separate.

@JAForbes and @joneshf, do you find these points compelling?


[1]: https://sanctuary.js.org/#encase
[2]: https://sanctuary.js.org/#toMaybe

JAForbes:
@davidchambers your point is compelling! :)

[...]

PR #249, opened 2016-09-02
Adds `toEither :: a -> b? -> Either a b`

### LINKED ISSUE #83: toEither :: a -> b? -> Either a b
Suggested by @raine on [Gitter][1].

```javascript
> S.toEither('Invalid protocol', 'ftp://example.com/'.match(/^(http:|https:)[/][/]/))
Left("Invalid protocol")

> S.toEither('Invalid protocol', 'http://example.com/'.match(/^(http:|https:)[/][/]/))
Right(["http://", "http:"])
```

This would mirror [`S.toMaybe`][2].

[1]: https://gitter.im/plaid/sanctuary
[2]: https://github.com/plaid/sanctuary#toMaybe

rjmk:
A function for converting a possibly null-y value to an Either
Closes #83

### DIFF
--- a/index.js
+++ b/index.js
@@ -1265,7 +1265,7 @@
 
   //# toMaybe :: a? -> Maybe a
   //.
-  //. Takes a value and returns Nothing if the value is null or undefined;
+  //. Takes a value and returns Nothing if the value is `null` or `undefined`;
   //. Just the value otherwise.
   //.
   //. ```javascript
@@ -1882,6 +1882,31 @@
       {},
       [b, $Either(a, b), b],
       function(x, either) { return either.isRight ? either.value : x; });
+
+  //# toEither :: a -> b? -> Either a b
+  //.
+  //. Converts an arbitrary value to an Either: a Left if the value is `null`
+  //. or `undefined`; a Right otherwise. The first argument specifies the
+  //. value of the Left in the "failure" case.
+  //.
+  //. ```javascript
+  //. > S.toEither('XYZ', null)
+  //. Left('XYZ')
+  //.
+  //. > S.toEither('XYZ', 'ABC')
+  //. Right('ABC')
+  //.
+  //. > R.map(R.head, S.toEither('Invalid protocol', 'ftp://example.com/'.match(/^https?:/)))
+  //. Left('Invalid protocol')
+  //.
+  //. > R.map(R.head, S.toEither('Invalid protocol', 'https://example.com/'.match(/^https?:/)))
+  //. Right('https:')
+  //. ```
+  S.toEither =
+  def('toEither',
+      {},
+      [a, b, $Either(a, b)],
+      function(x, y) { return y == null ? Left(x) : Right(y); });
 
   //# either :: (a -> c) -> (b -> c) -> Either a b -> c
   //.
--- a/test/toEither.js
+++ b/test/toEither.js
@@ -1,0 +1,23 @@
+'use strict';
+
+var eq = require('./utils').eq;
+var S = require('..');
+
+
+describe('toEither', function() {
+
+  it('is a binary function', function() {
+    eq(typeof S.toEither, 'function');
+    eq(S.toEither.length, 2);
+  });
+
+  it('returns Left of the first argument when second argument is `null`-y', function() {
+    eq(S.toEither('a', null), S.Left('a'));
+    eq(S.toEither('a', undefined), S.Left('a'));
+  });
+
+  it('returns a Right of the second argument when it is not `null`-y', function() {
+    eq(S.toEither('a', 42), S.Right(42));
+  });
+
+});


davidchambers:
Very nice! I'll make several comments about minor issues, but they should all be easy to address. :)

test/toEither.js
```diff
@@ -0,0 +1,22 @@
...
+describe('toEither', function() {
+
+  it('is a binary function', function() {
+    eq(typeof S.toEither, 'function');
+    eq(S.toEither.length, 2);
+  });
+
+  it('returns Left of the first argument when second argument is `null`', function() {
+    eq(S.toEither('a', null), S.Left('a'));
+  });
```
davidchambers: Let's include an assertion for `S.toEither('a', undefined)`.

rjmk:
@davidchambers Do you prefer that I add commits separately during the review process and squash at the end, or just push up the amended commit as I fix?

index.js
```diff
@@ -1943,6 +1943,29 @@
         return either.isRight ? [either.value] : [];
       }));
 
+  //# toEither :: a -> b? -> Either a b
+  //.
+  //. Takes a failure value, 'x', and a possibly null or undefined value, 'y'.
+  //. Returns a Right of 'y' if it is defined and not null, otherwise returns
+  //. a Left of 'x'.
```
davidchambers: I find mention of "'x'" and "'y'" a bit confusing. I think we could get away with referring to "the first argument" and "the second argument".

How about this?

_Converts an arbitrary value to an Either: a Left if the value is `null` or `undefined`; a Right otherwise. The first argument specifies the value of the Left in the "failure" case._

I'd like to improve the descriptions of many of the library's functions. In many cases the description is just an English translation of the type signature. In this case, for example, a direct translation would be:

_Returns a Left of its first argument if its second argument is `null` or `undefined`; a Right of its second argument otherwise._

I like the mention of a "failure value" in your description. I'd like to avoid mention of "'x'" and "'y'" though.

index.js
```diff
@@ -1943,6 +1943,29 @@
...
+  //.
+  //. See also ['encaseEither'](#encaseEither)
```
davidchambers: I'm not convinced this link is warranted. Handling `null`/`undefined` and handling exceptions strike me as quite different concerns.

index.js
rjmk: Agreed! I'd modelled this after `encaseEither`, but I case naming arguments is more important when some are functions

index.js
rjmk: Fair enough. I was thinking of #241 here. But perhaps the link (if there should be one) should go the other way?

index.js
```diff
@@ -1943,6 +1943,29 @@
...
+  //. > S.toEither("Invalid protocol", "ftp://example.com/".match(/^https?:\/\//))
+  //. Left("Invalid protocol")
+  //. ```
+  S.toEither =
+  def('toEither',
+      {},
+      [a, b, $Either(a, b)],
+      function(x, y) {
+        return y == null ? Left(x) : Right(y);
+      });
```
davidchambers: Let's define the anonymous function on one line: `function(x, y) { return y == null ? Left(x) : Right(y); }`.

index.js
```diff
@@ -1943,6 +1943,29 @@
...
+  //.
+  //. See also ['encaseEither'](#encaseEither)
+  //.
+  //. ```javascript
+  //. > S.toEither("No string", "ftp://example.com/")
+  //. Right("ftp://example.com/")
+  //.
+  //. > S.toEither("Invalid protocol", "ftp://example.com/".match(/^https?:\/\//))
+  //. Left("Invalid protocol")
+  //. ```
```
davidchambers: I like the symmetry of the examples in #83. It shows that `s.match(/^(http:|https:)[/][/]/)` can result in `null` or some other value depending on the value of `s`.

I think the examples should either be realistic, as in #83, or very simple (e.g. `S.toEither('XXX', null)` and `S.toEither('XXX', 42)`).

davidchambers:
> Do you prefer that I add commits separately during the review process and squash at the end, or just push up the amended commit as I fix?

I'm happy either way. Use whichever process works better for you.

index.js
davidchambers: If we're to include a link I'd like to include links in both directions. I believe all the current links are reciprocal.

index.js
rjmk: So, annoyingly, there's a difference between the results of `match` in node and the browser. In node, the result is an array with some additional properties tacked on. It looks like

```js
[ "https://", index: 0, input "https://example.com" ] // Yes, no `{`!
```

We could use that, though it's a little confusing. I considered dropping `match`, but it's a shame because it's the only standard library function I'm aware of that returns `null` (apart from `querySelector` and its ilk).

I think maybe the very simple example is best?

index.js
rjmk: Do you prefer no link?

index.js
rjmk: What's very annoying is though it stringifies this way, it's not easy to create. I could `arrayify` the result with the old `[].slice.call`, but I think that's too complex for an example. Thoughts?

davidchambers: Thanks for the explanation. I understand the problem now. I agree that it would be nice to use `String#match` (or `RegExp#exec`). Rather than use `[].slice.call` we could use `R.head`:

```javascript
> R.map(R.head, S.toEither('Invalid protocol', 'ftp://example.com/'.match(/^https?:/)))
Left('Invalid protocol')

> R.map(R.head, S.toEither('Invalid protocol', 'http://example.com/'.match(/^https?:/)))
Right('http:')
```

It's a bit longer than I would like, but it's quite a nice demonstration of the function. Perhaps we should precede these examples with two trivial examples.

index.js
davidchambers: It's your call. If you think people might find it useful, let's keep it (and add a link in the other direction).

index.js
rjmk: I hadn't though of showing off the `map`. I like the above (definitely with trivial examples first)

rjmk:
@davidchambers I think that's everything addressed

davidchambers:
Thanks for the quick updates. I have three more requests:

  - Please use single-quoted strings in the examples (sorry, the linter doesn't catch this).

  - Let's drop the `//` from the protocol. `location.protocol` does not include the slashes, and the regexp literal will look less noisy without the `\/\/`.

  - `toMaybe` is defined between `fromMaybe` and `maybe`. Let's move the definition of `toEither` to between `fromEither` and `either`.

rjmk:
⚡

index.js
```diff
@@ -1265,7 +1265,7 @@
 
   //# toMaybe :: a? -> Maybe a
   //.
-  //. Takes a value and returns Nothing if the value is null or undefined;
+  //. Takes a value and returns Nothing if the value is `null` or `undefined`;
```
davidchambers: :thumbsup:

davidchambers:
Very nice patch, @rjmk. Thank you for tolerating my nitpicking. :)

:deciduous_tree:

davidchambers:
```

$ node
undefined
> S.toEither
toEither :: a -> b -> Either a b
```

Issue #248, opened 2016-09-02
Document required node version in `CONTRIBUTING.md`


rjmk:
The doctest for

```console
> S.encaseEither(S.I, JSON.parse, '[')
Left(new SyntaxError('Unexpected end of input'))
```

will break as `JSON.parse`'s error message changed at some point between node 4 and node 6 to "Unexpected end of JSON input"

We should document which node versions can be used by people developing the library

davidchambers:
How do you feel about this now that #250 has been merged, @rjmk?

rjmk:
Much better solution!

PR #250, opened 2016-09-02
switch from Node v5.x.x to Node v6.x.x

davidchambers:
This addresses the test failure noted in #248.

As can be seen in the diff, we currently have a guard which prevents doctests from being run in Node v0.10.x (which doesn't support arrow functions). It's completely reasonable to run doctests in just one environment, since their purpose is to keep the documentation up to date rather than guarantee compatibility across Node versions. I suggest the latest stable version if we're to choose just one.

I hope these changes obviate the need to mention a version in __CONTRIBUTING.md__, as it would be easy for such a note to become out of date. Let's merge this pull request then discuss in #248 whether to add a note.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,8 @@ setup:
 test:
 	$(ISTANBUL) cover node_modules/.bin/_mocha -- --recursive --timeout 10000
 	$(ISTANBUL) check-coverage --branches 100
-ifneq ($(shell node --version | sed 's/[.][^.]*$$//'),v0.10)
+ifeq ($(shell node --version | cut -d . -f 1),v6)
 	$(DOCTEST) -- index.js
+else
+	@echo '[WARN] Doctests are only run in Node v6.x.x (current version is $(shell node --version))' >&2
 endif
--- a/circle.yml
+++ b/circle.yml
@@ -1,7 +1,8 @@
 dependencies:
   override:
+    - printf '%s\n' color=false progress=false >.npmrc
     - rm -rf node_modules
-    - case $CIRCLE_NODE_INDEX in 0) make setup ;; 1) nvm exec 0.12.7 make setup ;; 2) nvm exec 4 make setup ;; 3) nvm exec 5 make setup ;; esac:
+    - case $CIRCLE_NODE_INDEX in 0) make setup ;; 1) nvm exec 0.12.7 make setup ;; 2) nvm exec 4 make setup ;; 3) nvm install 6 && nvm exec 6 make setup ;; esac:
         parallel: true
 
 machine:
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -60,6 +60,6 @@ setup:
 test:
 	$(ISTANBUL) cover node_modules/.bin/_mocha -- --recursive --timeout 10000
 	$(ISTANBUL) check-coverage --branches 100
-ifneq ($(shell node --version | sed 's/[.][^.]*$$//'),v0.10)
+ifeq ($(shell node --version | cut -d . -f 1),v6)
```
rjmk: Perhaps this belongs in the issue, but (assuming makefiles have a concept of else), it might be appropriate to place a message here? That way, a contributor will realise if doctests didn't run and they may need to fix something before opening a PR

rjmk:
👍

Makefile
davidchambers: Very good idea! I'll make a change so that a warning message is emitted in the `else` branch. :)

davidchambers: Updated. See below. Sample output:

>     ⚠️  Doctests are only run in Node v6.x.x (current version is v4.5.0)

rjmk: Great! I often run in a terminal without escape code support, but the string of escape codes would at least draw my attention :)

davidchambers: Good point. It doesn't print properly in the CircleCI logs either. I'll use colour codes instead:

```console
$ echo $'\e[33m\e[7mWARN\e[27m Doctests are only run in Node v6.x.x (current version is '"$(node --version)"$')\e[0m'
```

rjmk: My primary environment (currently) doesn't actually support *any* escape codes, but I still think they're a good idea here. (If you were very worried about this, you could have a `$TERM=dumb` check)

davidchambers: After some trial and error, I think I have a version which will do the right thing in most environments:

1.  If `$(TERM)` is `dumb`, <code>$(<a href="https://www.gnu.org/software/make/manual/html_node/Text-Functions.html">patsubst</a> dumb,,$(TERM))</code> will evaluate to empty string; otherwise it will evaluate to the value of `$(TERM)` (which may be empty string).

2.  <code>$(<a href="https://www.gnu.org/software/make/manual/html_node/Conditional-Functions.html#Conditional-Functions">if</a> $(patsubst dumb,,$(TERM)),\033[33m\033[7mWARN\033[27m,[WARN])</code> will evaluate to `[WARN]` if the value of `$(TERM)` is `dumb` (or empty string); `\033[33m\033[7mWARN\033[27m` otherwise. `33m` sets the foreground colour to yellow, `7m` switches foreground and background colours, and `27m` undoes the switch.

3.  This appears inside a `$''` string, but since `$` has special meaning in a makefile we must escape it with an extra dollar sign.

@rjmk, would you mind sanity checking the line below? There's no doubt we're overengineering this, but I believe the only cost is the time I've spent fiddling (which I've quite enjoyed).

One can test this by commenting out lines 61 and 62 then running `make test` and `TERM=dumb make test`. CircleCI is happy now too.

rjmk: Makes sense to me!

davidchambers: Actually, I decided to restrict myself to ASCII regardless of environment. I had too much trouble getting CircleCI to do the right thing. Also, I'm now redirecting the output to stderr.

PR #251, opened 2016-09-03
Adds property tests for `S.I`

rjmk:
I imagine that there's a lot of stuff here that people feel
strongly about. Is a `properties` file OK? The naming of the
arbitraries? Using "is idempotent" rather than "satisfies
idempotency"? I thought it would be useful to have a concrete
example to discuss

### DIFF
--- a/test/I.js
+++ b/test/I.js
@@ -3,6 +3,7 @@
 var S = require('..');
 
 var eq = require('./internal/eq');
+var properties = require('./properties');
 
 
 describe('I', function() {
@@ -16,6 +17,14 @@ describe('I', function() {
     eq(S.I([1, 2, 3]), [1, 2, 3]);
     eq(S.I(['foo', 42]), ['foo', 42]);
     eq(S.I({'@@type': 'my-package/Foreign'}), {'@@type': 'my-package/Foreign'});
+  });
+
+  it('is idempotent', function() {
+    eq(properties.idempotent(S.I), true);
+  });
+
+  it('is an involution', function() {
+    eq(properties.involution(S.I), true);
   });
 
 });
--- a/test/properties.js
+++ b/test/properties.js
@@ -1,0 +1,21 @@
+'use strict';
+
+var jsc = require('jsverify');
+var R = require('ramda');
+
+
+var basic = jsc.sum([jsc.integer, jsc.string, jsc.bool, jsc.falsy]);
+
+var useful = jsc.sum([basic, jsc.array(basic), jsc.dict(basic), jsc.fn(basic)]);
+
+exports.idempotent = function(f) {
+  return jsc.checkForall(useful, function(x) {
+    return R.equals(f(f(x)), f(x));
+  });
+};
+
+exports.involution = function(f) {
+  return jsc.checkForall(useful, function(x) {
+    return R.equals(f(f(x)), x);
+  });
+};


test/properties.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+var jsc = require('jsverify');
+
+exports.idempotent = function(f) {
+  return jsc.checkForall(useful, function(x) {
+    return f(x) === f(f(x));
```
rjmk: Are these equality checks OK?

test/properties.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+var jsc = require('jsverify');
```
rjmk: This library and the `jsverify` docs use `jsc` instead of `jsv`. Is that a reference to `JSCheck`?

davidchambers: That's right. [FAQ][1]:

> ### Why do all the examples import the library as jsc instead of jsv?
>
> Does JSC originate with [JSCheck][2]?
>
> __A:__ Yes


[1]: https://github.com/jsverify/jsverify#why-do-all-the-examples-import-the-library-as-jsc-instead-of-jsv
[2]: http://www.jscheck.org/

test/properties.js
davidchambers: I don't think so. `x` is not necessarily a primitive value, presumably, so:

```javascript
> const f = R.sortBy(S.I)
> const x = ['c', 'a', 'b']
> f(x)
['a', 'b', 'c']
> f(f(x))
['a', 'b', 'c']
> f(x) === f(f(x))
false
```

Let's use `R.equals` here.

test/I.js
```diff
@@ -18,4 +20,12 @@ describe('I', function() {
        {'@@type': 'my-package/Foreign'});
   });
 
+  it('is idempotent', function() {
+    ok(properties.idempotent(S.I), true);
```
davidchambers: I think you mean `ok(...)` rather than `ok(..., true)`. I would prefer `eq(..., true)`, though. I appreciate the simplicity of using a single assertion function, and the added correctness guarantee it provides (if `...` does not evaluate to a Boolean value our error will be brought to our attention if we use `eq(..., true)` but not if we use `ok(...)`, which is equivalent to `eq(Boolean(...), true)`).

davidchambers:
> Is a `properties` file OK?

Works for me. :)

> The naming of the arbitraries?

`basic` and `useful`, do you mean? I've no problem with these names. Also, I won't fret about these names as they're not exposed anywhere.

> Using "is idempotent" rather than "satisfies idempotency"?

It reads nicely to me. Thanks for considering this, as whatever we choose is likely to get copied and pasted many times.

> I thought it would be useful to have a concrete example to discuss

Absolutely. Thanks for opening this pull request!

test/I.js
rjmk: Good catch! Yeah the origin of `ok(..., true)` is a misdiagnosed error message (I thought it was about `eq`). Will change

rjmk:
Something that choosing `I` doesn't bring to the fore is the case of non-generic properties. I'd like to test all sensible properties of a function, if possible. But many of them won't be as generic as idempotency and involution. I assume we'll be fine to define some properties in the file itself?

For example, I believe an implementation of reverse just needs to satisfy its type signature, being identity on all singleton lists and "reverseFlip" -- that is `R.equal(reverse(xs.concat(ys), reverse(ys).concat(reverse(xs))))`. I imagine it would be OK to define these properties in the `reverse` test file?

davidchambers:
> I assume we'll be fine to define some properties in the file itself?

Sounds good to me. In fact, we're already testing a couple of [properties][1] of `regexEscape`. :)


[1]: https://github.com/sanctuary-js/sanctuary/blob/v0.11.1/test/regexEscape.js#L36-L46

test/properties.js
```diff
@@ -0,0 +1,21 @@
...
+exports.involution = function(f) {
+  return jsc.checkForall(useful, function(x) {
+    return R.equals(f(f(x)), x);
+  });
+};
+
+
+var basic = jsc.sum([jsc.integer, jsc.string, jsc.bool, jsc.falsy]);
+
+var useful = jsc.sum([basic, jsc.array(basic), jsc.dict(basic), jsc.fn(basic)]);
```
davidchambers: I realize there are conflicting styles, but I like to declare constants before declaring functions which reference them. If this seems reasonable to you, please move these definitions.

One other *very* minor stylistic issue: I like to leave two empty lines between imports and the rest of the code in a module. For example:

```javascript
'use strict';

const R = require('ramda');


const _ = R.__;

//  shout :: String -> String
exports.shout = S.compose(S.concat(_, '!'), S.toUpper);
```

I got this from [PEP 8][1], I believe.


[1]: https://www.python.org/dev/peps/pep-0008/

rjmk: > If this seems reasonable to you, please move these definitions.

Consistency is king!

> I like to leave two empty lines between imports and the rest of the code in a module.

I like it

Will sort tomorrow

davidchambers:
This pull request slipped through the cracks, I'm sorry. LGTM! Would you mind rebasing this branch to verify that the tests are still passing? Feel free to merge this pull request afterwards assuming they are. :)

davidchambers:
:deciduous_tree:

PR #256, opened 2016-09-18
string, array: add S.splitOn and S.joinWith

davidchambers:
``` haskell
splitOn  :: String -> String -> Array String
joinWith :: String -> Array String -> String
```

Related:
- [`split`](http://ramdajs.com/0.22.1/docs/#split), [`join`](http://ramdajs.com/0.22.1/docs/#join) (Ramda)
- [`split`](https://pursuit.purescript.org/packages/purescript-strings/1.1.0/docs/Data.String#v:split), [`joinWith`](https://pursuit.purescript.org/packages/purescript-strings/1.1.0/docs/Data.String#v:joinWith) (PureScript)

At some point I'd like to add a Sanctuary equivalent of [`R.unnest`](http://ramdajs.com/0.22.1/docs/#unnest), which in Haskell is named [`join`](http://hackage.haskell.org/package/base-4.9.0.0/docs/Control-Monad.html#v:join). I'd like to keep `S.join` open for this. Also, `joinWith` is more suggestive of a binary function than `join` is, so I consider it a better name even though it's a little longer.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2473,6 +2473,26 @@
   }
   S.prepend = def('prepend', {}, [a, $.Array(a), $.Array(a)], prepend);
 
+  //# joinWith :: String -> Array String -> String
+  //.
+  //. Joins the strings of the second argument separated by the first argument.
+  //.
+  //. Properties:
+  //.
+  //.   - `forall s :: String, t :: String. S.joinWith(s, S.splitOn(s, t)) = t`
+  //.
+  //. See also [`splitOn`](#splitOn).
+  //.
+  //. ```javascript
+  //. > S.joinWith(':', ['foo', 'bar', 'baz'])
+  //. 'foo:bar:baz'
+  //. ```
+  function joinWith(separator, ss) {
+    return ss.join(separator);
+  }
+  S.joinWith =
+  def('joinWith', {}, [$.String, $.Array($.String), $.String], joinWith);
+
   //# find :: (a -> Boolean) -> Array a -> Maybe a
   //.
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -2672,6 +2672,26 @@
       [a, $.Array(a), $.Array(a)],
       function(x, xs) { return [x].concat(xs); });
 
+  //# joinWith :: String -> Array String -> String
+  //.
+  //. Joins the strings of the second argument separated by the first argument.
+  //.
+  //. Properties:
+  //.
+  //.   - `forall s :: String, t :: String. S.joinWith(s, S.splitOn(s, t)) = t`
```
rjmk: :heart:

safareli:
+1 on preserving `join`

test/joinWith.js
```diff
@@ -0,0 +1,57 @@
...
+                   '                            ^^^^^^\n' +
+                   '                              1\n' +
+                   '\n' +
+                   '1)  1 :: Number, FiniteNumber, NonZeroFiniteNumber, Integer, ValidNumber\n' +
+                   '\n' +
+                   'The value at position 1 is not a member of ‘String’.\n'));
+  });
+
+  it('joins an array of strings with a separator', function() {
+    eq(S.joinWith('', ['a', 'b', 'c']), 'abc');
```
svozza: What happens in the case of an empty array?

davidchambers: The result is `''`. I updated the patch. :)

PR #259, opened 2016-09-24
expand documentation of type signatures and fix inconsistencies

davidchambers:
This is the first of several improvements I plan to extract from #216 in order to make them easier to review. :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,17 +32,59 @@
 //. ## Types
 //.
 //. Sanctuary uses Haskell-like type signatures to describe the types of
-//. values, including functions. `'foo'`, for example, has type `String`;
-//. `[1, 2, 3]` has type `Array Number`. The arrow (`->`) is used to express
-//. a function's type. `Math.abs`, for example, has type `Number -> Number`.
-//. That is, it takes an argument of type `Number` and returns a value of
-//. type `Number`.
+//. values, including functions. `'foo'`, for example, is a member of `String`;
+//. `[1, 2, 3]` is a member of `Array Number`. The double colon (`::`) is used
+//. to mean "is a member of", so one could write:
 //.
-//. [`R.map`][R.map] has type `(a -> b) -> Array a -> Array b`. That is,
-//. it takes an argument of type `a -> b` and returns a value of type
-//. `Array a -> Array b`. `a` and `b` are type variables: applying `R.map`
-//. to a value of type `String -> Number` will result in a value of type
-//. `Array String -> Array Number`.
+//.     'foo' :: String
+//.     [1, 2, 3] :: Array Number
+//.
+//. An identifier may appear to the left of the double colon:
+//.
+//.     Math.PI :: Number
+//.
+//. The arrow (`->`) is used to express a function's type:
+//.
+//.     Math.abs :: Number -> Number
+//.
+//. That states that `Math.abs` is a unary function which takes an argument
+//. of type `Number` and returns a value of type `Number`.
+//.
+//. Some functions are parametrically polymorphic: their types are not fixed.
+//. Type variables are used in the representations of such functions:
+//.
+//.     S.I :: a -> a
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -32,17 +32,58 @@
...
+//. Ditto for each other type variable. In the case of the function above, the
+//. types to which `a` and `b` resolve may be different, but need not be.
+//.
+//. Since all Sanctuary functions are curried it's common to see a binary
+//. function represented as `* -> * -> *`: a unary function which returns
+//. a unary function. This aligns neatly with Haskell, where all functions
+//. are curried. In JavaScript, though, we may wish to represent the types of
+//. functions with arities less than or greater than one. The general form is
+//. `(<input-types>) -> <output-type>`, where `<input-types>` comprises zero
+//. or more comma–space (<code>, </code>) -separated type representations:
```
davidchambers: We need to use `(<code>, </code>)` here because GitHub renders `` (`, `) `` as (`, `) rather than (<code>, </code>).

I first tried the NO-BREAK SPACE character, but our linter complained about [irregular whitespace][1].


[1]: http://eslint.org/docs/rules/no-irregular-whitespace

index.js
```diff
@@ -32,17 +32,58 @@
...
+//. above states that `S.I` takes a value of any type and returns a value of
+//. the same type. Some functions require multiple type variables:
+//.
+//.     S.K :: a -> b -> a
+//.
+//. It must be possible to replace all occurrences of `a` with a concrete type.
+//. Ditto for each other type variable. In the case of the function above, the
+//. types to which `a` and `b` resolve may be different, but need not be.
+//.
+//. Since all Sanctuary functions are curried it's common to see a binary
```
Bradcomp: Currying comes into the picture here with no prelude earlier in the documentation.  While I would think it's probably likely that people using Sanctuary will have knowledge of functional terminology, including currying, my experience has been that it isn't a very common term outside of Haskell and FP circles.  

Depending on the target audience, would it be helpful to add a parenthetical remark here?  Something like: 

> Since all Sanctuary functions are curried (arguments can be passed in one at a time instead of all at once) it's common to see a binary function ...

or would that be overkill?

davidchambers: I :heart: this suggestion. I've updated the pull request. :)

index.js
```diff
@@ -182,25 +228,20 @@
     return f(g(x));
   };
 
-  //  filter :: (Monad m, Monoid m) => ((a -> Boolean), m a) -> m a
+  //  filter :: (Monad m, Monoid m a) => (a -> Boolean, m a) -> m a
```
safareli: Should it be `Monoid (m a)` instead of `Monoid m a`? (not sure)

index.js
```diff
@@ -32,17 +32,59 @@
...
+//. `a` is a type variable. Type variables are not capitalized, so they
+//. are differentiable from type identifiers (which are always capitalized).
+//. By convention type variables have single-character names. The signature
+//. above states that `S.I` takes a value of any type and returns a value of
+//. the same type. Some functions require multiple type variables:
+//.
+//.     S.K :: a -> b -> a
+//.
+//. It must be possible to replace all occurrences of `a` with a concrete type.
+//. Ditto for each other type variable. In the case of the function above, the
```
safareli: Users for which english is not native, maybe it would be easier to understand `The same` instead of `Ditto`

index.js
davidchambers: I wrote this function in Haskell to find out, and you're quite right!

>         ‘Monoid’ is applied to too many type arguments
>         In the type signature for ‘filter'’:
>           filter' :: (Applicative m, Monad m, Monoid m a) =>
>                      (a -> Bool) -> m a -> m a

index.js
davidchambers: Good point. I'll update this. :)

Issue #246, opened 2016-08-19
support esnext module format for tree-shaking purposes


mrtnbroder:
It would be great if sanctuary could introduce a esnext version to support tree-shaking and a more modular approach.

davidchambers:
I'm aware that this is possible, but I don't know any of the details. Can you recommend a resource for learning about this, @mrtnbroder?

mrtnbroder:
If you want, I could give this a shot myself. It's actually not that 'hard' to do, just tedious. But the effort is worth it.

Actually, it's just looking at the modularity of your project and splitting parts of them into their own file/section, so one could `import maybe from 'sanctuary/lib/maybe` directly as an example.

It's a bit cumbersome to do this in es5 syntax, so I'd probably go the way of adding babel and a little `prepublish` script that produces an esnext and an es5 version of this lib.

if you want to learn about the module format in general I can recommend this section of "Understanding ES6" https://leanpub.com/understandinges6/read#leanpub-auto-encapsulating-code-with-modules

mrtnbroder:
Take a look at https://github.com/reactjs/react-router for a pretty good example of how this can be done, @davidchambers !

rjmk:
@mrtnbroder I may well be wrong on this, but is it true that

a) It's as easy to split the file up and do `require('sanctuary/maybe')` in commonJS
and b) It's not necessary to split the file up to get the benefits of tree-shaking, but rather the move to the static `export`s of ES6 is what does the work?

My understanding is that splitting into files and doing the `require('sanctuary/maybe')` allows people to pick 'by hand' any code they wish to bundle and the `export`s allow automated tools to figure out what code you are using

davidchambers:
I do :heart: the lack of any build step in this project, but I do appreciate the advantage this change would provide.

> If you want, I could give this a shot myself.

I'd love to see a proof of concept. Perhaps you could put together a patch to show this working for a handful of functions.

This is all new to me, @rjmk, but I'm very interested to know whether your understanding is correct.

rjmk:
@davidchambers I'd like to echo your sentiments on the lack of build step

mrtnbroder:
> It's not necessary to split the file up to get the benefits of tree-shaking

Correct. It just requires the `export` syntax.

@davidchambers whats wrong with a build step? It doesn't necessarily require a build tool like gulp or something. Babel would be enough to compile to ES5. It's a one-liner in the package.json file.

[...]

PR #261, opened 2016-09-29
#246 modular sanctuary

mrtnbroder:
This is a complete refactor of the library. Every function is now isolated and can be imported directly into your project.

E.g. like this:

``` js
import { toMaybe } from 'sanctuary'
```

However, there are still some problems,  see #246 for further details
- [ ] `create` function must be rethought
- [ ] `def` function must be rethought
- [ ] fix creation of `README.md`
- [x] Make tests pass


davidchambers:
Thanks, @mrtnbroder! This represents a lot of work!

I've had a cursory glance at the changes. I look forward to reading them more carefully tomorrow. :)

src/Alternative/index.js
```diff
@@ -0,0 +1,8 @@
+
+//. ### Alternative
+
+export { and } from './and'
+export { empty } from './empty'
+export { or } from './or'
+export { toBoolean } from './toBoolean'
+export { xor } from './xor'
```
davidchambers: Do you expect people to continue to write `const S = require('sanctuary');` or to cherry-pick the functions of interest?

What's the reason for favouring `src/${category}/${function}.js` over `src/${function}.js`? Having functions listed in some sort of order in the readme and on the website is useful, but the categories do feel arbitrary at times. There are many functions which could justifiably be included in multiple categories. I'm reluctant to make the categories part of the API, as they will be if we support `require('sanctuary/Alternative/or')`.

src/Alternative/and.js
```diff
@@ -0,0 +1,25 @@
+
```
davidchambers: Is there a reason to omit `'use strict';`?

package.json
```diff
@@ -3,31 +3,48 @@
   "version": "0.11.1",
   "description": "Refuge from unsafe JavaScript",
   "license": "MIT",
+  "main": "lib/index",
```
davidchambers: Is there a reason to favour __lib/index.js__ over __index.js__?

package.json
```diff
@@ -3,31 +3,48 @@
...
     "ramda": "0.22.x",
     "sanctuary-def": "0.6.x"
   },
   "devDependencies": {
+    "babel-cli": "^6.14.0",
+    "babel-core": "^6.13.2",
+    "babel-eslint": "^6.1.2",
+    "babel-loader": "^6.2.5",
+    "babel-plugin-add-module-exports": "^0.2.1",
+    "babel-preset-es2015": "^6.13.2",
```
davidchambers: Let's use `"1.2.x"` rather than `"^1.2.3"` for consistency with the way other dependencies are declared.

package.json
```diff
@@ -3,31 +3,48 @@
...
+    "webpack": "^2.1.0-beta.21",
     "xyz": "1.0.x"
   },
   "files": [
     "/LICENSE",
     "/README.md",
-    "/index.js",
+    "es6",
+    "lib",
+    "umd",
```
davidchambers: If a user runs `npm install sanctuary` does she really want all these different versions?

Makefile
```diff
@@ -58,7 +58,7 @@ setup:
 
 .PHONY: test
 test:
-	$(ISTANBUL) cover node_modules/.bin/_mocha -- --recursive --timeout 10000
+	$(ISTANBUL) cover node_modules/.bin/_mocha -- --recursive --timeout 10000 --compilers js:babel-register
 	$(ISTANBUL) check-coverage --branches 100
 ifeq ($(shell node --version | cut -d . -f 1),v6)
 	$(DOCTEST) -- index.js
```
davidchambers: Should this reference __lib/index.js__ instead?

Shard: Note that I moved mocha params to /test/mocha.opts in #298

davidchambers:
I'm going to close this pull request as it has gone stale. Please revive it if you're interested in doing so, @mrtnbroder. :)

PR #264, opened 2016-10-08
list: make the List type constructor simpler and more restrictive

davidchambers:
Commit message:

> The List type constructor enables each polymorphic array/string function to be represented by a single type signature (parametrically polymorphic functions do not require such shenanigans).
>
> `List` is poorly defined, as evidenced by its lengthy explanation. Its use should be limited.
>
> This commit also updates type signatures to use `List _` rather than `[_]` to represent the type of an array/string value.

### DIFF
--- a/index.js
+++ b/index.js
@@ -328,10 +328,7 @@
   var List = $.UnaryType(
     'sanctuary/List',
     function(x) {
-      return x != null &&
-             R.type(x) !== 'Function' &&
-             $.Integer._test(x.length) &&
-             x.length >= 0;
+      return R.contains(R.type(x), ['Array', 'String']);
     },
     function(list) {
       return R.type(list) === 'String' ? [] : Array.prototype.slice.call(list);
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -2385,11 +2403,11 @@
         var Z = negativeZero(end) ? len : end < 0 ? end + len : end;
 
         return Math.abs(start) <= len && Math.abs(end) <= len && A <= Z ?
-          Just(R.slice(A, Z, xs)) :
+          Just(xs.slice(A, Z)) :
```
safareli: I think this will not work with Arguments array, is it intentional?

safareli: I think this will not work with Arguments array, is it intentional?

index.js
```diff
@@ -2320,11 +2317,35 @@
...
+  //.
+  //.     String -> List String -> Maybe Integer
+  //.
+  //. Since `List String` and `String` are interchangeable, the former can be
+  //. replaced with the latter:
+  //.
+  //.     String -> String -> Maybe Integer
+  //.
+  //. It's then apparent that the first argument needn't be a single-character
+  //. string; the correspondence between arrays and strings does not hold.
```
safareli: What if we we also add `Char` type which checks if length of a `String` is 1?
Then we can say that `String` is `List Char`. This will have an issue with emojis `"⏰".length == 2`  but I think we can check it like this:
```js
const isChar = s => s.charAt(1) == "" && s.length != 0
isChar("⏰🗒") // false
isChar("⏰") // true
isChar("") // false
```

index.js
```diff
@@ -328,10 +328,7 @@
...
     'sanctuary/List',
     function(x) {
-      return x != null &&
-             R.type(x) !== 'Function' &&
-             $.Integer._test(x.length) &&
-             x.length >= 0;
+      return R.contains(R.type(x), ['Array', 'String']);
     },
     function(list) {
       return R.type(list) === 'String' ? [] : Array.prototype.slice.call(list);
```
davidchambers: `Array.prototype.slice.call(list)` could now simply read `list`. See #321.

Issue #269, opened 2016-11-01
S.apFirst and S.apSecond


zhangchiqing:
Does it make sense to provide the following two functions?

S.maybeFirst returns the first Maybe value only if both arguments are Just, otherwise return Nothing.

```
S.maybeFirst :: Maybe a -> Maybe b -> Maybe a
> S.maybeFirst(S.Nothing(), S.Nothing())
Nothing()

> S.maybeFirst(S.Just(1), S.Nothing())
Nothing()

> S.maybeFirst(S.Just(1), S.Just('2'))
Just(1)
```

S.maybeSecond returns the second Maybe value only if both arguments are Just, otherwise return Nothing.
```
S.maybeSecond :: Maybe a -> Maybe b -> Maybe b
> S.maybeSecond(S.Nothing(), S.Nothing())
Nothing()

> S.maybeSecond(S.Just(1), S.Nothing())
Nothing()

> S.maybeSecond(S.Just(1), S.Just('2'))
Just('2')
```

Or the above functions can be made by composing existing sanctuary functions? Please advice :)

safareli:
```js
const maybeFirst = first = lift2(a => _ => a)
const maybeSecond = second = lift2(_ => a => a)
```

zhangchiqing:
Thanks for the trick @safareli

davidchambers:
Very nice, @safareli. We can even replace `a => _ => a` with `K`, and `_ => a => a` with `K(I)`:

```javascript
//    maybeFirst :: Maybe a -> Maybe b -> Maybe a
const maybeFirst = S.lift2(S.K);

//    maybeSecond :: Maybe a -> Maybe b -> Maybe b
const maybeSecond = S.lift2(S.K(S.I));
```

safareli:
also they would work on any applicative Either, Future ...

joneshf:
As @safareli points out, `maybeFirst` and `maybeSecond` are `(<*)` and `(*>)` respectively. I don't know if that already exists in the sanctuary ecosystem, but it'd be great to have.

scott-christopher:
> As safareli points out, maybeFirst and maybeSecond are `(<*)` and `(*>)` respectively. I don't know if that already exists in the sanctuary ecosystem, but it'd be great to have.

Only if we can call them `S.icecreamL` and `S.icecreamR`. In all seriousness though, these would be useful additions.

davidchambers:
:icecream:

davidchambers:
[...]

PR #270, opened 2016-11-03
apply: add apFirst, apSecond

zhangchiqing:
This PR will add two functions `apFirst` and `apSecond`, and a new section `Apply` with these two functions in the document.

Please see the discussion here https://github.com/sanctuary-js/sanctuary/issues/269

Ready for review

### DIFF
--- a/index.js
+++ b/index.js
@@ -2224,6 +2224,46 @@
         return toBoolean(x) === toBoolean(y) ? empty(x) : or(x, y);
       });
 
+  //. ### Apply
+  //.
+  //# apFirst :: (Apply a, Apply b) => a -> b -> a
+  //.
+  //. Combine two effectful actions, keeping only the result of the first.
+  //. ```javascript
+  //. > S.apFirst(S.Nothing, S.Nothing)
+  //. Nothing
+  //.
+  //. > S.apFirst(S.Just(1), S.Nothing)
+  //. Nothing
+  //.
+  //. > S.apFirst(S.Just(1), S.Just('2'))
+  //. Just(1)
+  //. ```
+  S.apFirst =
+  def('apFirst',
+    {a: [Apply], b: [Apply]},
+    [a, b, a],
+    S.lift2(S.K));
+
+  //# apSecond :: (Apply a, Apply b) => a -> b -> b
+  //.
+  //. Combine two effectful actions, keeping only the result of the second.
+  //. ```javascript
+  //. > S.apSecond(S.Nothing, S.Nothing)
+  //. Nothing
+  //.
+  //. > S.apSecond(S.Just(1), S.Nothing)
+  //. Nothing
+  //.
+  //. > S.apSecond(S.Just(1), S.Just('2'))
+  //. Just('2')
+  //. ```
+  S.apSecond =
+  def('apSecond',
+    {a: [Apply], b: [Apply]},
+    [a, b, b],
+    S.lift2(S.K(S.I)));
+
   //. ### Logic
 
   //# not :: Boolean -> Boolean
--- a/test/apFirst.js
+++ b/test/apFirst.js
@@ -1,0 +1,63 @@
+'use strict';
+
+var throws = require('assert').throws;
+
+var eq = require('./utils').eq;
+var errorEq = require('./utils').errorEq;
+var S = require('..');
+
+
+describe('apFirst', function() {
+
+  it('is a binary function', function() {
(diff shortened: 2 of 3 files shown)


davidchambers:
Thanks for the pull request, @zhangchiqing!

I like @safareli's suggestion that these functions could be defined in [sanctuary-type-classes][1] (and added to Sanctuary in #216). Would you like to port this pull request to that project?


[1]: https://github.com/sanctuary-js/sanctuary-type-classes

safareli:
What if we use `left/right` instead of first/second? this way the left/right would be consistent with foldLeft/Right as well. Also could define apLeftAll taking array of Applicatives and names would be consistent in this case too: `apLeftAll = reduce(apLeft)` instead of `apLeftAll = reduce(apFirst)`

```
ap1 `apLeft` ap2 `apLeft` ap3 = apLeft(apLeft(ap1, ap2), ap3) = apLeftAll([ap1, ap2, ap3])
```

I'm not saying to add this apLeftAll but the Left/Right i think is better name.
Also it would be nice if we add chainLeft/chainRight too.

davidchambers:
I'm reluctant to go with `ap{Left,Right}` when the PureScript team chose `apply{First,Second}`. Do you think they should have chosen `apply{Left,Right}` instead? These functions are new to me, so I don't have an opinion one way or the other, but I default to following in the footsteps of Haskell and PureScript where applicable.

It's not obvious to me why `apLeftAll` is a better name for the related function than `apFirstAll`.

> Also it would be nice if we add chainLeft/chainRight too.

Absolutely. These needn't be added in the same pull request as `(<*)` and `(*>)`, but we should release all four functions in `sanctuary-type-classes@1.1.0`.

safareli:
Personally I feel `Left/Right` are better than `First/Second`, because `Left/Right` are good names  for any number of arguments. There might not be big difference for `apLeftAll` and `apFirstAll`, but `apRightAll` is better than `apSecondAll` if you have more than 2 elements. We could also define `ap{Left,Right}{3,4,5,..}`,  for example `apRight4` is better than `apSecond4`.

Or if we want to use `First` I think at least we should use `Last` instead of `Second` in which case `apLast4` is basically same as `apRight4`.

So i think `Left/Right` is good name but `First/Last` is better than `First/Second`. If we want to be sortof consistent with purescript maybe `First/Last` is most optimal version.

davidchambers:
Unlike me, you know why these functions are useful, @safareli, so I give your view significant weight. I'd like to know what @joneshf and @scott-christopher think before committing to names, though.

safareli:
> why these functions are useful
I have never used them :D.
They are useful to sequence actions, In case of monadic `>>` one could do 
```
// `remove`, `touch` and `append` are FS actions returning `IO` actions
remove('./a') >> touch('./b') >> append('some text', '/b')
// or 
[remove('./a'), touch('./b'),append('some text', '/b')].reduce(chainRight)
// or 
chainRight(remove('./a'), chainRight(touch('./b'), append('some text', '/b'))
// instead Of
remove('./a').chain(_ => touch('./b').chain(_ => append('some text', '/b')))
```

scott-christopher:
I was initially going to suggest `apLeft`/`apRight` too but then thought they may cause confusion with having something relating specifically to `Either`.

My personal preference would perhaps be `apL`/`apR` to distinguish the difference from `Either`, but I'd be happy with any of the suggestions made so far.

`apFirst`/`apLast` could potentially be used as the names for the suggested `apLeftAll`/`apRightAll`, as they'll end up evaluating to either just the first or last result in the list of applicatives. The only downside with this function is that you will then have to somehow handle the empty list case, so perhaps it's simpler to just provide examples like `apL(ap1, sequence(of, [ap2, ap3]))` and `apR(sequence(of, [ap1, ap2]), ap3)` to cover those use cases instead?

I also don't think we need to be too concerned with aligning with PureScript if there is a preference for something else, I just threw that out as an example where someone has already given it a name.

And lastly, I think `chainLeft`/`chainRight` are perhaps a little redundant, as they would effectively just become aliases of `apL`/`apR`.

zhangchiqing:
>I like @safareli's suggestion that these functions could be defined in sanctuary-type-classes (and added to Sanctuary in #216). Would you like to port this pull request to that project?

OK, I will try to add these functions to `sanctuary-type-classes` and make a new PR for adding them to #216

davidchambers:
Closing in favour of sanctuary-js/sanctuary-type-classes#11

Issue #281, opened 2016-11-13
on :: (b -> b -> c) -> (a -> b) -> a -> a -> c


Bradcomp:
While the general `useWith` and `converge` from Ramda aren't very well typed, I do find them to be useful functions.  One of my main use cases for `useWith` is to emulate Haskell's `on` function (this is also known as the psi combinator).

Example:
```javascript
//Point :: {x, y}
const x = prop('x');

//Point -> Point -> Number
const leftmost = on(min, x);
const rightmost = on(max, x); 
```

I struggle with the parameter order for this function. In Haskell it's typically used in infix form:
```haskell
leftmost = min `on` x
```
As a result, I generally define it using the signature above, even though there could be an argument made that the opposite order makes just as much sense.

Is there enough interest for this to be part of Sanctuary?

davidchambers:
> Is there enough interest for this to be part of Sanctuary?

This function seems useful to me. Would you like to submit a pull request?

I think the type signature you provided is missing one `a ->`, by the way.

Bradcomp:
Thanks!  I was missing the `a`.  I will work on a PR.

PR #282, opened 2016-11-14
Added function S.on

Bradcomp:
Closes https://github.com/sanctuary-js/sanctuary/issues/281

### DIFF
--- a/index.js
+++ b/index.js
@@ -750,6 +750,30 @@
   }
   S.pipe = def('pipe', {}, [$.Array($.Function), a, b], pipe);
 
+  //# on :: (b -> b -> c) -> (a -> b) -> a -> a -> c
+  //.
+  //. Takes a binary function `f`, a unary function `g`, and two
+  //. values `x` and `y`. Returns `f(g(x))(g(y))`.
+  //.
+  //. See also [`on_`](#on_).
+  //.
+  //. ```javascript
+  //. > S.on(S.concat, S.reverse, [1, 2, 3], [4, 5, 6])
+  //. [3, 2, 1, 6, 5, 4]
+  //. ```
+  function on(f, g, x, y) {
+    return on_(uncurry2(f), g, x, y);
+  }
+  S.on = def('on', {}, [$.Function, $.Function, a, a, c], on);
+
+  //# on_ :: ((b, b) -> c) -> (a -> b) -> a -> a -> c
+  //.
+  //. Version of [`on`](#on) accepting uncurried functions.
+  function on_(f, g, x, y) {
+    return f(g(x), g(y));
+  }
+  S.on_ = def('on_', {}, [$.Function, $.Function, a, a, c], on_);
+
   //. ### Maybe type
   //.
   //. The Maybe type represents optional values: a value of type `Maybe a` is
--- a/test/on.js
+++ b/test/on.js
@@ -1,0 +1,21 @@
+'use strict';
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+var rem = require('./internal/rem');
+
+
+describe('on', function() {
+
+  it('is a quaternary function', function() {
+    eq(typeof S.on, 'function');
+    eq(S.on.length, 4);
+  });
+
+  it('passes the last two arguments through the modifying function into the converging function', function() {
+    eq(S.on(rem, S.prop('x'), {x: 5, y: 5}, {x: 3, y: 3}), 2);
+    eq(S.on(S.concat, S.reverse)([1, 2, 3], [4, 5, 6]), [3, 2, 1, 6, 5, 4]);
+  });
+
+});
--- a/test/on_.js
+++ b/test/on_.js
@@ -1,0 +1,20 @@
+'use strict';
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+var rem_ = require('./internal/rem_');
+
+
+describe('on_', function() {
+
+  it('is a quaternary function', function() {
+    eq(typeof S.on_, 'function');
+    eq(S.on_.length, 4);
+  });
+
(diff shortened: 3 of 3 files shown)


davidchambers:
Thanks for the pull request, @Bradcomp! There are a few minor things on which I will comment, but overall it looks very good. :)

index.js
```diff
@@ -713,6 +713,25 @@
       [$.Function, a, b, c, d],
       lift3);
 
+  //# on :: (b -> b -> c) -> (a -> b) -> a -> a -> c
```
davidchambers: We must decide whether the type of the first argument should be `b -> b -> c` or `(b, b) -> c`. Currently the type signature and the implementation contradict each other.

There are a handful of higher-order functions for which we provide both forms. In each case `S.<name>` takes a curried function while `S.<name>_` takes an uncurried function. We could provide both `S.on` and `S.on_`, but `S.on` should take a `b -> b -> c` function regardless.

index.js
```diff
@@ -713,6 +713,25 @@
...
+  //. [3, 2, 1, 6, 5, 4]
+  //. ```
+  function on(f, g, x, y) {
+    return f(g(x), g(y));
+  }
+  S.on =
+  def('on',
+  {},
+  [$.Function, $.Function, a, a, c],
+  on);
```
davidchambers: It would be fine to have this on one line:

```javascript
  S.on = def('on', {}, [$.Function, $.Function, a, a, c], on);
```

index.js
```diff
@@ -713,6 +713,25 @@
...
+
   //. ### Composition
```
davidchambers: Perhaps `on` should live in the __Composition__ section rather than the __Function__ section. When I consider what it does I imagine something like this:

       x           y
       |           |
       |           |
       |           |
       |           |
       |           |
    +-----+     +-----+
    |  g  |     |  g  |
    +-----+     +-----+
        \         /
         \       /
     g(x) \     / g(y)
           \   /
            \ /
          +-----+
          |  f  |
          +-----+
             |
             |
             |
             |
             |
       f(g(x))(g(y))

test/on.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+
+describe('on', function() {
```
davidchambers: Nit: Two empty lines after imports, please.

test/on.js
```diff
@@ -0,0 +1,19 @@
...
+
+  it('has an arity of four', function() {
```
davidchambers: For consistency, let's write `'is a quaternary function'`.

index.js
```diff
@@ -750,6 +750,31 @@
   }
   S.pipe = def('pipe', {}, [$.Array($.Function), a, b], pipe);
 
+  //# on :: (b -> b -> c) -> (a -> b) -> a -> a -> c
+  //.
+  //. Takes a binary function, a unary function, and two values. Returns the
+  //. result of applying the unary function to each value, then passing the
+  //. results through the binary function.
```
davidchambers: "Passing through" doesn't sound right to me for some reason. Combinators are sometimes best explained by their definitions. We could write something along these lines:

_Takes a binary function `f`, a unary function `g`, and two values `x` and `y`. Returns `f(g(x))(g(y))`._

test/on_.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+var rem_ = require('./internal/rem_');
+
+
+describe('on', function() {
```
davidchambers: `s/on/on_/`

test/on_.js
```diff
@@ -0,0 +1,15 @@
...
+
```
davidchambers: Let's include an `'is a quaternary function'` block for this function too.

test/on.js
```diff
@@ -0,0 +1,21 @@
...
+
+
+describe('on', function() {
+
+  it('is a quaternary function', function() {
+    eq(typeof S.on, 'function');
+    eq(S.on.length, 4);
+  });
+
+  it('should pass the last two arguments through the modifying function into the converging function', function() {
```
davidchambers: Nit: I prefer _it does X_ to _it should do X_. In this case, _it passes the last two arguments…_

davidchambers:
Thanks for the updates. Providing both `on` and `on_` will be convenient.

One downside of my preference for rebasing is that I don't receive notifications when pull requests are updated. Please leave a comment after making the second round of changes to let me know I should take one last look. :)

Bradcomp:
Thanks for your feedback!  I've updated the PR based on your comments.

index.js
```diff
@@ -750,6 +750,30 @@
   }
   S.pipe = def('pipe', {}, [$.Array($.Function), a, b], pipe);
 
+  //# on :: (b -> b -> c) -> (a -> b) -> a -> a -> c
+  //.
+  //. Takes a binary function `f`, a unary function `g`, and two
+  //. values `x` and `y`. Returns f(g(x))(g(y)).
```
davidchambers: Let's wrap `f(g(x))(g(y))` in backticks.

davidchambers:
I'll merge this once you've made the final copy tweak. I appreciate your patience. :)

Bradcomp:
Alright!  I think we should be good to go.  Thanks for you time in looking this over and considering it!

davidchambers:
:deciduous_tree:

davidchambers:
```

$ node
undefined
> S.on
on :: (b -> b -> c) -> (a -> b) -> a -> a -> c
> S.on_
on_ :: ((b, b) -> c) -> (a -> b) -> a -> a -> c
```

PR #283, opened 2016-11-14
regexp: revise S.match and add S.matchAll

### LINKED ISSUE #253: S.match
I have a couple of quibbles with `S.match`

> Each match has type Maybe String, where a Nothing represents an unmatched optional capturing group.

This is correct, but maybe a bit misleading? One might be therefore tempted to think that a `Just String` represents a matched optional capturing group, whereas it represents a matched capturing group (not necessarily optional)

The second quibble is that I think the API should be `RegExp -> String -> Maybe (Array (Maybe String), String)`

I realise it's a bit more unwieldy, but I like that it removes `Just ([ Nothing ])` and the like, which don't have any meaning.

I don't think Sanctuary is able to deal with tuples yet, so we can't actually make the change. Perhaps it's useful as a motivating example, though?

### LINKED ISSUE #260: function similar to Python's re.findall
_Parsing HTML with regular expressions is usually a bad idea. I believe there are other uses for this function, though._

---

Assume we wish to generate a table of contents for a neatly formatted HTML document. In order to do so we'd like to find all occurrences of `<(h[1-6]) id="([^"]*)">([^<]*)</\1>` to produce a value of type `Array (Array String)`.

In Python this is easy thanks to [`re.findall`](https://docs.python.org/3/library/re.html#re.findall):

``` python
>>> re.findall(r'<(h[1-6]) id="([^"]*)">([^<]*)</\1>', '<h1 id="foo">Foo</h1> <h2 id="bar">Bar</h2> <h3 id="baz">Baz</h3>')
[('h1', 'foo', 'Foo'), ('h2', 'bar', 'Bar'), ('h3', 'baz', 'Baz')]
```

In JavaScript the [recommended approach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#Finding_successive_matches) is as follows:

``` javascript
const $pattern = /<(h[1-6]) id="([^"]*)">([^<]*)<[/]\1>/g;
const $matches = [];
let match;
while ((match = $pattern.exec(s)) != null) $matches.push(match.slice(1));
```

This is nasty. `RegExp#exec` mutates the `lastIndex` property of `$pattern`, allowing `$pattern.exec(s)` to produce a different result each time it's evaluated.

Sanctuary could provide a `matchAll` function to save users from sullying their programs. It would need to reset the `lastIndex` property before returning, as `test` does (#128).

The function's type should be decided in conjunction with #253. Perhaps:

``` haskell
matchAll :: GlobalRegExp -> String -> Array { match :: String, groups :: Array (Maybe String) }
```

What is the GlobalRegExp type? This function requires a regexp with its `global` flag set to `true`. We could define GlobalRegExp as the type comprising all global regexps.

davidchambers:
Closes #253; closes #260

[`String#match`][1] performs one of two very different operations depending on whether the provided regexp has `global` set to `true`:

```javascript
> 'xxx'.match(/x/)
['x', index: 0, input: 'xxx']

> 'xxx'.match(/x/g)
['x', 'x', 'x']
```

Typical JavaScript nonsense! The sane approach, of course, is to provide two functions:

```purescript
match :: NonGlobalRegExp -> String -> Maybe { match :: String, groups :: Array (Maybe String) }
matchAll :: GlobalRegExp -> String -> Array { match :: String, groups :: Array (Maybe String) }
```

Ideally `RegExp` values would not have `global`-ness baked in. One could then use a pattern in different ways depending on one's requirements. Alas, this is not the case. As a result, `match` requires a `RegExp` value with `global` set to `false`, while `matchAll` requires a `RegExp` value with `global` set to `true`. Having to state that we'd like all the matches in two different ways is unfortunate, but I don't like the idea of ignoring the `global` flag of the provided regexp.

This commit introduces two new types, GlobalRegExp and NonGlobalRegExp. These seem quite useful, so could one day be defined in sanctuary-def instead.

I look forward to feedback on this pull request. It's not often that we make breaking changes, but this change seems worthwhile.

:two_men_holding_hands: :couple: :two_women_holding_hands:

[1]: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/match

### DIFF
--- a/index.js
+++ b/index.js
@@ -318,6 +318,8 @@
     $.NonZeroFiniteNumber,
     $Either,
     Fn($.Unknown, $.Unknown),
+    $.GlobalRegExp,
+    $.NonGlobalRegExp,
     $.Integer,
     $Maybe,
     $.Pair,
@@ -3234,6 +3236,25 @@
 
   //. ### RegExp
 
+  //  Match :: Type
+  var Match = $.RecordType({
+    match: $.String,
+    groups: $.Array($Maybe($.String))
+  });
+
+  //  toMatch :: Array String? -> Match
+  function toMatch(ss) {
+    return {match: ss[0], groups: ss.slice(1).map(toMaybe)};
+  }
+
+  //  withRegex :: (RegExp, () -> a) -> a
+  function withRegex(pattern, thunk) {
+    var lastIndex = pattern.lastIndex;
+    var result = thunk();
+    pattern.lastIndex = lastIndex;
+    return result;
+  }
+
   //# regex :: RegexFlags -> String -> RegExp
   //.
   //. Takes a [RegexFlags][] and a pattern, and returns a RegExp.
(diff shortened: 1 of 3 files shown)


svozza:
I really like this change. I was working with `S.match` recently and this new API would have made my code a lot more readable.

index.js
```diff
@@ -3231,36 +3264,67 @@
...
+  //. []
+  //.
+  //. > S.matchAll(/@([a-z]+)/g, 'Hello, @foo! Hello, @bar! Hello, @baz!')
+  //. [ {match: '@foo', groups: [Just('foo')]},
+  //. . {match: '@bar', groups: [Just('bar')]},
+  //. . {match: '@baz', groups: [Just('baz')]} ]
+  //. ```
+  function matchAll(pattern, s) {
+    return withRegex(pattern, function() {
+      return unfoldr(function(_) {
```
svozza: That is a beautiful implementation,

index.js
```diff
@@ -3187,6 +3201,25 @@
...
+  //  toMatch :: Array String? -> Match
+  var toMatch = function(ss) {
+    return {match: ss[0], groups: ss.slice(1).map(toMaybe)};
+  };
+
+  //  withRegex :: (RegExp, () -> a) -> a
+  var withRegex = function(pattern, thunk) {
+    var lastIndex = pattern.lastIndex;
+    var result = thunk();
+    pattern.lastIndex = lastIndex;
```
svozza: There's no way this mutation will leak out, is there?

svozza: Here be dragons!

Issue #284, opened 2016-11-16
Add maybe_


syaiful6:
Maybe we need a function with signature:
```
maybe :: (unit -> b) -> (a -> b) -> Maybe a -> b
```
Purescript call this maybe', it helpful if we want to compute the Nothing case lazily.

Avaq:
I've done it like this:

```js
maybe(() => computeNothing(), a => () => computeJust(a), ...)()
```

So essentially you're using an `() -> b`  for `b` in the Maybe signature, making the whole thing "lazy":

```hs
maybe :: (() -> b) -> (a -> () -> b) -> Maybe a -> () -> b
```

davidchambers:
I'd like this function! The only things I'd like to tweak are the name (because we already have a function named [`maybe`][1]) and the type of the first argument:

```purescript
maybe_ :: (() -> b) -> (a -> b) -> Maybe a -> b
```

`() -> b` represents a nullary function.

Would you like to submit a pull request, @syaiful6?


[1]: https://sanctuary.js.org/#maybe

syaiful6:
@davidchambers yup, i will submit it shortly.

PR #285, opened 2016-11-17
added S.maybe_ function

syaiful6:
closed https://github.com/sanctuary-js/sanctuary/issues/284. Here the disscussion: how we should call
the function if Maybe known to be Nothing? If we call it without argument, then If we do:

```
S.maybe(S.K(0), R.length, S.Nothing) // it return a function because S.K still wait one more argument
``` 

The first implementation just call the function without argument, let me know if we want to update it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1264,6 +1264,8 @@
   //. a Just, the return value is the result of applying the function to
   //. the Just's value. Otherwise, the first argument is returned.
   //.
+  //. See also [`maybe_`](#maybe_).
+  //.
   //. ```javascript
   //. > S.maybe(0, R.length, S.Just('refuge'))
   //. 6
@@ -1275,6 +1277,28 @@
     return fromMaybe(x, maybe.map(f));
   }
   S.maybe = def('maybe', {}, [b, $.Function, $Maybe(a), b], maybe);
+
+  //# maybe_ :: (() -> b) -> (a -> b) -> Maybe a -> b
+  //.
+  //. Variant of [`maybe`](#maybe) which takes a thunk so the default value
+  //. is only computed if required.
+  //.
+  //. ```javascript
+  //. > global.fib = function fib(n) {
+  //. .   return n <= 1 ? n : fib(n - 2) + fib(n - 1);
+  //. . }
+  //. fib
+  //.
+  //. > S.maybe_(() => fib(30), Math.sqrt, S.Just(1000000))
+  //. 1000
+  //.
+  //. > S.maybe_(() => fib(30), Math.sqrt, S.Nothing)
+  //. 832040
+  //. ```
+  function maybe_(f, g, maybe) {
+    return maybe.isJust ? g(maybe.value) : f();
+  }
+  S.maybe_ = def('maybe_', {}, [$.Function, $.Function, $Maybe(a), b], maybe_);
 
   //# justs :: Array (Maybe a) -> Array a
   //.
--- a/test/maybe_.js
+++ b/test/maybe_.js
@@ -1,0 +1,72 @@
+'use strict';
+
+var throws = require('assert').throws;
+
+var R = require('ramda');
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+var errorEq = require('./internal/errorEq');
+
+
+describe('maybe_', function() {
+
+  it('is a ternary function', function() {
+    eq(typeof S.maybe_, 'function');
+    eq(S.maybe_.length, 3);
+  });
+
+  it('type checks its arguments', function() {
+    throws(function() { S.maybe_(0, R.length); },
+           errorEq(TypeError,
+                   'Invalid value\n' +
+                   '\n' +
+                   'maybe_ :: Function -> Function -> Maybe a -> b\n' +
+                   '          ^^^^^^^^\n' +
+                   '             1\n' +
+                   '\n' +
+                   '1)  0 :: Number, FiniteNumber, Integer, ValidNumber\n' +
+                   '\n' +
+                   'The value at position 1 is not a member of ‘Function’.\n'));
+
+    throws(function() { S.maybe_(R.always(0), [1, 2, 3]); },
+           errorEq(TypeError,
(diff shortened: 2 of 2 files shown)


davidchambers:
Thanks for the pull request, @syaiful6!

> The first implementation just call the function without argument, let me know if we want to update it.

`() => <expr>` seems right to me. Were one to use `S.K(<expr>)`, `<expr>` would be eagerly evaluated. [`maybe`][1] handles the eager use case more elegantly.

```javascript
//  eager
S.maybe(<expr>, f, m)

//  lazy
S.maybe_(() => <expr>, f, m)
```

There are a couple of linting errors to address:

>     /home/ubuntu/sanctuary/test/maybe_.js
>       58:54  error  Missing semicolon                              semi
>       65:2   error  Newline required at end of file but not found  eol-last


[1]: https://sanctuary.js.org/#maybe

index.js
```diff
@@ -1276,6 +1276,25 @@
   }
   S.maybe = def('maybe', {}, [b, $.Function, $Maybe(a), b], maybe);
 
+  //# maybe_ :: (() -> b) -> (a -> b) -> Maybe a -> b
+  //.
+  //. Similiar to `maybe` but for use in cases where you want to compute the
+  //. Nothing case lazily. So, instead takes a value for the first argument,
+  //. it takes a function that will be called without arguments if the `Maybe`
+  //. is known to be `Nothing`.
```
davidchambers: I think we could get away with a shorter description. How does this sound?

    Variant of [`maybe`](#maybe) which takes a thunk so the default value
    is only computed if required.

We should reference `maybe_` in the description of `maybe`:

    See also [`maybe_`](#maybe_).

test/maybe_.js
```diff
@@ -0,0 +1,65 @@
...
+var throws = require('assert').throws;
+
+var R = require('ramda');
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+var errorEq = require('./internal/errorEq');
+
+describe('maybe_', function() {
```
davidchambers: Nit: Two empty lines after imports, please.

test/maybe_.js
```diff
@@ -0,0 +1,65 @@
...
+                   'The value at position 1 is not a member of ‘Maybe a’.\n'));
+  });
+
+  it('can be applied to Nothing', function() {
+    eq(S.maybe_(R.always(0), R.length, S.Nothing), 0)
+  });
+
+  it('can be applied to a Just', function() {
+    eq(S.maybe_(R.always(0), R.length, S.Just([1, 2, 3])), 3);
+  });
```
davidchambers: We should assert that the thunk is not evaluated in this case. Something like this would be fine:

```javascript
var count = 0;
eq(S.maybe_(function() { return count += 1; }, Math.sqrt, S.Just(9)), 3);
eq(count, 0);
```

index.js
syaiful6: that's sound better i think

syaiful6:
@davidchambers thanks, i updated the docs and fix the linting error. For the doctest, i am using ```factorial```, because when i am using fib, it doesn't fit with 79 max-line. Note: ``fct``` stand for factorial.

test/maybe_.js
```diff
@@ -0,0 +1,72 @@
...
+    eq(S.maybe_(R.always(0), R.length, S.Nothing), 0);
+  });
+
+  it('can be applied to a Just', function() {
+    eq(S.maybe_(R.always(0), R.length, S.Just([1, 2, 3])), 3);
+  });
+
+  it('thunk only evaluated when it required', function() {
+    var count = 0;
+    S.maybe_(function() { return count += 1; }, Math.sqrt, S.Just(10));
```
davidchambers: We could assert that the result is √10. We could use `S.Just(9)` rather than `S.Just(10)` to avoid decimals.

index.js
```diff
@@ -1276,6 +1276,25 @@
...
+  //.
+  //. ```javascript
+  //. > S.maybe_(() => 0, R.length, S.Just('refuge'))
+  //. 6
+  //.
+  //. > S.maybe_(() => 0, R.length, S.Nothing)
+  //. 0
+  //. ```
```
davidchambers: > For the doctest, i am using `factorial`, because when i am using fib, it doesn't fit with 79 max-line.

If you'd like to use `fib` (which is more computationally complex than `fct`, I believe), we could use the undocumented line wrapping functionality (davidchambers/doctest#67):

```javascript
> global.fib = function fib(n) {
.   return n <= 1 ? n : fib(n - 2) + fib(n - 1);
. }
fib
```

test/maybe_.js
davidchambers: I'm sorry I was not clearer here. This is the [PEP 8][1] -inspired format I like to use:

```javascript
'use strict';

//  Standard library imports, in alphabetical order

//  Third-party imports, in alphabetical order

//  Relative imports, in alphabetical order


//  Code starts here
```

In this case there's actually a fourth import group for the test helpers:

```javascript
'use strict';

var throws = require('assert').throws;

var R = require('ramda');

var S = require('..');

var eq = require('./internal/eq');
var errorEq = require('./internal/errorEq');


describe('maybe_', function() {
//  ...
});
```

I realize many projects don't follow this system (or any other system), but I find this hierarchical approach useful.


[1]: https://www.python.org/dev/peps/pep-0008/

davidchambers:
Thanks for the updates, @syaiful6. :)

I hope my comments don't seem pedantic. I believe it's important to maintain consistency as a project grows, particularly for a project such as this which is likely to continue to grow in size for some time.

Taking the time to mention features of doctest and other tools used in the project will hopefully make it easier to contribute to the project again in the future. :)

syaiful6:
@davidchambers no worry, i am quite happy to learn these things. 👍

davidchambers:
:deciduous_tree:

davidchambers:
```

$ node
undefined
> S.maybe_
maybe_ :: (() -> b) -> (a -> b) -> Maybe a -> b
```

PR #289, opened 2016-11-19
Curry: Add functions curry2 through curry5

### LINKED ISSUE #277: curry2 :: ((a, b) -> c) -> a -> b -> c
Now that Sanctuary is close to being an alternative to Ramda rather than a library intended to be used alongside it, we should consider which Ramda functions we would miss were we to remove Ramda from our projects. [`R.curry`][1] and [`R.curryN`][2] are certainly important functions. [`def`][3] performs currying, of course, but not every project which depends on Sanctuary also depends on sanctuary-def.

The type signatures of these Ramda functions are misleading:

```haskell
curry  :: (* -> a) -> (* -> a)
curryN :: Number -> (* -> a) -> (* -> a)
```

Accurate type signatures are an important tool for reasoning about functions. The Sanctuary functions would be much simpler:

```haskell
curry2 :: ((a, b) -> c) -> a -> b -> c
curry3 :: ((a, b, c) -> d) -> a -> b -> c -> d
curry4 :: ((a, b, c, d) -> e) -> a -> b -> c -> d -> e
curry5 :: ((a, b, c, d, e) -> f) -> a -> b -> c -> d -> e -> f
```

We must choose the maximum supported arity. I don't think it's necessary to go all the way to `curry5`; `curry2` and `curry3` should suffice. If you call recall using `R.curry` or `R.curryN` to produce a function of arity greater than 3, please speak up.

:curry:

[1]: http://ramdajs.com/0.22.1/docs/#curry
[2]: http://ramdajs.com/0.22.1/docs/#curryN
[3]: https://github.com/sanctuary-js/sanctuary-def

laduke:
Hi, first pull request here. Thanks for the hand-holding. This closes #277. 
Things not complete or I'm not sure about: 
- [x] Examples for curry3 - 5. 
  Looks like all of the examples in the readme are one liners. 
- [x] Documentation text is probably too short, but I'm not the one that should be explaining currying to anyone. 
- [x] I probably skipped some test cases. Is there an S.__ or similar that needs to be tested? 
- [x] The definition for`var e = $.TypeVariable('e');` is skipped and I don't know why. So curry4 and 5 don't have an 'e': `S.curry4 = def('curry4', {}, [$.Function, a, b, c, d, f], curry4);`
- [x] Not sure what variable names people traditionally use for functions of large arity. 
  I have `function(v, w, x, y, z)`...
- [x] Should I `make` the README and commit, or does that happen later?

### DIFF
--- a/index.js
+++ b/index.js
@@ -300,6 +300,7 @@
   var b = $.TypeVariable('b');
   var c = $.TypeVariable('c');
   var d = $.TypeVariable('d');
+  var e = $.TypeVariable('e');
   var f = $.TypeVariable('f');
   var l = $.TypeVariable('l');
   var r = $.TypeVariable('r');
@@ -631,6 +632,89 @@
   S.S = def('S', {}, [$.Function, $.Function, a, c], S_);
 
   //. ### Function
+
+  //# curry2 :: ((a, b) -> c) -> a -> b -> c
+  //.
+  //. Curries the given binary function.
+  //.
+  //. ```javascript
+  //. > R.map(S.curry2(Math.pow)(10), [1, 2, 3])
+  //. [10, 100, 1000]
+  //.
+  //. > R.map(S.curry2(Math.pow, 10), [1, 2, 3])
+  //. [10, 100, 1000]
+  //. ```
+  function curry2(f, x, y) {
+    return f(x, y);
+  }
+  S.curry2 = def('curry2', {}, [$.Function, a, b, c], curry2);
+
+  //# curry3 :: ((a, b, c) -> d) -> a -> b -> c -> d
+  //.
+  //. Curries the given ternary function.
+  //.
+  //. ```javascript
+  //. > global.replaceString = S.curry3((what, replacement, string) =>
+  //. .   string.replace(what, replacement)
+  //. . )
+  //. replaceString
+  //.
+  //. > replaceString('banana')('orange')('banana icecream')
+  //. 'orange icecream'
+  //.
+  //. > replaceString('banana', 'orange', 'banana icecream')
+  //. 'orange icecream'
+  //. ```
+  function curry3(f, x, y, z) {
+    return f(x, y, z);
+  }
+  S.curry3 = def('curry3', {}, [$.Function, a, b, c, d], curry3);
+
+  //# curry4 :: ((a, b, c, d) -> e) -> a -> b -> c -> d -> e
+  //.
+  //. Curries the given quaternary function.
+  //.
+  //. ```javascript
+  //. > global.createRect = S.curry4((x, y, width, height) =>
+  //. .   ({x, y, width, height})
+  //. . )
+  //. createRect
+  //.
+  //. > createRect(0)(0)(10)(10)
+  //. {x: 0, y: 0, width: 10, height: 10}
+  //.
+  //. > createRect(0, 0, 10, 10)
+  //. {x: 0, y: 0, width: 10, height: 10}
+  //. ```
+  function curry4(f, w, x, y, z) {
+    return f(w, x, y, z);
+  }
+  S.curry4 = def('curry4', {}, [$.Function, a, b, c, d, e], curry4);
+
+  //# curry5 :: ((a, b, c, d, e) -> f) -> a -> b -> c -> d -> e -> f
+  //.
+  //. Curries the given quinary function.
+  //.
+  //. ```javascript
+  //. > global.toUrl = S.curry5((protocol, creds, hostname, port, pathname) =>
+  //. .   protocol + '//' +
+  //. .   S.maybe('', _ => _.username + ':' + _.password + '@', creds) +
+  //. .   hostname +
+  //. .   S.maybe('', S.concat(':'), port) +
+  //. .   pathname
+  //. . )
+  //. toUrl
+  //.
+  //. > toUrl('https:')(S.Nothing)('example.com')(S.Just('443'))('/foo/bar')
+  //. 'https://example.com:443/foo/bar'
+  //.
+  //. > toUrl('https:', S.Nothing, 'example.com', S.Just('443'), '/foo/bar')
+  //. 'https://example.com:443/foo/bar'
+  //. ```
+  function curry5(f, v, w, x, y, z) {
+    return f(v, w, x, y, z);
+  }
+  S.curry5 = def('curry5', {}, [$.Function, a, b, c, d, e, r], curry5);
 
   //# flip :: ((a, b) -> c) -> b -> a -> c
   //.
--- a/test/curry3.js
+++ b/test/curry3.js
@@ -1,0 +1,19 @@
+'use strict';
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+
+
+test('curry3', function() {
+
+  eq(typeof S.curry3, 'function');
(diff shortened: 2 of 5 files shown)


davidchambers:
Thanks for the pull request, @laduke! It's exciting to have received pull requests from three new code contributors in the past week!

I edited this pull request's description to include the substring `closes #277` so the issue this pull request addresses will automatically be closed when this pull request is merged. See [Closing Issues via Pull Requests][1] for details. :)


[1]: https://github.com/blog/1506-closing-issues-via-pull-requests

davidchambers:
> The definition for`var e = $.TypeVariable('e');` is skipped and I don't know why.

We simply haven't needed this so many [unary type variables][1] before, but we're free to define `e`. :)

> Should I `make` the README and commit, or does that happen later?

The readme gets updated automatically at release time. See https://github.com/sanctuary-js/sanctuary-set/pull/1#issuecomment-258782119 if you're interested in the details.


[1]: https://github.com/sanctuary-js/sanctuary-def#UnaryTypeVariable

index.js
```diff
@@ -632,6 +632,54 @@
 
   //. ### Function
 
+  //# curry2 :: ((a, b) -> c) -> a -> b -> c
+  //.
+  //. Curry a binary function.
+  //. ```javascript
```
davidchambers: Please include an "empty" (`//.`) line between the description and the code block.

index.js
```diff
@@ -632,6 +632,54 @@
 
   //. ### Function
 
+  //# curry2 :: ((a, b) -> c) -> a -> b -> c
+  //.
+  //. Curry a binary function.
```
davidchambers: I like these short descriptions. For consistency we should use _Curr**ies** a binary function_. Also, we could be more specific about which function gets curried:

_Curries the given binary function._

laduke:
Thank you for your patience @davidchambers.

index.js
```diff
@@ -632,6 +633,78 @@
...
+
+  //# curry4 :: ((a, b, c, d) -> e) -> a -> b -> c -> d -> e
+  //.
+  //. Curries the given quaternary function.
+  //.
+  //. ```javascript
+  //. > global.createRect = S.curry4((x, y, width, height) => (
+  //. .   {x, y, width, height}
+  //. .   )
+  //. . )
```
davidchambers: I'd prefer this formatting:

```javascript
> global.createRect = S.curry4((x, y, width, height) =>
.   ({x, y, width, height})
. )
```

index.js
```diff
@@ -632,6 +633,78 @@
 
   //. ### Function
 
+  //# curry2 :: ((a, b) -> c) -> a -> b -> c
+  //.
+  //. Curries the given binary function.
+  //.
+  //. ```javascript
+  //. > R.map(S.curry2(Math.pow, 10), [1, 2, 3])
+  //. [ 10, 100, 1000 ]
```
davidchambers: Let's use `[10, 100, 1000]` here (without the extra spaces).

index.js
```diff
@@ -632,6 +633,78 @@
...
+  //.
+  //. ```javascript
+  //. > global.createRect = S.curry4((x, y, width, height) => (
+  //. .   {x, y, width, height}
+  //. .   )
+  //. . )
+  //. createRect
+  //.
+  //. > createRect(0)(0)(10)(10)
+  //. { x: 0, y: 0, width: 10, height: 10 }
```
davidchambers: Let's use `{x: 0, y: 0, width: 10, height: 10}` here (without the extra spaces).

laduke:
I've become spoiled by jsfmt. Sorry about those.

davidchambers:
We should make it clear that the functions are curried Ramda-style. I'd like to see two applications of each function. For example:

```javascript
> toUrl('https:')(S.Nothing)('example.com')(S.Just('443'))('/foo/bar')
'https://example.com:443/foo/bar'

> toUrl('https:', S.Nothing, 'example.com', S.Just('443'), '/foo/bar')
'https://example.com:443/foo/bar'
```

index.js
```diff
@@ -632,6 +633,77 @@
...
+  //. . )
+  //. toUrl
+  //.
+  //. > toUrl('https:')(S.Nothing)('example.com')(S.Just('443'))('/foo/bar')
+  //. 'https://example.com:443/foo/bar'
+  //. ```
+  function curry5(f, v, w, x, y, z) {
+    return f(v, w, x, y, z);
+  }
+  S.curry5 = def('curry5', {}, [$.Function, a, b, c, d, e, f], curry5);
```
davidchambers: I'd like to use `r` (result?) in place of `f`, because in another pull request `f` will become a [unary type variable][1]. In other words, its type will be `Type -> Type` rather than `Type`, and it will no longer be suitable in this context. Using a different type variable now will save me from having to change it in the other pull request.


[1]: https://github.com/sanctuary-js/sanctuary-def#UnaryTypeVariable

laduke:
Chippin' away. Thanks for that script.

test/curry3.js
```diff
@@ -0,0 +1,19 @@
...
+test('curry3', function() {
+
+  eq(typeof S.curry3, 'function');
+  eq(S.curry3.length, 4);
+
+  var curried = S.curry3(function(x, y, z) { return x * y * z; });
+  eq(curried(1, 2, 3), 6);
+  eq(curried(1)(2)(3), 6);
+  eq(curried(1, 2)(3), 6);
+  eq(curried(1)(2, 3), 6);
```
davidchambers: For consistency, let's use the script to generate the assertions for `curry2` and `curry3` as well. You have all the combinations, but my obsessive streak wants to see them consistently ordered. :stuck_out_tongue_winking_eye:

davidchambers:
Looking good, @laduke! After addressing the final comment could you squash the commits and rebase on `master`? We'll then be ready to merge, I think. :)

laduke:
Oops!

test/curry3.js
laduke: I also changed them to use + instead of *  in the curried function. ;)

test/curry5.js
```diff
@@ -0,0 +1,30 @@
...
+  eq(curried(1, 2)(3)(4)(5), 15);
+  eq(curried(1)(2, 3, 4, 5), 15);
+  eq(curried(1)(2, 3, 4)(5), 15);
+  eq(curried(1)(2, 3)(4, 5), 15);
+  eq(curried(1)(2, 3)(4)(5), 15);
+  eq(curried(1)(2)(3, 4, 5), 15);
+  eq(curried(1)(2)(3, 4)(5), 15);
+  eq(curried(1)(2)(3)(4, 5), 15);
+  eq(curried(1)(2)(3)(4)(5), 15);
+});
```
laduke: I deserve every word of it.

laduke:
🎉

davidchambers:
```

$ node
undefined
> S.curry2
curry2 :: ((a, b) -> c) -> a -> b -> c
> S.curry3
curry3 :: ((a, b, c) -> d) -> a -> b -> c -> d
> S.curry4
curry4 :: ((a, b, c, d) -> e) -> a -> b -> c -> d -> e
> S.curry5
curry5 :: ((a, b, c, d, e) -> r) -> a -> b -> c -> d -> e -> r
```

PR #293, opened 2016-11-25
Change pluck to be more prop-like

### LINKED ISSUE #291: pluck :: Accessible a => TypeRep b -> String -> Array a -> Array (Maybe b)
In its infancy Sanctuary was a companion to Ramda, focused on providing safe versions of unsafe Ramda functions. Sanctuary is now close to being a Ramda alternative, and provides many more functions than it once did. It's important to occasionally remove poorly defined and/or unnecessary functions so the library becomes more coherent over time, even as it grows.

[`pluck`][1] provides a slightly more succinct way to express a composition of `map` and `get`:

```javascript
> S.pluck(Number, 'x', [{x: 1}, {x: 2}, {x: '3'}, {x: null}, {}])
[Just(1), Just(2), Nothing(), Nothing(), Nothing()]

> R.map(S.get(Number, 'x'), [{x: 1}, {x: 2}, {x: '3'}, {x: null}, {}])
[Just(1), Just(2), Nothing(), Nothing(), Nothing()]
```

I have several issues with `pluck`:

  - it doesn't provide much of a readability improvement over `map` + `get`;
  - we provide a shorthand for `map` + `get` but not for `map` + `gets`;
  - we provide a shorthand for `map` + `get` but not for `map` + `prop`; and
  - type representatives are a complex concept best referenced in as few places as possible.

If you're okay with `pluck` being removed from the library, vote :thumbsdown:; if you would like the function to stay, vote :thumbsup:. If in addition to keeping `pluck` you think we should add `plucks` as shorthand for `map` + `gets`, please say so. Thanks. :)

[1]: https://sanctuary.js.org/#pluck

svozza:
Closes #291

### DIFF
--- a/index.js
+++ b/index.js
@@ -2651,27 +2651,27 @@
   }
   S.find = def('find', {}, [$.Function, $.Array(a), $Maybe(a)], find);
 
-  //# pluck :: Accessible a => TypeRep b -> String -> Array a -> Array (Maybe b)
+  //# pluck :: Accessible a => String -> Array a -> Array b
   //.
-  //. Takes a [type representative](#type-representatives), a property name,
-  //. and an array of objects and returns an array of equal length. Each
-  //. element of the output array is Just the value of the specified property
-  //. of the corresponding object if the value is of the specified type
-  //. (according to [`is`](#is)); Nothing otherwise.
-  //.
-  //. See also [`get`](#get).
+  //. Combines [`map`][R.map] and [`prop`](#prop). `pluck(k, xs)` is equivalent
+  //. to `map(prop(k), xs)`.
   //.
   //. ```javascript
-  //. > S.pluck(Number, 'x', [{x: 1}, {x: 2}, {x: '3'}, {x: null}, {}])
-  //. [Just(1), Just(2), Nothing, Nothing, Nothing]
+  //. > S.pluck('x', [{x: 1}, {x: 2}, {x: 3}])
+  //. [1, 2, 3]
   //. ```
-  function pluck(typeRep, key, xs) {
-    return xs.map(function(x) { return get(typeRep, key, x); });
+  function pluck(key, xs) {
+    return xs.map(function(x, idx) {
+      if (key in Object(x)) return x[key];
+      throw new TypeError('‘pluck’ expected object at index ' + idx +
+                          ' to have a property named ‘' + key + '’; ' +
+                          R.toString(x) + ' does not');
+    });
   }
   S.pluck =
   def('pluck',
       {a: [Accessible]},
-      [TypeRep, $.String, $.Array(a), $.Array($Maybe(b))],
+      [$.String, $.Array(a), $.Array(b)],
       pluck);
 
   //# reduce :: Foldable f => (a -> b -> a) -> a -> f b -> a
@@ -2765,6 +2765,8 @@
   //. lacks the specified property, a type error is thrown.
   //.
   //. For accessing properties of uncertain objects, use [`get`](#get) instead.
+  //.
+  //. See also [`pluck`](#pluck).
   //.
   //. ```javascript
   //. > S.prop('a', {a: 1, b: 2})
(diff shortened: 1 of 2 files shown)


test/pluck.js
```diff
@@ -11,48 +11,56 @@ var throws = require('./internal/throws');
...
+         TypeError,
+         '‘pluck’ expected object at index 1 to have a property named ‘a’; {"b": 2} does not');
+
+  eq(S.pluck('x', []), []);
+  eq(S.pluck('x', [{x: 1}, {x: 2}, {x: 3}]), [1, 2, 3]);
 
-  eq(S.pluck(Array, 'x', [{x: vm.runInNewContext('[0]')}]), [S.Just([0])]);
-  eq(S.pluck(vm.runInNewContext('Array'), 'x', [{x: [0]}]), [S.Just([0])]);
+  eq(S.pluck('x', [{x: vm.runInNewContext('[0]')}]), [[0]]);
+  eq(S.pluck(vm.runInNewContext("'x'"), [{x: [0]}]), [[0]]);
```
davidchambers: These `vm` test are no longer relevant. Let's remove them.

index.js
```diff
@@ -2651,27 +2651,29 @@
...
-  //. and an array of objects and returns an array of equal length. Each
-  //. element of the output array is Just the value of the specified property
-  //. of the corresponding object if the value is of the specified type
-  //. (according to [`is`](#is)); Nothing otherwise.
-  //.
-  //. See also [`get`](#get).
+  //. Takes a property name and an array of objects and returns an array in
+  //. which each element is the value of the specified property of the
+  //. corresponding object. If for some reason the object lacks the specified
+  //. property, a type error is thrown.
```
davidchambers: I'd like to leverage the documentation for existing functions:

    Combines [`map`][R.map] and [`prop`](#prop). `pluck(k, xs)` is equivalent
    to `map(prop(k), xs)`.


    [R.map]: http://ramdajs.com/docs/#map

index.js
```diff
@@ -2651,27 +2651,29 @@
...
+  //. > S.pluck('x', [{x: 1}, {x: 2}, {x: 3}])
+  //. [1, 2, 3]
   //. ```
-  function pluck(typeRep, key, xs) {
-    return xs.map(function(x) { return get(typeRep, key, x); });
+  function pluck(key, xs) {
+    return xs.map(function(obj, idx) {
+      if (key in Object(obj)) return obj[key];
+      throw new TypeError('‘pluck’ expected object at index ' + idx + ' to ' +
+                          'have a property named ‘' + key + '’; ' +
```
davidchambers: Let's move `' to '` down one line for more pleasing line wrapping (saving one concatenation operation in the process).

davidchambers:
> refactor pluck to be more prop-like

Strictly speaking this is not refactoring since we're changing the function's behaviour. Would you mind updating the commit message and pull request title?

index.js
svozza: Yes, and let's remember to change `R.map` to `S.map` when we release 0.12.

index.js
```diff
@@ -2651,27 +2651,29 @@
...
   //. ```javascript
-  //. > S.pluck(Number, 'x', [{x: 1}, {x: 2}, {x: '3'}, {x: null}, {}])
-  //. [Just(1), Just(2), Nothing, Nothing, Nothing]
+  //. > S.pluck('x', [{x: 1}, {x: 2}, {x: 3}])
+  //. [1, 2, 3]
   //. ```
-  function pluck(typeRep, key, xs) {
-    return xs.map(function(x) { return get(typeRep, key, x); });
+  function pluck(key, xs) {
+    return xs.map(function(obj, idx) {
```

index.js
```diff
@@ -3500,6 +3500,7 @@
 //. [R.equals]:         http://ramdajs.com/docs/#equals
 //. [Ramda]:            http://ramdajs.com/
 //. [RegexFlags]:       https://github.com/sanctuary-js/sanctuary-def#regexflags
+//. [R.map]:            http://ramdajs.com/docs/#map
```
davidchambers: Could you insert this line after `[R.equals]` instead? I like to type `vip:sort u` in Vim to sort this "paragraph", so keeping this list sorted will avoid noise in future diffs.

index.js
```diff
@@ -2651,27 +2651,27 @@
...
   //.
-  //. Takes a [type representative](#type-representatives), a property name,
-  //. and an array of objects and returns an array of equal length. Each
-  //. element of the output array is Just the value of the specified property
-  //. of the corresponding object if the value is of the specified type
-  //. (according to [`is`](#is)); Nothing otherwise.
-  //.
-  //. See also [`get`](#get).
+  //. Combines [`map`][R.map] and [`prop`](#prop). `pluck(k, xs)` is equivalent
+  //. to `map(prop(k), xs)`.
```
davidchambers: Shall we add ``See also [`pluck`](#pluck).`` to the `prop` docstring?

index.js

index.js
svozza: Yep, good idea.

davidchambers:
:deciduous_tree:

svozza:
It's good to be back. ;)

PR #294, opened 2016-11-27
maybe, either: make type representatives uncallable

davidchambers:
This obviates the need for the sentinel value. :)

How does this look to you, @safareli?

### DIFF
--- a/index.js
+++ b/index.js
@@ -206,8 +206,6 @@
 
   'use strict';
 
-  var sentinel = {};
-
   var _slice = Array.prototype.slice;
   var _toString = Object.prototype.toString;
 
@@ -337,8 +335,8 @@
     function(x) {
       return R.type(x) === 'Function' ||
              (x != null &&
-              R.type(x.name) === 'String' &&
-              R.type(x.test) === 'Function');
+              x.prototype != null &&
+              R.type(x.prototype['@@type']) === 'String');
     }
   );
 
@@ -364,7 +362,7 @@
 
   /* eslint-disable indent */
 
-  var S = {EitherType: $Either, MaybeType: $Maybe};
+  var S = {};
 
   //# create :: { checkTypes :: Boolean, env :: Array Type } -> Module
   //.
@@ -498,12 +496,11 @@
   //. false
   //. ```
   function is(typeRep, x) {
-    if (_toString.call(typeRep.prototype['@@type']) === '[object String]') {
-      return x != null && x['@@type'] === typeRep.prototype['@@type'];
-    } else {
-      var match = /function (\w*)/.exec(typeRep);
-      return match != null && match[1] === type(x);
-    }
+    return type(x) ===
+           (typeRep.prototype != null &&
+            _toString.call(typeRep.prototype['@@type']) === '[object String]' ?
+              typeRep.prototype['@@type'] :
+              typeRep.name);
   }
   S.is = def('is', {}, [TypeRep, $.Any, $.Boolean], is);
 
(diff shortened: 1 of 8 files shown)


index.js
```diff
@@ -206,7 +206,10 @@
 
   'use strict';
 
-  var sentinel = {};
+  /* istanbul ignore next */
+  var global_ =
+  typeof self === 'object' && self != null && self.self === self ? self
+                                                                 : global;
```
davidchambers: I never feel good writing this sort of code. It looks error-prone. Let me know if you think there's a better approach.

index.js
```diff
@@ -335,10 +338,11 @@
   var TypeRep = $.NullaryType(
     'sanctuary/TypeRep',
     function(x) {
-      return R.type(x) === 'Function' ||
-             (x != null &&
-              R.type(x.name) === 'String' &&
-              R.type(x.test) === 'Function');
```
davidchambers: The three lines above allowed a `Type` value (such as `$.Number` or `S.MaybeType`) to be used as a type representative. I'm almost certain we should not support this.

index.js
```diff
@@ -335,10 +338,11 @@
...
+      return R.type(x) === 'Function'
+          && R.type(global_[x.name]) === 'Function'
```
davidchambers: The two lines above handle built-in type representatives such as `Array`:

```javascript
> R.type(Array) === 'Function' && R.type(global[Array.name]) === 'Function'
true
```

index.js
```diff
@@ -335,10 +338,11 @@
...
+          || x != null
+          && x.prototype != null
+          && R.type(x.prototype['@@type']) === 'String';
```
davidchambers: The three lines above handle user-defined type representatives, identified by the presence of a `@@type` property on the object's `prototype`.

index.js
```diff
@@ -364,7 +368,7 @@
 
   /* eslint-disable indent */
 
-  var S = {EitherType: $Either, MaybeType: $Maybe};
+  var S = {};
```
davidchambers: I moved these property assignments alongside their respective docstrings.

index.js
```diff
@@ -498,12 +502,11 @@
   //. false
   //. ```
   function is(typeRep, x) {
-    if (_toString.call(typeRep.prototype['@@type']) === '[object String]') {
-      return x != null && x['@@type'] === typeRep.prototype['@@type'];
-    } else {
-      var match = /function (\w*)/.exec(typeRep);
```
davidchambers: I don't know why we were doing this rather than simply `typeRep.name`!

index.js
safareli: Is it sort of breaking change? if one was using  `data.task` for example user should itself add the `@@type` prop to it's prototype

index.js
```diff
@@ -1547,7 +1552,7 @@
   //. Left('Cannot divide by zero')
   //. ```
   function Left(x) {
-    return new Either(sentinel, 'Left', x);
+    return new _Either('Left', x);
```
safareli: I think `Left('x').constructor` would be `_Either` when by conforming FL it should be `Either`

index.js
davidchambers: I like it! :thumbsup:

index.js
davidchambers: Good point. Let's reinstate the `R.type(x) === 'Function'` predicate, so we continue to treat *any* function as a type representative. This will obviate the need for `R.type(global_[x.name]) === 'Function'` and a way to locate the global object, so I'm very much in favour!

index.js
davidchambers: Good point. We may need to tweak things in #232 in order to accommodate the current FL spec.

index.js
safareli: 👍 
so TypeRep will check if `x` is a Function or if it has `.prototype[@@type]` ?

davidchambers:
Please have another look, @safareli. :)

index.js
davidchambers: Yes, that's exactly right.

index.js
safareli: we could be fixed by:
```js
_Either.prototype.constructor = Either
```

safareli:
Looks good.
one thing left is FL constructor issue, which i think could be fixed like this:
```js
function makeTypeRep(ctr){
  var typeRep = { prototype: ctr.prototype }
  ctr.constructor = typeRep
  return typeRep;
}
var Maybe = S.Maybe = makeTypeRep(_Maybe)
var Either = S.Either = makeTypeRep(_Either)
```

index.js
davidchambers: That's elegant! I've pushed a commit which includes this change and some straightforward test cases. :)

safareli:
LGTM 👍

index.js
davidchambers: There actually *was* a good reason for this: `Function#name` is not specified by ES5. This change introduced a regression which was fortunately caught by the browser tests @Shard has hooked up in #298.

Mark, could you revert this change to `is` in #298? I believe it will fix some—perhaps all—of the test failures in Internet Explorer.

PR #298, opened 2016-12-02
Implement browser testing

### LINKED ISSUE #292: Drop Support For Node 0.10
Node 0.10 is now [officially unsupported](https://github.com/nodejs/LTS) so I think we should stop testing against it and use the slot freed up to test Node 7. Node 0.12 is EOL in a month's time and I think we should drop support for that too: this will allow us to write our tests in ES6 without involving Babel, the main benefit being string templates which will make testing our error messages _much_ easier.

Shard:
This PR when it's finished will add support for running the unit tests through Karma and hookup CircleCI to execute them on multiple browsers via SauceLabs, so that we can properly prove browser support and potentially resolve #292.

At the moment I have karma successfully being able to execute the test suite using browserify to handle the test suite translation to the browser (module loading mostly). Browserify adds a little bit of weight to DevDeps but does do a good job of bundling the tests correctly.

One thing I will probably want your input on @davidchambers is how to go about integrating the `karma start` command into the Makefile. Since we can run the tests in many different ways via karma:

- Simple quick test run using a headless browser (PhantomJS)
- Test manually by bringing your own browser
- Send tests to Saucelabs for cross browser testing

I'm thinking maybe put something like the first in the current test command and creating a new command to send off the tests.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -2,15 +2,16 @@ dependencies:
   override:
     - printf '%s\n' color=false progress=false >.npmrc
     - rm -rf node_modules
-    - case $CIRCLE_NODE_INDEX in 0) make setup ;; 1) nvm exec 0.12.7 make setup ;; 2) nvm exec 4 make setup ;; 3) nvm install 6 && nvm exec 6 make setup ;; esac:
+    - case $CIRCLE_NODE_INDEX in 0) make setup ;; 1) nvm exec 4 make setup ;; 2) nvm install 6 && nvm exec 6 make setup ;; esac:
         parallel: true
 
 machine:
   node:
-    version: 0.10.34
+    version: 0.12.7
 
 test:
   override:
     - make lint
-    - case $CIRCLE_NODE_INDEX in 0) make test ;; 1) nvm exec 0.12.7 make test ;; 2) nvm exec 4 make test ;; 3) nvm exec 6 make test ;; esac:
+    - case $CIRCLE_NODE_INDEX in 0) make test ;; 1) nvm exec 4 make test ;; 2) nvm exec 6 make test ;; esac:
         parallel: true
+    - node_modules/.bin/karma start
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,0 +1,115 @@
+'use strict';
+
+//  depMain :: String -> String
+var depMain = function(name) {
+  var pkg = require(name + '/package.json');
+  var main = 'main' in pkg ? pkg.main : 'index.js';
+  return './node_modules/' + name + '/' + main;
+};
+
+//  dependencies :: Array String
+var dependencies = Object.keys(require('./package.json').dependencies);
+
+//  https://saucelabs.com/platforms
+var customLaunchers = {
+  sl_chrome: {
+    base: 'SauceLabs',
+    browserName: 'chrome',
+    version: '51'
+  },
+  sl_firefox: {
+    base: 'SauceLabs',
+    browserName: 'firefox',
+    version: '47'
+  },
+  sl_safari_8: {
+    base: 'SauceLabs',
+    browserName: 'safari',
+    platform: 'OS X 10.10',
+    version: '8'
+  },
+  sl_safari_10: {
+    base: 'SauceLabs',
+    browserName: 'safari',
+    platform: 'OS X 10.11',
+    version: '10'
+  },
+  sl_ios_safari_8: {
+    base: 'SauceLabs',
+    browserName: 'iphone',
+    version: '8.1'
+  },
+  sl_ie_9: {
+    base: 'SauceLabs',
+    browserName: 'internet explorer',
+    platform: 'Windows 2008',
+    version: '9'
+  },
+  sl_ie_11: {
+    base: 'SauceLabs',
+    browserName: 'internet explorer',
+    platform: 'Windows 8.1',
+    version: '11'
+  }
+};
+
+var options = {
+
+  browserDisconnectTimeout: 10000,
+  browserDisconnectTolerance: 2,
+  browserNoActivityTimeout: 90000,
+  captureTimeout: 120000,
+
+  client: {
+    mocha: {opts: 'test/mocha.opts'}
+  },
+
+  plugins: [
+    require('karma-browserify'),
+    require('karma-mocha'),
+    require('karma-sauce-launcher')
+  ],
+
+  frameworks: [
+    'browserify',
+    'mocha'
+  ],
+
+  files: dependencies.map(depMain).concat(['index.js', 'test/**/*.js']),
+
+  preprocessors: {
+    'test/**/*.js': ['browserify']
+  },
+
+  reporters: [
+    'dots',
+    'saucelabs'
+  ],
+
+  singleRun: true,
+
+  browserify: {
+    configure: function(bundle) {
+      bundle.on('prebundle', function() {
+        dependencies.forEach(function(name) {
+          bundle.require(depMain(name), {expose: name});
+        });
+      });
+    }
+  },
+
+  sauceLabs: {
+    testName: 'Sanctuary',
+    project: 'Sanctuary',
+    name: 'Sanctuary Test Suite',
+    startTunnel: true,
+    timeout: 600
+  },
+
+  customLaunchers: customLaunchers,
+
+  browsers: Object.keys(customLaunchers)
+
+};
+
+module.exports = function(config) { config.set(options); };
--- a/package.json
+++ b/package.json
@@ -15,10 +15,15 @@
     "sanctuary-def": "0.6.x"
   },
   "devDependencies": {
+    "browserify": "13.1.x",
     "doctest": "0.10.x",
     "eslint": "2.9.x",
     "istanbul": "0.4.x",
     "jsverify": "0.7.x",
+    "karma": "1.3.x",
+    "karma-browserify": "5.1.x",
+    "karma-mocha": "1.3.x",
+    "karma-sauce-launcher": "1.1.x",
     "mocha": "3.x.x",
     "remember-bower": "0.1.x",
     "sanctuary-style": "0.3.x",
--- a/test/mocha.opts
+++ b/test/mocha.opts
@@ -1,0 +1,3 @@
+--recursive
+--timeout 20000
+--ui tdd
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,9 @@ lint:
 	  -- index.js
 	$(ESLINT) \
 	  --env node \
+	  -- karma.conf.js
+	$(ESLINT) \
+	  --env node \
 	  --global suite \
 	  --global test \
 	  --rule 'dot-notation: [error, {allowKeywords: true}]' \
@@ -62,7 +65,7 @@ setup:
 
 .PHONY: test
 test:
-	$(ISTANBUL) cover node_modules/.bin/_mocha -- --recursive --timeout 20000 --ui tdd
+	$(ISTANBUL) cover node_modules/.bin/_mocha
 	$(ISTANBUL) check-coverage --branches 100
 ifeq ($(shell node --version | cut -d . -f 1),v6)
 	$(DOCTEST) -- index.js
--- a/index.js
+++ b/index.js
@@ -496,11 +496,12 @@
   //. false
   //. ```
   function is(typeRep, x) {
-    return type(x) ===
-           (typeRep.prototype != null &&
-            _toString.call(typeRep.prototype['@@type']) === '[object String]' ?
-              typeRep.prototype['@@type'] :
-              typeRep.name);
+    if (_toString.call(typeRep.prototype['@@type']) === '[object String]') {
+      return x != null && x['@@type'] === typeRep.prototype['@@type'];
+    } else {
+      var match = /function (\w*)/.exec(typeRep);
+      return match != null && match[1] === type(x);
+    }
   }
   S.is = def('is', {}, [TypeRep, $.Any, $.Boolean], is);
 


Shard:
So the tests seem to be running ok through karma and saucelabs now, the issues being raised now is have to do with IE failing some unit tests :scream_cat:

davidchambers:
This is tremendously exciting, @Shard!

test/mocha.opts
```diff
@@ -0,0 +1,3 @@
+--recursive
+--timeout 20000
+--ui tdd
```
davidchambers: [`\n`][1], please. :)


[1]: https://github.com/editorconfig/editorconfig/wiki/Newline-at-End-of-File-Support

package.json
```diff
@@ -15,14 +15,20 @@
...
     "jsverify": "0.7.x",
+    "karma": "^1.3.0",
+    "karma-browserify": "5.1.0",
+    "karma-mocha": "1.3.0",
+    "karma-sauce-launcher": "1.1.0",
     "mocha": "3.x.x",
     "remember-bower": "0.1.x",
     "sanctuary-style": "0.3.x",
     "transcribe": "0.5.x",
+    "watchify": "3.7.0",
```
davidchambers: Is this dependency referenced anywhere?

package.json
```diff
@@ -15,14 +15,20 @@
...
   "devDependencies": {
+    "browserify": "13.1.1",
     "doctest": "0.10.x",
     "eslint": "2.9.x",
     "istanbul": "0.4.x",
     "jsverify": "0.7.x",
+    "karma": "^1.3.0",
+    "karma-browserify": "5.1.0",
+    "karma-mocha": "1.3.0",
+    "karma-sauce-launcher": "1.1.0",
```
davidchambers: Let's not specify the patch version. `"1.1.x"` should be fine.

karma.conf.js
```diff
@@ -0,0 +1,114 @@
+process.env.SAUCE_USERNAME = 'shard'
+var im_lazy_dot_tiff = '372040cd-d46b-4a2d-9d8b-cab91ef9efaa'
+process.env.SAUCE_ACCESS_KEY =  im_lazy_dot_tiff //shh, don't tell anyone
+
+module.exports = function(config){
```
davidchambers: It would be nice to have this file adhere to the [Sanctuary style guide][1]. Could you add __karma.conf.js__ to the list of files to be linted?


[1]: https://github.com/sanctuary-js/sanctuary-style

package.json
Shard: I needed it initially before Saucelabs but i can remove it now, it's only needed if you want to use watch mode for karma.

davidchambers: Cool. I prefer not to specify optional dependencies. :)

svozza:
Can we do this in such a way that we can write the tests in ES6?

karma.conf.js
Shard: Can do, also since github preview is showing off my 'secure code', could we get the two variables (SAUCE_USERNAME, SAUCE_ACCESS_KEY) added to CircleCI?

karma.conf.js
```diff
@@ -0,0 +1,114 @@
+process.env.SAUCE_USERNAME = 'shard'
+var im_lazy_dot_tiff = '372040cd-d46b-4a2d-9d8b-cab91ef9efaa'
+process.env.SAUCE_ACCESS_KEY =  im_lazy_dot_tiff //shh, don't tell anyone
```
davidchambers: I have created a Sauce Labs account for this project, and have added `SAUCE_USERNAME` and `SAUCE_ACCESS_KEY` to the CircleCI configuration. You could remove this now and see whether the new account is configured correctly.

Also, you should generate a new access key for your Sauce Labs account. ;)

Shard:
By upgrading to what I assume will be a test matrix of node4+, we will get about 50% of es6 features. If we decide node 6+ we get 97% of the features. http://kangax.github.io/compat-table/es6/

Only other option to get proper es6 would be babel, which can be integrated into mocha without too much effort.

davidchambers:
Let's stick to ES5 in the test suite for now. Let's change as few things as necessary to get the browser tests running. We can discuss the merits of using ES6 in the test suite in a separate thread.

svozza:
Yeah, that was more of a general question, I didn't mean to do it in this PR. When we remove support for Node 0.10 and 0.12 then our baseline version will be Node 4 and I think we should take advantage of that.

davidchambers:
Since we're currently only including Node 0.10 in the test matrix to approximate browser testing, let's make this change as part of the pull request:

```diff
--- a/circle.yml
+++ b/circle.yml
@@ -2,15 +2,15 @@ dependencies:
   override:
     - printf '%s\n' color=false progress=false >.npmrc
     - rm -rf node_modules
-    - case $CIRCLE_NODE_INDEX in 0) make setup ;; 1) nvm exec 0.12.7 make setup ;; 2) nvm exec 4 make setup ;; 3) nvm install 6 && nvm exec 6 make setup ;; esac:
+    - case $CIRCLE_NODE_INDEX in 0) make setup ;; 1) nvm exec 4 make setup ;; 2) nvm install 6 && nvm exec 6 make setup ;; esac:
         parallel: true
 
 machine:
   node:
-    version: 0.10.34
+    version: 0.12.7
 
 test:
   override:
     - make lint
-    - case $CIRCLE_NODE_INDEX in 0) make test ;; 1) nvm exec 0.12.7 make test ;; 2) nvm exec 4 make test ;; 3) nvm exec 6 make test ;; esac:
+    - case $CIRCLE_NODE_INDEX in 0) make test ;; 1) nvm exec 4 make test ;; 2) nvm exec 6 make test ;; esac:
         parallel: true
```

Shard:
Ok I have made all the changes, linting all passing. One thing I didn't get to add successfully was andriod browsers since they do have pretty big market shares these days, although they mostly use webkit/chrome based browsers so they are usually more advanced than IE9 (probably the worst browser in the matrix at the moment).

Let me know @davidchambers if there is anything else you want me to check over.

circle.yml
```diff
@@ -2,15 +2,16 @@ dependencies:
...
-    version: 0.10.34
+    version: 0.12.7
 
 test:
   override:
     - make lint
-    - case $CIRCLE_NODE_INDEX in 0) make test ;; 1) nvm exec 0.12.7 make test ;; 2) nvm exec 4 make test ;; 3) nvm exec 6 make test ;; esac:
+    - case $CIRCLE_NODE_INDEX in 0) make test ;; 1) nvm exec 4 make test ;; 2) nvm exec 6 make test ;; esac:
         parallel: true
+    - SAUCE_USERNAME=$SAUCE_USERNAME SAUCE_ACCESS_KEY=$SAUCE_ACCESS_KEY ./node_modules/.bin/karma start
```
davidchambers: `SAUCE_USERNAME=$SAUCE_USERNAME SAUCE_ACCESS_KEY=$SAUCE_ACCESS_KEY` seems unnecessary to me. It's passing these two environment variables to `karma` as environment variables unchanged. My understanding is that this will have no discernible effect.

I think we could drop the `./` at the beginning of the path too, giving:

    node_modules/.bin/karma start

karma.conf.js
```diff
@@ -0,0 +1,106 @@
+module.exports = function(config) {
+
+  'use strict';
```
davidchambers: Could this be moved to the top of the file?

karma.conf.js
```diff
@@ -0,0 +1,106 @@
...
+
+    browserify: {
+      configure: function(bundle) {
+        bundle.on('prebundle', function() {
+          bundle.require(
+            './node_modules/ramda/dist/ramda.js',
+            {expose: 'ramda'});
+          bundle.require(
+            './node_modules/sanctuary-def/index.js',
+            {expose: 'sanctuary-def'});
```
davidchambers: I'd prefer this formatting:

```javascript
bundle.require('./node_modules/ramda/dist/ramda.js',
               {expose: 'ramda'});
bundle.require('./node_modules/sanctuary-def/index.js',
               {expose: 'sanctuary-def'});
```

Actually, since we're referencing these paths more than once we could define them as "constants" near the top of the file, giving:

```javascript
bundle.require(ramdaMain, {expose: 'ramda'});
bundle.require(sanctuaryDefMain, {expose: 'sanctuary-def'});
```

karma.conf.js
```diff
@@ -0,0 +1,106 @@
...
+        });
+      }
+    },
+
+    sauceLabs: {
+      testName: 'Sanctuary JS',
+      project: 'Sanctuary JS',
+      name: 'Sanctuary JS Test Suite',
```
davidchambers: I would like people to refer to the project as "Sanctuary" rather than as "Sanctuary JS". Let's use `'Sanctuary'` and `'Sanctuary test suite'` as the values above. :)

test/mocha.opts
davidchambers: It looks as though you :zap:'d this. Thanks. :)

package.json
```diff
@@ -15,10 +15,15 @@
     "sanctuary-def": "0.6.x"
   },
   "devDependencies": {
+    "browserify": "13.1.x",
     "doctest": "0.10.x",
     "eslint": "2.9.x",
     "istanbul": "0.4.x",
     "jsverify": "0.7.x",
+    "karma": "^1.3.x",
```
davidchambers: Can we drop the `^`?

karma.conf.js
Shard: The `use strict`? When i move it to the top outside the module.exports i get a lint error `Use the function form of 'use strict'`. Let me know if I'm missing something.

Shard:
I made all the changes mentioned except for the `use strict` one, not sure about that one.

karma.conf.js
davidchambers: Oh, I see the problem. By changing `index.js` to `*.js` in the makefile we're linting __karma.conf.js__ with the same options as __index.js__. Instead, let's add a separate command:

```make
	$(ESLINT) \
	  --env node \
	  -- karma.conf.js
```

Shard: Ah ok! Reading makefiles is definitially not my strengths :P I fixed this up

karma.conf.js
```diff
@@ -0,0 +1,106 @@
...
+      'saucelabs'
+    ],
+
+    singleRun: true,
+
+    browserify: {
+      configure: function(bundle) {
+        bundle.on('prebundle', function() {
+          bundle.require(ramdaMain, {expose: 'ramda'});
+          bundle.require(sanctuaryDefMain, {expose: 'sanctuary-def'});
```
davidchambers: I've realized it's possible to avoid hard-coding these paths. What do you think of this approach?

```javascript
var depMain = function(name) {
  var pkg = require(name + '/package.json');
  var main = 'main' in pkg ? pkg.main : 'index.js';
  return './node_modules/' + name + '/' + main;
};
```

We could then replace the two lines above with the following:

```javascript
for (var name in require('./package.json').dependencies) {
  bundle.require(depMain(name), {expose: name});
}
```

We could use `depMain` to specify `files` as well.

This approach is a bit more complex, but would hopefully save us from having to change this file if we add or remove a dependency or if the layout of one of our dependencies changes.

karma.conf.js
```diff
@@ -0,0 +1,107 @@
+'use strict';
+
+module.exports = function(config) {
+
+  var ramdaMain = './node_modules/ramda/dist/ramda.js';
+  var sanctuaryDefMain = './node_modules/sanctuary-def/index.js';
+
+  // https://saucelabs.com/platforms
+  var customLaunchers = {
```
davidchambers: Since this doesn't reference `config` let's move it to the top level of the file. In fact, since we only reference `config` in one place I'd like to see `module.exports` reduced to a one-liner:

```javascript
module.exports = function(config) { config.set(options); };
```

By defining `options` at the top level we'd be able to replace a 100-line impure function with a one-line impure function. :)

karma.conf.js
davidchambers: :thumbsup:

karma.conf.js
Shard: It's a little more complex but easy enough to reason about, and yea should future proof any changes.

karma.conf.js
```diff
@@ -0,0 +1,107 @@
...
+    browserName: 'safari',
+    platform: 'OS X 10.11',
+    version: '10'
+  },
+  SL_iOS_Safari_8: {
+    base: 'SauceLabs',
+    browserName: 'iphone',
+    version: '8.1'
+  },
+  SL_IE_9: {
```
davidchambers: Are these keys case-sensitive? Having `SL_IE_9` alongside `sl_ie_11` bothers me a little. ;)

karma.conf.js
```diff
@@ -0,0 +1,107 @@
...
+  },
+
+  frameworks: [
+    'browserify',
+    'mocha'
+  ],
+
+  files: Object.keys(require('./package.json').dependencies)
+    .map(depMain)
+    .concat(['index.js', 'test/**/*.js']),
```
davidchambers: We now referencing `require('./package.json').dependencies` in two places. We could define `dependencies` near the top of the file:

```javascript
var dependencies = Object.keys(require('./package.json').dependencies);
```

This would allow the following changes:

```diff
-  files: Object.keys(require('./package.json').dependencies)
-    .map(depMain)
-    .concat(['index.js', 'test/**/*.js']),
+  files: dependencies.map(depMain).concat(['index.js', 'test/**/*.js']),
```

```diff
-        for (var name in require('./package.json').dependencies) {
-          bundle.require(depMain(name), {expose: name});
-        }
+        dependencies.forEach(function(name) {
+          bundle.require(depMain(name), {expose: name});
+        });
```

karma.conf.js
```diff
@@ -0,0 +1,107 @@
+'use strict';
+
+var depMain = function(name) {
```
davidchambers: Let's add a type signature:

```javascript
//  depMain :: String -> String
var depMain = function(name) {
```

davidchambers:
This is looking great, @Shard! I left a few more comments, but I think these will be my last. Feel free to squash the commits when you make these final tweaks. :)

karma.conf.js
Shard: It's not case sensitive, i just missed it when copying snippets around. Ill make them all lower

Shard:
Fixed up those last changes, everything now squashed up :+1:

package.json
```diff
@@ -15,10 +15,15 @@
...
   "devDependencies": {
+    "browserify": "13.1.x",
     "doctest": "0.10.x",
     "eslint": "2.9.x",
     "istanbul": "0.4.x",
     "jsverify": "0.7.x",
+    "karma": "1.3.x",
+    "karma-browserify": "5.1.x",
+    "karma-mocha": "1.3.x",
+    "karma-sauce-launcher": "1.1.x",
```
davidchambers: Is there a straightforward way to determine whether these dependencies are used? My normal approach, `ack 'require.*karma-sauce-launcher'`, is not applicable.

karma.conf.js
```diff
@@ -0,0 +1,109 @@
...
+  var pkg = require(name + '/package.json');
+  var main = 'main' in pkg ? pkg.main : 'index.js';
+  return './node_modules/' + name + '/' + main;
+};
+
+//  dependencies :: Array String
+var dependencies = Object.keys(require('./package.json').dependencies);
+
+// https://saucelabs.com/platforms
```
davidchambers: If you're making other changes would you mind adding a second space after `//`? :)

package.json
Shard: I believe the reason why that wouldn't is how karma handles plugin loading. Since we don't specify any plugins to load, it requires any compatible plugin avaiable. One way we could rectify this is doing something like which i think will work:

```
plugins: [
  require('karma-browserify'),
  require('karma-mocha'),
  require('karma-sauce-launcher'),
]
```

Shard: Good point, funny enough I had issues with yarn when working on this and ended up using npm instead.... I need to look more into how this magic works.

Shard:
Fixed up those last two and resquashed

davidchambers:
:deciduous_tree:

svozza:
This is brilliant! Fantastic work.

PR #305, opened 2016-12-07
test: check conformance to Fantasy Land laws with JSVerify

davidchambers:
This is not quite ready to merge. I'll first check that the Haskell-like expressions of the laws are accurate. I'll add links (in pull request comments) to the relevant Haskell documentation as I do so.

### DIFF
--- a/test/Either/Either.js
+++ b/test/Either/Either.js
@@ -1,46 +1,172 @@
 'use strict';
 
 var jsc = require('jsverify');
+var R = require('ramda');
 
 var S = require('../..');
 
-var Compose = require('../internal/Compose');
 var EitherArb = require('../internal/EitherArb');
 var Identity = require('../internal/Identity');
 var IdentityArb = require('../internal/IdentityArb');
+var laws = require('../internal/laws');
+var squareRoot = require('../internal/squareRoot');
 
 
 suite('Either', function() {
 
+  suite('Setoid laws', function() {
+
+    var setoidLaws = laws.Setoid;
+
+    setoidLaws.reflexivity(
+      EitherArb(jsc.string, jsc.falsy)
+    );
+
+    setoidLaws.symmetry(
+      EitherArb(jsc.bool, jsc.bool),
+      EitherArb(jsc.bool, jsc.bool)
+    );
+
+    setoidLaws.transitivity(
+      EitherArb(jsc.bool, jsc.bool),
+      EitherArb(jsc.bool, jsc.bool),
+      EitherArb(jsc.bool, jsc.bool)
+    );
+
+  });
+
+  suite('Semigroup laws', function() {
+
+    var semigroupLaws = laws.Semigroup(R.equals);
+
+    semigroupLaws.associativity(
+      EitherArb(jsc.string, jsc.string),
+      EitherArb(jsc.string, jsc.string),
+      EitherArb(jsc.string, jsc.string)
+    );
+
+  });
+
+  suite('Functor laws', function() {
+
+    var functorLaws = laws.Functor(R.equals);
+
+    functorLaws.identity(
+      EitherArb(jsc.string, jsc.number)
+    );
+
+    functorLaws.composition(
+      EitherArb(jsc.string, jsc.number),
+      jsc.constant(Math.sqrt),
+      jsc.constant(Math.abs)
+    );
+
+  });
+
+  suite('Apply laws', function() {
+
+    var applyLaws = laws.Apply(R.equals);
+
+    applyLaws.composition(
+      EitherArb(jsc.string, jsc.constant(Math.sqrt)),
+      EitherArb(jsc.string, jsc.constant(Math.abs)),
+      EitherArb(jsc.string, jsc.number)
+    );
+
+  });
+
+  suite('Applicative laws', function() {
+
+    var applicativeLaws = laws.Applicative(R.equals, S.Either);
+
+    applicativeLaws.identity(
+      EitherArb(jsc.string, jsc.number)
+    );
+
+    applicativeLaws.homomorphism(
+      jsc.constant(Math.abs),
+      jsc.number
+    );
+
+    applicativeLaws.interchange(
+      EitherArb(jsc.string, jsc.constant(Math.abs)),
+      jsc.number
+    );
+
+  });
+
+  suite('Chain laws', function() {
+
+    var chainLaws = laws.Chain(R.equals);
+
+    chainLaws.associativity(
+      EitherArb(jsc.string, jsc.array(jsc.number)),
+      jsc.constant(function(xs) { return xs.length > 0 ? S.Right(xs[0]) : S.Left('Empty list'); }),
+      jsc.constant(squareRoot)
+    );
+
+  });
+
+  suite('Monad laws', function() {
+
+    var monadLaws = laws.Monad(R.equals, S.Either);
+
+    monadLaws.leftIdentity(
+      jsc.constant(squareRoot),
+      jsc.number
+    );
+
+    monadLaws.rightIdentity(
+      EitherArb(jsc.string, jsc.number)
+    );
+
+  });
+
+  suite('Foldable laws', function() {
+
+    var foldableLaws = laws.Foldable(R.equals);
+
+    foldableLaws.associativity(
+      jsc.constant(function(x, y) { return x + y; }),
+      jsc.number,
+      EitherArb(jsc.string, jsc.number)
+    );
+
+  });
+
   suite('Traversable laws', function() {
 
-    test('satisfies naturality', function() {
-      var identityToMaybe = S.compose(S.Just, S.prop('value'));
-      jsc.assert(jsc.forall(EitherArb(jsc.integer, IdentityArb(jsc.string)), function(either) {
-        var lhs = identityToMaybe(either.sequence(Identity.of));
-        var rhs = either.map(identityToMaybe).sequence(S.Maybe.of);
-        return lhs.equals(rhs);
-      }));
-    });
+    var traversableLaws = laws.Traversable(R.equals);
 
-    test('satisfies identity', function() {
-      jsc.assert(jsc.forall(EitherArb(jsc.integer, jsc.string), function(either) {
-        var lhs = either.map(Identity).sequence(Identity.of);
-        var rhs = Identity.of(either);
-        return lhs.equals(rhs);
-      }));
-    });
+    traversableLaws.naturality(
+      jsc.constant(S.compose(S.Just, S.prop('value'))),
+      EitherArb(jsc.string, IdentityArb(jsc.number)),
+      jsc.constant(Identity),
+      jsc.constant(S.Maybe)
+    );
 
-    test('satisfies composition', function() {
-      jsc.assert(jsc.forall(EitherArb(jsc.string, IdentityArb(EitherArb(jsc.string, jsc.integer))), function(u) {
-        var C = Compose(Identity)(S.Either);
-        var lhs = u.map(C).sequence(C.of);
-        var rhs = C(u.sequence(Identity.of).map(function(x) {
-          return x.sequence(S.Either.of);
-        }));
-        return lhs.equals(rhs);
-      }));
-    });
+    traversableLaws.identity(
+      EitherArb(jsc.string, jsc.number),
+      jsc.constant(Identity)
+    );
+
+    traversableLaws.composition(
+      EitherArb(jsc.string, IdentityArb(EitherArb(jsc.string, jsc.number))),
+      jsc.constant(Identity),
+      jsc.constant(S.Either)
+    );
+
+  });
+
+  suite('Extend laws', function() {
+
+    var extendLaws = laws.Extend(R.equals);
+
+    extendLaws.associativity(
+      EitherArb(jsc.string, jsc.integer),
+      jsc.constant(function(either) { return either.value + 1; }),
+      jsc.constant(function(either) { return either.value * either.value; })
+    );
 
   });
 
--- a/test/internal/laws/Applicative.js
+++ b/test/internal/laws/Applicative.js
@@ -1,0 +1,36 @@
+'use strict';
+
+var $ = require('./$');
+var ap = require('./ap');
+var forall = require('./forall');
+var id = require('./id');
+var of = require('./of');
+
+
+module.exports = function(equals, A) {
+  var pure = of(A);
+  return forall({
+
+    //  pure id <*> v = v
+    identity: function(v) {
+      var lhs = ap(pure(id))(v);
+      var rhs = v;
+      return equals(lhs, rhs);
+    },
+
+    //  pure f <*> pure x = pure (f x)
+    homomorphism: function(f, x) {
+      var lhs = ap(pure(f))(pure(x));
+      var rhs = pure(f(x));
+      return equals(lhs, rhs);
+    },
+
+    //  u <*> pure y = pure ($ y) <*> u
+    interchange: function(u, y) {
+      var lhs = ap(u)(pure(y));
+      var rhs = ap(pure($(y)))(u);
+      return equals(lhs, rhs);
+    }
+
+  });
+};
--- a/test/internal/laws/Apply.js
+++ b/test/internal/laws/Apply.js
@@ -1,0 +1,20 @@
+'use strict';
+
+var ap = require('./ap');
+var compose = require('./compose');
+var forall = require('./forall');
+var map = require('./map');
+
+
+module.exports = function(equals) {
+  return forall({
+
+    //  (.) <$> u <*> v <*> w = u <*> (v <*> w)
+    composition: function(u, v, w) {
+      var lhs = ap(ap(map(compose)(u))(v))(w);
+      var rhs = ap(u)(ap(v)(w));
+      return equals(lhs, rhs);
+    }
+
+  });
+};
--- a/test/internal/laws/Chain.js
+++ b/test/internal/laws/Chain.js
@@ -1,0 +1,19 @@
+'use strict';
+
+var chain = require('./chain_');
+var composeK = require('./composeK');
+var forall = require('./forall');
+
+
+module.exports = function(equals) {
+  return forall({
+
+    //  (m >>= f) >>= g = m >>= (f >=> g)
+    associativity: function(m, f, g) {
+      var lhs = chain(g)(chain(f)(m));
+      var rhs = chain(composeK(g)(f))(m);
+      return equals(lhs, rhs);
+    }
+
+  });
+};
--- a/test/internal/laws/Extend.js
+++ b/test/internal/laws/Extend.js
@@ -1,0 +1,19 @@
+'use strict';
+
+var compose = require('./compose');
+var extend = require('./extend_');
+var forall = require('./forall');
+
+
+module.exports = function(equals) {
+  return forall({
+
+    //  (extend f . extend g) w = extend (f . extend g) w
+    associativity: function(w, f, g) {
+      var lhs = compose(extend(f))(extend(g))(w);
+      var rhs = extend(compose(f)(extend(g)))(w);
+      return equals(lhs, rhs);
+    }
+
+  });
+};
--- a/test/internal/laws/Functor.js
+++ b/test/internal/laws/Functor.js
@@ -1,0 +1,27 @@
+'use strict';
+
+var compose = require('./compose');
+var forall = require('./forall');
+var id = require('./id');
+var map = require('./map');
+
+
+module.exports = function(equals) {
+  return forall({
+
+    //  id <$> u = u
+    identity: function(u) {
+      var lhs = map(id)(u);
+      var rhs = u;
+      return equals(lhs, rhs);
+    },
+
+    //  (f . g) <$> u = f <$> g <$> u
+    composition: function(u, f, g) {
+      var lhs = map(compose(f)(g))(u);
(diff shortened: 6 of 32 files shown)


test/internal/laws/Monoid.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+var concat = require('./concat');
+var forall = require('./forall');
+
+
+module.exports = forall({
+
+  //  (empty m) `concat` m = m
```
davidchambers: <https://hackage.haskell.org/package/semigroups-0.18/docs/Data-Semigroup.html#t:Monoid>:

>     mappend mempty x = x

test/internal/laws/Monoid.js
```diff
@@ -0,0 +1,23 @@
...
+  leftIdentity: function(m) {
+    var lhs = concat(m.empty())(m);
+    var rhs = m;
+    return [lhs, rhs];
+  },
+
+  //  m `concat` (empty m) = m
```
davidchambers: <https://hackage.haskell.org/package/semigroups-0.18/docs/Data-Semigroup.html#t:Monoid>:

>     mappend x mempty = x

test/internal/laws/forall.js
```diff
@@ -0,0 +1,23 @@
...
+//  forall :: StrMap Function -> StrMap Function
+module.exports = function forall(laws) {
+  return Object.keys(laws).reduce(function(result, name) {
+    result[name] = function() {
+      var args = Array.prototype.slice.call(arguments);
+      function pred() {
+        var pair = laws[name].apply(null, arguments);
+        var lhs = pair[0];
+        var rhs = pair[1];
+        return R.equals(lhs, rhs);
```
safareli: hypothetically if we add some `Lazy` structure (like IO, Future) which contains function, we might want to use custom `equals` function to test their equality (like invoke the function and check if results are same)

davidchambers: That sounds reasonable to me. By the way, I'd love this to become part of a new fantasy-laws project to replace the __laws__ directory in the fantasy-land project. :)

safareli: That would be nice!
I don't understand why are you returning `lhs` and `rhs` as `Tuple` when `equals` function could be passed as argument (which will take as arguments the lhs and rhs)

davidchambers: Are you suggesting this change?

```diff
--- a/test/internal/laws/forall.js
+++ b/test/internal/laws/forall.js
@@ -9,10 +9,8 @@ module.exports = function forall(laws) {
     result[name] = function() {
       var args = Array.prototype.slice.call(arguments);
       function pred() {
-        var pair = laws[name].apply(null, arguments);
-        var lhs = pair[0];
-        var rhs = pair[1];
-        return R.equals(lhs, rhs);
+        var args = [R.equals].concat(Array.prototype.slice.call(arguments));
+        return laws[name].apply(null, args);
       }
       test(name, function() {
         jsc.assert(jsc.forall.apply(jsc, args.concat([pred])));
```

```diff
--- a/test/internal/laws/Monoid.js
+++ b/test/internal/laws/Monoid.js
@@ -7,17 +7,17 @@ var forall = require('./forall');
 module.exports = forall({
 
   //  (empty m) `concat` m = m
-  leftIdentity: function(m) {
+  leftIdentity: function(equals, m) {
     var lhs = concat(m.empty())(m);
     var rhs = m;
-    return [lhs, rhs];
+    return equals(lhs, rhs);
   },
 
   //  m `concat` (empty m) = m
-  rightIdentity: function(m) {
+  rightIdentity: function(equals, m) {
     var lhs = concat(m)(m.empty());
     var rhs = m;
-    return [lhs, rhs];
+    return equals(lhs, rhs);
   }
 
 });
```

safareli: almost yes, but if this should be moved to fl-laws then I think the R.equals should not be hardcoded here, also some functions don't need the `eq` (Setoid), I think the "laws.js" should have api something like [this](https://github.com/fantasyland/fantasy-land/pull/195#issuecomment-256894273):
```js
{
  Setoid: () => ({ ... }),
  Semigroup: (eq) => ({ ... }),
  Applicative: (T, eq) => ({ ... }),
}
```

davidchambers: Something more like this?

```diff
--- a/test/internal/laws/forall.js
+++ b/test/internal/laws/forall.js
@@ -6,16 +6,13 @@ var R = require('ramda');
 //  forall :: StrMap Function -> StrMap Function
 module.exports = function forall(laws) {
   return Object.keys(laws).reduce(function(result, name) {
-    result[name] = function() {
-      var args = Array.prototype.slice.call(arguments);
+    result[name] = function(equals, arbs) {
       function pred() {
-        var pair = laws[name].apply(null, arguments);
-        var lhs = pair[0];
-        var rhs = pair[1];
-        return R.equals(lhs, rhs);
+        var args = [equals].concat(Array.prototype.slice.call(arguments));
+        return laws[name].apply(null, args);
       }
       test(name, function() {
-        jsc.assert(jsc.forall.apply(jsc, args.concat([pred])));
+        jsc.assert(jsc.forall.apply(jsc, arbs.concat([pred])));
       });
     };
     return result;
```

```diff
--- a/test/internal/laws/Monoid.js
+++ b/test/internal/laws/Monoid.js
@@ -7,17 +7,17 @@ var forall = require('./forall');
 module.exports = forall({
 
   //  (empty m) `concat` m = m
-  leftIdentity: function(m) {
+  leftIdentity: function(equals, m) {
     var lhs = concat(m.empty())(m);
     var rhs = m;
-    return [lhs, rhs];
+    return equals(lhs, rhs);
   },
 
   //  m `concat` (empty m) = m
-  rightIdentity: function(m) {
+  rightIdentity: function(equals, m) {
     var lhs = concat(m)(m.empty());
     var rhs = m;
-    return [lhs, rhs];
+    return equals(lhs, rhs);
   }
 
 });
```

```diff
--- a/test/Maybe/Maybe.js
+++ b/test/Maybe/Maybe.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var jsc = require('jsverify');
+var R = require('ramda');
 
 var S = require('../..');
 
@@ -44,13 +45,13 @@ suite('Maybe', function() {
 
   suite('Monoid laws', function() {
 
-    laws.Monoid.leftIdentity(
+    laws.Monoid.leftIdentity(R.equals, [
       MaybeArb(jsc.string)
-    );
+    ]);
 
-    laws.Monoid.rightIdentity(
+    laws.Monoid.rightIdentity(R.equals, [
       MaybeArb(jsc.string)
-    );
+    ]);
 
   });
 
```

safareli:
```js
// forall.js
module.exports = function forall(laws) {
  return Object.keys(laws).reduce(function(result, name) {
   result[name] = function(...arbs) {
      test(name, function() {
       jsc.assert(jsc.forall(...arbs.concat([laws[name]])));
      });
    };
    return result;
  }  
}

// Monoid.js
module.exports = (equals) => forall({
  //  (empty m) `concat` m = m
  leftIdentity: function(m) {
    var lhs = concat(m.empty())(m);
    var rhs = m;
    return equals(lhs, rhs);
  },

  //  m `concat` (empty m) = m
  rightIdentity: function(m) {
    var lhs = concat(m)(m.empty());
    var rhs = m;
    return equals(lhs, rhs);
  }
});

//Maybe.js
var MonadLaws = laws.Monoid(R.equals, Maybe)
var monoidLaws = laws.Monoid(R.equals)
monoidLaws.leftIdentity(MaybeArb(jsc.string));  
monoidLaws.rightIdentity( MaybeArb(jsc.string) );

```

davidchambers: I like it! I'll make those changes when I get some time (probably later this week or this weekend).

davidchambers: I've pushed a commit with these changes. :)

test/internal/laws/Semigroup.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+var concat = require('./concat');
+var forall = require('./forall');
+
+
+module.exports = function(equals) {
+  return forall({
+
+    //  (x `concat` y) `concat` z = x `concat` (y `concat` z)
```
davidchambers: <https://hackage.haskell.org/package/semigroups-0.18/docs/Data-Semigroup.html#t:Monoid>:

>     mappend x (mappend y z) = mappend (mappend x y) z

test/internal/laws/Monoid.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+var concat = require('./concat');
+var forall = require('./forall');
+
+
+module.exports = function(equals) {
+  return forall({
+
+    //  (empty m) `concat` m = m
```
davidchambers: <https://hackage.haskell.org/package/semigroups-0.18/docs/Data-Semigroup.html#t:Monoid>:

>     mappend mempty x = x

test/internal/laws/Monoid.js
```diff
@@ -0,0 +1,25 @@
...
+    leftIdentity: function(m) {
+      var lhs = concat(m.empty())(m);
+      var rhs = m;
+      return equals(lhs, rhs);
+    },
+
+    //  m `concat` (empty m) = m
```
davidchambers: <https://hackage.haskell.org/package/semigroups-0.18/docs/Data-Semigroup.html#t:Monoid>:

>     mappend x mempty = x

test/internal/laws/Functor.js
```diff
@@ -0,0 +1,27 @@
...
+var compose = require('./compose');
+var forall = require('./forall');
+var id = require('./id');
+var map = require('./map');
+
+
+module.exports = function(equals) {
+  return forall({
+
+    //  id <$> u = u
```
davidchambers: <http://hackage.haskell.org/package/base-4.9.0.0/docs/Prelude.html#t:Functor>:

>     fmap id  ==  id

test/internal/laws/Functor.js
```diff
@@ -0,0 +1,27 @@
...
+    identity: function(u) {
+      var lhs = map(id)(u);
+      var rhs = u;
+      return equals(lhs, rhs);
+    },
+
+    //  (f . g) <$> u = f <$> g <$> u
```
davidchambers: <http://hackage.haskell.org/package/base-4.9.0.0/docs/Prelude.html#t:Functor>:

>     fmap (f . g)  ==  fmap f . fmap g

test/internal/laws/Apply.js
```diff
@@ -0,0 +1,20 @@
...
+var ap = require('./ap');
+var compose = require('./compose');
+var forall = require('./forall');
+var map = require('./map');
+
+
+module.exports = function(equals) {
+  return forall({
+
+    //  (.) <$> u <*> v <*> w = u <*> (v <*> w)
```
davidchambers: <https://hackage.haskell.org/package/base-4.9.0.0/docs/Control-Applicative.html#t:Applicative>:

> __composition__
>
>     pure (.) <*> u <*> v <*> w = u <*> (v <*> w)

test/internal/laws/Applicative.js
```diff
@@ -0,0 +1,36 @@
...
+var forall = require('./forall');
+var id = require('./id');
+var of = require('./of');
+
+
+module.exports = function(equals, A) {
+  var pure = of(A);
+  return forall({
+
+    //  pure id <*> v = v
```
davidchambers: <https://hackage.haskell.org/package/base-4.9.0.0/docs/Control-Applicative.html#t:Applicative>:

> __identity__
>
>     pure id <*> v = v

test/internal/laws/Applicative.js
```diff
@@ -0,0 +1,36 @@
...
+    identity: function(v) {
+      var lhs = ap(pure(id))(v);
+      var rhs = v;
+      return equals(lhs, rhs);
+    },
+
+    //  pure f <*> pure x = pure (f x)
```
davidchambers: <https://hackage.haskell.org/package/base-4.9.0.0/docs/Control-Applicative.html#t:Applicative>:

> __homomorphism__
>
>     pure f <*> pure x = pure (f x)

test/internal/laws/Applicative.js
```diff
@@ -0,0 +1,36 @@
...
+    homomorphism: function(f, x) {
+      var lhs = ap(pure(f))(pure(x));
+      var rhs = pure(f(x));
+      return equals(lhs, rhs);
+    },
+
+    //  u <*> pure y = pure ($ y) <*> u
```
davidchambers: <https://hackage.haskell.org/package/base-4.9.0.0/docs/Control-Applicative.html#t:Applicative>:

> __interchange__
>
>     u <*> pure y = pure ($ y) <*> u

test/internal/laws/Chain.js
```diff
@@ -0,0 +1,19 @@
...
+
+var compose = require('./compose');
+var flatMap = require('./flatMap');
+var forall = require('./forall');
+
+
+module.exports = function(equals) {
+  return forall({
+
+    //  m >>= f >>= g = m >>= (f >=> g)
```
davidchambers: <https://hackage.haskell.org/package/base-4.9.0.0/docs/Control-Monad.html#t:Monad>:

>     m >>= (x -> k x >>= h)  =  (m >>= k) >>= h

test/internal/laws/Monad.js
```diff
@@ -0,0 +1,27 @@
...
+var flatMap = require('./flatMap');
+var forall = require('./forall');
+var of = require('./of');
+
+
+module.exports = function(equals, M) {
+  var pure = of(M);
+  return forall({
+
+    //  pure x >>= f = f x
```
davidchambers: <https://hackage.haskell.org/package/base-4.9.0.0/docs/Control-Monad.html#t:Monad>:

>     return a >>= k  =  k a

test/internal/laws/Monad.js
```diff
@@ -0,0 +1,27 @@
...
+    leftIdentity: function(f, x) {
+      var lhs = flatMap(f)(pure(x));
+      var rhs = f(x);
+      return equals(lhs, rhs);
+    },
+
+    //  m >>= pure = m
```
davidchambers: <https://hackage.haskell.org/package/base-4.9.0.0/docs/Control-Monad.html#t:Monad>:

>     m >>= return  =  m

test/internal/laws/Extend.js
```diff
@@ -0,0 +1,19 @@
...
+
+var compose = require('./compose');
+var extend = require('./extend_');
+var forall = require('./forall');
+
+
+module.exports = function(equals) {
+  return forall({
+
+    //  (extend f . extend g) w = extend (f . extend g) w
```
davidchambers: <https://hackage.haskell.org/package/comonad-5/docs/Control-Comonad.html#t:Comonad>:

>     extend f . extend g = extend (f . extend g)

davidchambers:
This is ready to review. Please have a look. :)

test/internal/laws/Setoid.js
```diff
@@ -0,0 +1,23 @@
...
+  },
+
+  //  a `equals` b = b `equals` a
+  symmetry: function(a, b) {
+    return a.equals(b) === b.equals(a);
+  },
+
+  //  a `equals` b & b `equals` c => a `equals` c
+  transitivity: function(a, b, c) {
+    return !a.equals(b) || !b.equals(c) || a.equals(c);
```
safareli: what you think on `a.equals(b) && b.equals(c) === a.equals(c)`, I think this easier to understand.

test/internal/laws/chain_.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+var R = require('ramda');
+
+//  flatMap :: Chain m => (a -> m b) -> m a -> m b
+module.exports = function flatMap(f) {
```
davidchambers: Good catch. To avoid a naming conflict on case-insensitive file systems I originally named this file __flatMap.js__ (__chain.js__ is not an option due to the presence of __Chain.js__). I later decided to go with __chain\_.js__ instead, but forgot to update these.

:zap:

test/Either/Either.js
```diff
@@ -1,46 +1,172 @@
...
+
+    var functorLaws = laws.Functor(R.equals);
+
+    functorLaws.identity(
+      EitherArb(jsc.string, jsc.number)
+    );
+
+    functorLaws.composition(
+      EitherArb(jsc.string, jsc.number),
+      jsc.constant(Math.sqrt),
```
davidchambers: To avoid lots of `NaN` comparisons do you mean?

test/internal/laws/Setoid.js
davidchambers: `a` and `c` are allowed to be equal even if `b` is different:

```javascript
> const a = 1
> const b = 2
> const c = 1
> a === b && b === c
false
> a === c
true
```

I agree that the current expression takes a few moments to understand, but I like the alternative even less:

```javascript
if (a.equals(b) && b.equals(c)) {
  return a.equals(c);
} else {
  return true;
}
```

test/internal/laws/Chain.js
davidchambers: @scott-christopher suggested a different approach on [Gitter][1]:

> Yep, that's the associativity law.
>
> It looks more similar to other associative laws if you use [Kleisli] arrows (`>=>`) instead of `>>=`
>
> ```haskell
> (f >=> g) >=> h ≡ f >=> (g >=> h)
> ```

I like this! There's no clearer expression of associativity than ``(a `op` b) `op` c = a `op` (b `op` c)``.


[1]: https://gitter.im/sanctuary-js/sanctuary?at=5863a0afaf6b364a29049f16

safareli: I like that too

test/Either/Either.js
safareli: I saw `squareRoot` used in couple places and when I saw `Math.sqrt`, though that it was a mistake.

test/internal/laws/Setoid.js
safareli: what about this? `a.equals(b) && b.equals(c) ? a.equals(c) : true`

test/Either/Either.js
davidchambers: `squareRoot` is used where we need `Chain m => a -> m b` or `Monad m => a -> m b`. It might be clearer to only use it in such cases, and use "simpler" functions where any `a -> b` will do.

test/internal/laws/Setoid.js
davidchambers: Sure. Let me know which of these you like best and we'll go with it:

  - `!a.equals(b) || !b.equals(c) || a.equals(c)`
  - `!(a.equals(b) && b.equals(c)) || a.equals(c)`
  - `a.equals(b) && b.equals(c) ? a.equals(c) : true`

safareli: I think it takes fewer moments to understand version using ternary, then with just boolean logic. So I would wrote for `a.equals(b) && b.equals(c) ? a.equals(c) : true`. It's also direct translation of the `if/else` expression you wrote above, which is more descriptive and easier to understand. it reads as: `if a equals b and b equals c then a must equal c` which is not the case for other variations

test/internal/laws/Chain.js
```diff
@@ -0,0 +1,19 @@
...
+    associativity: function(m, f, g) {
+      var lhs = chain(g)(chain(f)(m));
+      var rhs = compose(chain(g))(chain(f))(m);
```
safareli: I think this law hasn't been updated.

if we use kleisli compose than we need 3 functions instead of 2 (f, g)
```js
var lhs = composeKleisli(f, composeKleisli(g, h))(m);
var rhs = composeKleisli(composeKleisli(f, g), h))(m);
```

test/internal/laws/Setoid.js
davidchambers: :zap:

test/internal/laws/Functor.js
```diff
@@ -0,0 +1,27 @@
...
+    composition: function(u, f, g) {
+      var lhs = map(compose(f)(g))(u);
+      var rhs = map(f)(map(g)(u));
```
davidchambers: :zap:

test/internal/laws/Chain.js
davidchambers: I've updated `rhs` as you first suggested. I'm reluctant to switch to the three-function version (even though it reads very nicely) as doing so would increase the burden on those using this function to test their data types. Also, the two-function version more closely resembles the expressions in the FL spec.

Do you agree?

test/internal/laws/Chain.js
davidchambers: As I mentioned in my previous comment (in another thread) I'd like to stick to a two-function version even though the three-function `>=>` version is so elegant. :fish:

> I think we should add parentheses to (m >>= f)

Parens appear not to be required:

```haskell
λ Just 10 >>= Just . (* 2) >>= Just . (+ 1)
Just 21
```

Are you suggesting we add parens for clarity?

test/internal/laws/Chain.js
safareli: 👍

test/internal/laws/Chain.js
safareli: yes, in [hackage](https://hackage.haskell.org/package/base-4.9.0.0/docs/Control-Monad.html#t:Monad) it's also written with parens

davidchambers: :zap:

safareli:
🎅

PR #311, opened 2016-12-23
Karma Improvements

### LINKED ISSUE #309: Browser tests not properly loading modules
I noticed some errors when looking at #232 and they seemed to be caused by `sanctuary-def` loading `sanctuary-type-classes` as `undefined` in the browser tests.

I believe it has something to do with browserify configuration, still looking into it.

Shard:
This PR fixes #309 by implementing some extra code to determine the correct loading order for browserify.

I also decided to break up the config into a base/remote configs so that devs can actually use `karma start` without having to go and get an API key for saucelabs to run a browser test. As long as `CI=true` is not in your local dev environment for some reason, `karma start` will now give you a local port to point your browser at to run the tests like that.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -7,9 +7,19 @@ function depMain(name) {
   return './node_modules/' + name + '/' + main;
 }
 
+//  getDeps :: String -> StrMap String String
+function getDeps(path) {
+  return Object.keys(require(path)).dependencies;
+}
+
 //  dependencies :: Array String
-var dependencies = Object.keys(require('./package.json').dependencies)
-                         .reverse();
+var dependencies = getDeps('./package.json')
+      .map(function(x) {
+        return [x, getDeps('./node_modules/' + x + '}/package.json')];
+      })
+      .sort(function(a, b) { return a[1].indexOf(b[0]) === -1 ? -1 : 1; })
+      .sort(function(b, a) { return a[1].indexOf(b[0]) === -1 ? 1 : -1; })
+      .map(function(x) { return x[0]; });
 
 //  https://saucelabs.com/platforms
 var customLaunchers = {
@@ -78,6 +88,7 @@ var baseOptions = {
   ],
 
   plugins: [
+    require('karma-sauce-launcher'),
     require('karma-browserify'),
     require('karma-mocha')
   ],
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -8,7 +8,8 @@ function depMain(name) {
 }
 
 //  dependencies :: Array String
-var dependencies = Object.keys(require('./package.json').dependencies);
+var dependencies = Object.keys(require('./package.json').dependencies)
+                         .reverse();
 
 //  https://saucelabs.com/platforms
 var customLaunchers = {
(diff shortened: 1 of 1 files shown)


karma.conf.js
```diff
@@ -7,8 +7,19 @@ function depMain(name) {
...
+}
+
 //  dependencies :: Array String
-var dependencies = Object.keys(require('./package.json').dependencies);
+var dependencies = getDeps('./package.json')
+      .map(function(x) {
+        return [x, getDeps('./node_modules/' + x + '/package.json')];
+      })
+      .sort(function(a, b) { return a[1].indexOf(b[0]) === -1 ? -1 : 1; })
+      .sort(function(b, a) { return a[1].indexOf(b[0]) === -1 ? 1 : -1; })
```
davidchambers: Could you explain why it's necessary to sort twice?

By the way, I prefer `xs.indexOf(x) < 0` to `xs.indexOf(x) === -1`. `-1` is the "bogus" value; it's nice to avoid referencing it directly.

karma.conf.js
```diff
@@ -7,8 +7,19 @@ function depMain(name) {
...
+function getDeps(path) {
+  return Object.keys(require(path).dependencies);
+}
+
 //  dependencies :: Array String
-var dependencies = Object.keys(require('./package.json').dependencies);
+var dependencies = getDeps('./package.json')
+      .map(function(x) {
+        return [x, getDeps('./node_modules/' + x + '/package.json')];
+      })
```
davidchambers: Doesn't this need to be recursive? What about the dependencies of our dependencies' dependencies? :stuck_out_tongue_winking_eye:

karma.conf.js
```diff
@@ -53,63 +64,77 @@ var customLaunchers = {
...
+  browserNoActivityTimeout: 90000,
+  captureTimeout: 120000,
+
+  customLaunchers: customLaunchers,
+
+  plugins: [
+    require('karma-browserify'),
+    require('karma-mocha'),
+    require('karma-sauce-launcher')
+  ],
```
davidchambers: This is defined in `baseOptions`. Is there a reason to define it here as well?

karma.conf.js
```diff
@@ -53,63 +64,77 @@ var customLaunchers = {
...
-  customLaunchers: customLaunchers,
+  }
 
-  browsers: Object.keys(customLaunchers)
+}, baseOptions);
 
+module.exports = function(config) {
+  process.env.CI === 'true' ?
+    config.set(remoteOptions) :
+    config.set(baseOptions);
```
davidchambers: Slightly cleaner:

```javascript
config.set(process.env.CI === 'true' ? remoteOptions : baseOptions);
```

Also, we could add [envvar][1] as a dev dependency:

```javascript
var CI = envvar.boolean('CI', false);
```


[1]: https://github.com/plaid/envvar

karma.conf.js
Shard: There was, but not anymore. Removed.

karma.conf.js
Shard: I should probably understand sort better but when I was writing it with a larger list of deps it would only "mostly" sort it but have a couple of items slightly out of order. When I ran the same pass in reverse it seemed to complete the sorts without issue.

karma.conf.js
Shard: Oops, i guess i need to make it recursive.

davidchambers:
Now that #309 has been resolved do you still wish to continue with this pull request, @Shard? If so, you'll need to rebase this branch.

PR #312, opened 2016-12-24
move @@type properties to type representatives

davidchambers:
This pull request depends on sanctuary-js/sanctuary-type-identifiers#1. Furthermore, this pull request must not be merged until sanctuary-js/sanctuary-def#112 has been merged, `sanctuary-def@0.9.0` has been published, and Sanctuary has been updated to use that version. Here's the current problem:

```javascript
var $ = require('sanctuary-def');

function Identity(x) {
  if (!(this instanceof Identity)) return new Identity(x);
  this.value = x;
}

Identity['@@type'] = 'my-package/Identity';

$.Function._test(Math.abs);  // => true
$.Function._test(Identity);  // => false
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -198,21 +198,27 @@
 
   'use strict';
 
+  //  deps :: Array String
+  var deps = [
+    'sanctuary-def',
+    'sanctuary-type-classes',
+    'sanctuary-type-identifiers'
+  ];
+
   /* istanbul ignore else */
   if (typeof module === 'object' && typeof module.exports === 'object') {
-    module.exports = f(require('sanctuary-type-classes'),
-                       require('sanctuary-def'));
+    module.exports = f(require(deps[0]), require(deps[1]), require(deps[2]));
   } else if (typeof define === 'function' && define.amd != null) {
-    define(['sanctuary-type-classes', 'sanctuary-def'], f);
+    define(deps, f);
   } else {
-    self.sanctuary = f(self.sanctuaryTypeClasses, self.sanctuaryDef);
+    self.sanctuary = f(self.sanctuaryDef,
+                       self.sanctuaryTypeClasses,
+                       self.sanctuaryTypeIdentifiers);
   }
 
-}(function(Z, $) {
+}(function($, Z, type) {
 
   'use strict';
-
-  var _toString = Object.prototype.toString;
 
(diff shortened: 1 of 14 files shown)


safareli:
> ```js
> $.Function._test(Math.abs);  // => true
> $.Function._test(Identity);  // => false
> ```
Interesting problem, but if i'm not wrong, it's not quite a problem, consider this example:
```js
// data Identity a = Identity a
const IdentityRep = {
  Identity:  (value) => ({ value, constructor: IdentityRep, tag:'Identity' }),
  '@@type': 'my/Identity',
};
```
Even tho `Identity`, is on left and right side of the haskell version, they are different things, first one used in type signatures and second one is used to construct values of type `:: Identity a`, it's more clear with Either:

```js
// data Either a b = Left a | Right b
const Either = {
  Left:  (value) => ({ value, constructor: Either, tag:'Left' }),
  Right: (value) => ({ value, constructor: Either, tag:'Right' }),
  '@@type': 'my/Either',
};
```

davidchambers:
We want to be able to do things like this:

```javascript
> S.map(Identity, [1, 2, 3])
[Identity(1), Identity(2), Identity(3)]
```

It's thus important that `Identity` is a member of `a -> b`.

index.js
```diff
@@ -897,7 +905,12 @@
   }
   S.Just = def('Just', {}, [a, $Maybe(a)], Just);
 
-  //# Maybe.fantasy-land/empty :: () -> Maybe a
+  //# Maybe.@@type :: String
+  //.
+  //. Maybe type identifier, `'sanctuary/Maybe'`.
+  Maybe['@@type'] = maybeTypeIdent;
+
+  //# Maybe.empty :: () -> Maybe a
```
davidchambers: #322

index.js
```diff
@@ -1515,7 +1523,12 @@
   }
   S.Right = def('Right', {}, [b, $Either(a, b)], Right);
 
-  //# Either.fantasy-land/of :: b -> Either a b
+  //# Either.@@type :: String
+  //.
+  //. Either type identifier, `'sanctuary/Either'`.
+  Either['@@type'] = eitherTypeIdent;
+
+  //# Either.of :: b -> Either a b
```
davidchambers: #322

PR #313, opened 2016-12-25
maybe, either: add Setoid to list of type classes satisfied

davidchambers:
I've realized the Fantasy Land URLs should be anchored to a specific tag. It's misleading to link to whatever is on `master`, particularly since we're currently targeting such an old version of the spec.


davidchambers:
I created fantasyland/fantasy-land#216 so we'll soon have a tag to reference.

davidchambers:
I'll merge #314 then rebase this pull request. :)

PR #316, opened 2016-12-29
maybe: add S.fromMaybe_

### LINKED ISSUE #296: fromMaybe_ :: (() -> a) -> Maybe a -> a
We added `maybe_`, a "lazy" version of [`maybe`][1], in #285. `maybe` is a convenient function for situations in which one wants to `map` then apply [`fromMaybe`][2]. It seems to me we should also provide a "lazy" version of `fromMaybe`. Otherwise we'll end up writing `S.maybe_(thunk, S.I, m)` in some cases, which seems a bit strange.

Do you agree?

[1]: https://sanctuary.js.org/#maybe
[2]: https://sanctuary.js.org/#fromMaybe

davidchambers:
Closes #296

### DIFF
--- a/index.js
+++ b/index.js
@@ -1296,7 +1296,8 @@
   //. Takes a default value and a Maybe, and returns the Maybe's value
   //. if the Maybe is a Just; the default value otherwise.
   //.
-  //. See also [`maybeToNullable`](#maybeToNullable).
+  //. See also [`fromMaybe_`](#fromMaybe_) and
+  //. [`maybeToNullable`](#maybeToNullable).
   //.
   //. ```javascript
   //. > S.fromMaybe(0, S.Just(42))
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -1310,6 +1311,28 @@
...
+  //# fromMaybe_ :: (() -> a) -> Maybe a -> a
+  //.
+  //. Variant of [`fromMaybe`](#fromMaybe) which takes a thunk so the default
+  //. value is only computed if required.
+  //.
+  //. ```javascript
+  //. > global.fib = function fib(n) {
+  //. .   return n <= 1 ? n : fib(n - 2) + fib(n - 1);
+  //. . }
+  //. fib
```
Avaq: I had to read this three times before understanding that you're *just* using it as an example of a computationally expensive function. The fact that you explicitly state the return value of the function definition, and that you're explicitly attaching it to the global scope, makes no sense without knowledge of `doctest`. It seemed to me at first as if you're trying to demonstrate something here, whereas in fact you're really just working around the semantics of your linter.

I think it might be better to use something native to JavaScript (or in fact exposed by Sanctuary) to demonstrate. Maybe even something like `/*perform expensive calculation*/ 1 + 1`. Or if you really feel `fib` demonstrates it best, it would've helped me to see a comment above fibs definition explaining me why it's being defined there.

Avaq: Ah. I see you've already taken the same approach for documenting `maybe_`. I suppose multiple instances justify the learning curve.

Avaq: Perhaps if you wouldn't have to assert the return value of the function definition, it would read more easily, and it wouldn't look like a "stand-alone" example.

davidchambers: Thanks for highlighting this, Aldwin. We are indeed working around a limitation of the tool. I just opened davidchambers/doctest#73. At some point I will address that issue, publish a new version of doctest, and update these `fib` definitions. This needn't all happen before this pull request is merged, though. :)

index.js
```diff
@@ -1373,19 +1393,16 @@
   //. is only computed if required.
   //.
   //. ```javascript
-  //. > global.fib = function fib(n) {
-  //. .   return n <= 1 ? n : fib(n - 2) + fib(n - 1);
-  //. . }
-  //. fib
+  //. > function fib(n) { return n <= 1 ? n : fib(n - 2) + fib(n - 1); }
```
davidchambers: Here you go, Aldwin. :)

Avaq: \o/

davidchambers:
```

$ node
undefined
> S.fromMaybe_
fromMaybe_ :: (() -> a) -> Maybe a -> a
```

PR #318, opened 2017-01-03
remove support for node 0.12

svozza:
Now that Node 0.12 is officially end of life we should remove it from our test matrix. I have replaced it with Node 7 but can remove it if we want to speed up build times.


davidchambers:
:deciduous_tree:

Please do the same for other Sanctuary projects, if you feel so inclined. :)

svozza:
Yep, will do!

PR #326, opened 2017-01-21
test: use curried Z functions more broadly

davidchambers:
Motivating example:

```diff
     ComposeFG.prototype[FL.ap] = function(other) {
-      return ComposeFG(Z.ap(Z.map(function(u) { return function(y) { return Z.ap(u, y); }; }, other.value), this.value));
+      return ComposeFG(ap(map(ap)(other.value))(this.value));
     };
```


test/internal/laws/Applicative.js
```diff
@@ -15,21 +15,21 @@ module.exports = function(equals, A) {
     identity: function(v) {
       var lhs = ap(pure(id))(v);
       var rhs = v;
-      return equals(lhs, rhs);
+      return equals(lhs)(rhs);
```
safareli: So with this changes API of "laws", from taking `equals` in uncurried form to curried form. rest of API is in uncurried form which could be confusing (in case we move it to FL, but i think it's ok for now).

davidchambers: Good point. Let's cross that bridge when we get to it.

PR #330, opened 2017-02-04
Remove extraneous combinators

Avaq:
#### Implications for #216

* [ ] Add a note about `map` being `B` to `S.map` documentation.
* [ ] Remove `S.S` in favor of `S.ap` and add a note to its documentation.
* [ ] Add a note about `join` being `W` to `S.join` documentation.

#### Implications for the future

Maybe we should rename the remaining combinators to their full names.


index.js
```diff
@@ -727,21 +689,29 @@
...
-  //# flip :: ((a, b) -> c) -> b -> a -> c
+  //# flip :: (a -> b -> c) -> b -> a -> c
   //.
-  //. Takes a binary function and two values, and returns the result of
-  //. applying the function to the values in reverse order.
+  //. Takes a curried binary function and two values, and returns the
+  //. result of applying the function to the values in reverse order.
   //.
-  //. See also [`C`](#C).
+  //. This is the `C` combinator from combinatory logic
```
davidchambers: Three minor points:

  - In this context C should not be rendered as code. The description of `I`, for example, simply states:

    > The I combinator.

  - The sentence should end with a full stop (`.`).

  - We don't mention combinatory logic in the descriptions of the other combinators, but it seems reasonable to do so here given that `flip` is not listed in the __Combinator__ section.

I suggest:

_This is the C combinator from combinatory logic._

index.js
```diff
@@ -727,21 +689,29 @@
...
   //.
   //. ```javascript
-  //. > Z.map(S.flip(Math.pow)(2), [1, 2, 3, 4, 5])
-  //. [1, 4, 9, 16, 25]
+  //. > S.flip(S.concat, 'foo', 'bar')
+  //. 'barfoo'
```
davidchambers: :thumbsup:

index.js
```diff
@@ -806,13 +776,16 @@
   //. In general terms, `compose` performs right-to-left composition of two
   //. unary functions.
   //.
-  //. See also [`B`](#B) and [`pipe`](#pipe).
+  //. This is the `B` combinator from combinatory logic.
```
davidchambers: See comments about the C combinator above.

index.js
```diff
@@ -727,21 +689,29 @@
       [$.Function([a, b, c, d, e, r]), a, b, c, d, e, r],
       curry5);
 
-  //# flip :: ((a, b) -> c) -> b -> a -> c
+  //# flip :: (a -> b -> c) -> b -> a -> c
   //.
-  //. Takes a binary function and two values, and returns the result of
-  //. applying the function to the values in reverse order.
+  //. Takes a curried binary function and two values, and returns the
+  //. result of applying the function to the values in reverse order.
```
davidchambers: Let's append this sentence to the paragraph:

_Equivalent to Haskell's `flip` function._

We do this for several other functions, and it seems helpful for those coming from Haskell.

Avaq: We've only done that for functions where the name in Sanctuary differs from the name in Haskell. Are you sure we want to do it for `flip`, which maps to .. `flip`?

index.js
```diff
@@ -806,13 +776,16 @@
   //. In general terms, `compose` performs right-to-left composition of two
   //. unary functions.
```
davidchambers: We could simplify this description. The first paragraph is overly wordy. What do you think of the following?

_Composes two unary functions, from right to left. Equivalent to Haskell's `(.)` function._

That says it all, really.

Avaq: It's not really related to this PR, but I'm happy to make that change. ;)

index.js
davidchambers: Haha, fair point. It may not add value. The type signature already indicates that the function takes three arguments (unlike Ramda's version which takes just one).

I leave it to you to decide whether or not to include the sentence. I'm happy either way. :)

index.js
davidchambers: It's not really related, that's true, but an issue entitled "improve all function descriptions" is too daunting to consider. We can at least make small improvements as they occur to us. :)

index.js
Avaq: Let's leave it this way. Once we go down this road we may feel the urge to document how *every* function relates to Haskell. I'll save us that trouble. ;)

index.js
```diff
@@ -799,20 +769,19 @@
...
-  //. and returns the result of applying the first function to the result
-  //. of applying the second function to the given value.
-  //.
-  //. In general terms, `compose` performs right-to-left composition of two
-  //. unary functions.
+  //. Composes two unary functions, from right to left. Equivalent to Haskell's
+  //. `(.)` function.
   //.
-  //. See also [`B`](#B) and [`pipe`](#pipe).
+  //. This is the B combinator from combinatory logic.
```
davidchambers: Let's keep the link to `pipe`:

_See also [`pipe`][1]._


[1]: https://sanctuary.js.org/#pipe

Avaq: Oh dear. Good catch. :zap:

davidchambers:
I pushed three commits to #216 to complete the tasks on the to-do list. How do they look, @Avaq?

davidchambers:
> Maybe we should rename the remaining combinators to their full names.

Once #216 is merged we'll be left with `I`, `K`, `A`, and `T`. I'm open to renaming these, but I don't think this should delay the v0.12.0 release. :)

PR #331, opened 2017-02-11
use yarn instead of npm

### LINKED ISSUE #265: use Yarn
npm is a piece of software which is quite possibly beyond repair. [Yarn](https://code.facebook.com/posts/1840075619545360) strikes me as a well-considered, sane piece of software. I :heart: that `yarn` is deterministic, whereas `npm install` can produce different directory structures given different starting states.

I can think of three ways to use Yarn:
- have `make setup` run `npm install yarn` then `yarn`;
- rely on a globally installed `yarn` (as we currently do with `npm`); or
- support `NPM=/path/to/global/yarn make setup`.

Does anyone have a preference for one of these three options?

svozza:
Closes #265. I wonder do we need the `rm -rf node_modules` part of `circle.yml` now? Also, if this works I will do the rest of the repos tomorrow.


svozza:
Hmmm, not working. I might not be able to investigate until tomorrow.

svozza:
I literally copy and pasted the example from [here](https://circleci.com/docs/yarn/) and it still didn't work.

svozza:
So it looks like you have to install Yarn in a prebuild step. All working now.

davidchambers:
What's our strategy for keeping __yarn.lock__ up to date?

Makefile
```diff
@@ -1,7 +1,7 @@
 DOCTEST = node_modules/.bin/doctest --module commonjs --prefix .
 ESLINT = node_modules/.bin/eslint --config node_modules/sanctuary-style/eslint-es3.json --env es3
 ISTANBUL = node_modules/.bin/istanbul
-NPM = npm
+YARN = yarn
```
davidchambers: :abc:

svozza:
So it looks like we can add a check to the build script:

https://yarnpkg.com/en/docs/cli/check/

I'll have a play around with it.

circle.yml
```diff
@@ -1,13 +1,19 @@
 dependencies:
+  pre:
+    - curl -o- -L https://yarnpkg.com/install.sh | bash
```
davidchambers: I like to avoid shorthands when shell scripting. I particularly dislike programs which permit `-abc123` as shorthand for `-a -b -c 123` – how confusing!

Here I'd like to see `--location --output -` in place of `-o- -L`. Isn't `--output -` the default behaviour, though? `--location` should be sufficient.

circle.yml
```diff
@@ -1,13 +1,19 @@
...
   override:
     - printf '%s\n' color=false progress=false >.npmrc
     - rm -rf node_modules
     - case $CIRCLE_NODE_INDEX in 0) make setup ;; 1) nvm install 6 && nvm exec 6 make setup ;; 2) nvm install 7 && nvm exec 7 make setup ;; esac:
         parallel: true
+  cache_directories:
+      - ~/.cache/yarn
```
davidchambers: :arrow_left: :arrow_left:

circle.yml
```diff
@@ -1,13 +1,19 @@
...
 
 machine:
   node:
     version: 4
+  environment:
+      PATH: "${PATH}:${HOME}/${CIRCLE_PROJECT_REPONAME}/node_modules/.bin"
```
davidchambers: :arrow_left: :arrow_left:

circle.yml
```diff
@@ -1,13 +1,19 @@
 dependencies:
+  pre:
+    - curl -o- -L https://yarnpkg.com/install.sh | bash
   override:
     - printf '%s\n' color=false progress=false >.npmrc
     - rm -rf node_modules
```
davidchambers: > I wonder do we need the `rm -rf node_modules` part of `circle.yml` now?

Let's try removing the line above. Also, the line above that is no longer necessary.

circle.yml
svozza: I've no idea really about the default behaviour, I just copied this from a Gist I found. I'll leave `-o` out and see if it still works.

davidchambers: I ran `curl --location https://yarnpkg.com/install.sh | cat` and got the expected result. :)

svozza: Actually `--location` isn't even necessary so I've removed that too.

circle.yml
svozza: :zap:

circle.yml
svozza: :zap

circle.yml
svozza: :zap:

Makefile
svozza: :zap:

svozza:
I'll update the other repos when we're happy with this one. Btw, there is so much repetition in the build scripts across all of our repos, it would be really nice to centralise it in one place.

circle.yml
davidchambers: `--location` may not currently be necessary but I think it's a good idea to include it. If the canonical location of the file were to change, and the current location were to redirect to the new location, it would be nice for the Yarn installation to continue to work.

Makefile
```diff
@@ -63,7 +63,8 @@ release-major release-minor release-patch:
 
 .PHONY: setup
 setup:
-	$(NPM) install
+	$(YARN)
+	$(YARN) check
```
davidchambers: Is this the right place for `yarn check`? I was thinking it would be part of the `lint` target. I'd like `make lint` to fail for me locally were I to update a dependency in __package.json__ without regenerating __yarn.lock__. This leads naturally to the question of which command one runs to update __yarn.lock__. Let's make it clear: `make yarn.lock`.

davidchambers:
> I'll update the other repos when we're happy with this one.

Sounds good to me. I've limited my comments to this pull request with this in mind.

> Btw, there is so much repetition in the build scripts across all of our repos, it would be really nice to centralise it in one place.

I agree. Perhaps we need a makefile to generate the makefiles? :stuck_out_tongue_winking_eye:

Makefile
svozza: I'm not quite sure I follow. Should I rename `setup` to `yarn.lock`?

svozza:
Haha. I had a little dig around the CircleCI docks but there doesn't seem to be a way to share config files across multiple projects.

Makefile
davidchambers: I only spent a few moments reading about `yarn check`; it's possible I misunderstood its purpose. I believe it is run to confirm that __yarn.lock__ is consistent with __package.json__. If __package.json__ were to specify `foo@2.x.x` while __yarn.lock__ specified `foo@1.2.3`, `yarn lock` would exit with a non-zero exit code. This strikes me as a linting concern rather than a set-up concern. I would like to make this assertion each time I run `make lint` (assuming that it's not prohibitively slow). Let's move `$(YARN) check` from the `setup` target to the `lint` target.

If we're to require a contributor to update __yarn.lock__ if she updates a dependency in __package.json__, we shouldn't require her to spend five minutes reading the Yarn documentation to find out how to do so. Let's specify the build step in the makefile:

```make
yarn.lock:
	$(YARN) whatever-one-does-to-generate-lockfile
```

svozza: The lock file gets generated when you run the `yarn` command, that's what I meant by renaming `setup` to `yarn.lock`, that's when the file is either generated or updated.

davidchambers: Oh, I see. For some reason I assumed there to be separate commands for generating the lockfile and installing dependencies, but it makes sense that they happen together.

I'd like to continue to be able to run `make setup` to install dependencies. I suggest this structure:

```make
.PHONY: setup
setup: yarn.lock

yarn.lock: package.json
	$(YARN)
```

svozza: Hmmm, `Make` doesn't seem to like the dot in `yarn.lock`.

svozza: How about an underscore?

davidchambers: The dot should not be a problem. After all, many filenames include an extension, and we currently have a `README.md` target in the makefile. What's the error you're seeing?

svozza: If I use the dot, I get this output:

`make: Nothing to be done for 'setup'.`

Replacing hte dot with an underscore makes the command run properly.

davidchambers: That's Make telling you that the dependencies of the `setup` target have not changed, so there's no work to be done.

```make
.PHONY: foo
foo: bar

bar:
	echo hello >'$@'
```

```console
$ make foo
echo hello >'bar'

$ make foo
make: Nothing to be done for `foo'.
```

svozza: But I've deleted my `node_modules` folder and it's still giving me that message.

davidchambers: The snippet I suggested earlier specifes that __yarn.lock__ depends only on __package.json__. Unless the modification date of __package.json__ changes, Make will refuse to rebuild __yarn.lock__. To force Make to do your bidding you could run `touch package.json && make yarn.lock` or simply `rm yarn.lock && make yarn.lock`.

svozza: Aha. I see. Does that not mean that the very first time a contributor clones the repo that `make setup` won't install the project dependencies?

davidchambers: Good point! I think what we actually need is this:

```make
.PHONY: setup
setup:
	$(YARN)

yarn.lock: package.json
	$(YARN)
```

svozza: :zap:

davidchambers:
This looks good to me! Thank you, Stefano. :)

Let's merge this as soon as I've released v0.12.0, just in case it has unintended consequences.

svozza:
Sounds good, I most definitely don't want to be the person that screws that up! In the meantime I'll tidy up the other PRs.

svozza:
I've rebased against master here and updated all the other repos.

Makefile
```diff
@@ -63,7 +64,10 @@ release-major release-minor release-patch:
 
 .PHONY: setup
 setup:
-	$(NPM) install
+	$(YARN)
```
davidchambers: Let's change this to `$(YARN) install --pure-lockfile` to avoid updating __yarn.lock__ here (as that's not our intention).

Makefile
```diff
@@ -63,7 +64,10 @@ release-major release-minor release-patch:
...
+
+yarn.lock: package.json
+	$(YARN)
```
davidchambers: Is `yarn` short for `yarn install`? If so let's use `$(YARN) install` here for clarity.

circle.yml
```diff
@@ -1,13 +1,17 @@
 dependencies:
+  pre:
+    - curl https://yarnpkg.com/install.sh | bash
```
davidchambers: Let's pass the `--location` flag to `curl`.

Makefile
svozza: The canonical usage is `yarn`,  they only have `install` for familiarity with npm.

davidchambers: Okay. It's fine to use the short form in that case. :)

circle.yml
svozza: :zap:

Makefile
svozza: :zap:

davidchambers:
I'm sorry this pull request is taking quite a long time to review, Stefano. It's an important change. It's also important that we have a clear plan. The one thing that's not clear to me is the plan for updating __yarn.lock__. Should we have a scheduled job which creates a pull request to update this every week? Should one of us get into the habit of doing this manually? Should we wait until a change is made to __package.json__ and accept that it could add hundreds of lines of unrelated dependency updates to the patch? Should we specify exact versions in __package.json__ to prevent __yarn.lock__ from falling behind?

svozza:
The way we do it in work is that lock down our dependency versions and if you update any of the dependencies in `package.json` you also check in the new lock file with that change. We also run `yarn outdated` as a post build step to let us know if we're falling behind. We have specific update dependencies tasks in our backlog to keep things up to date.

davidchambers:
> The way we do it in work is that lock down our dependency versions and if you update any of the dependencies in `package.json` you also check in the new lock file with that change.

This sounds good to me. Could you lock each of the dependencies in __package.json__ to the most recent version satisfying the specification? `xyz@2.0.x` will become `xyz@2.0.1`, for example. You should then rebuild __yarn.lock__.

svozza:
:zap:

davidchambers:
I've realized that even if we specify exact versions for all our dependencies, __yarn.lock__ will *still* get out of date due to new releases of our dependencies' dependencies. If we depend on `foo@1.2.3` which in turn depends on `bar@4.5.x`, our __yarn.lock__ will get out of date if and when `bar@4.5.6` is released. If the plan is to update __yarn.lock__ in any commit which updates dependencies in __package.json__, these commits may contain hundreds of lines of unrelated changes. I'd like to avoid this.

I'm beginning to think that Yarn is a good fit for applications but not for libraries. Perhaps [ied][1] is a more natural fit. Are you still using ied, @Avaq?


[1]: https://github.com/alexanderGugel/ied

svozza:
>If the plan is to update yarn.lock in any commit which updates dependencies in package.json, these commits may contain hundreds of lines of unrelated changes. I'd like to avoid this.

I'm actually fine with this, the diff is normally hidden by the UI anyway and at least we have a record in source control of when our transitative dependencies have changed.

davidchambers:
The convention when updating dependencies could be to add one commit *just* to update __yarn.lock__, followed by one commit to update __package.json__ and __yarn.lock__. The changes to __yarn.lock__ in the second commit would be minimal. I could live with this. :)

svozza:
Yes, I think that makes a lot of sense. We can add it to the `Contributing` guide.

davidchambers:
After giving this more thought, I'm now of the opinion that we should not merge this pull request. npm is far from perfect, but I've never had problems with `rm -r node_modules && npm install` other than long waits. The benefits of Yarn are outweighed by the complexity that __yarn.lock__ would add to our development process.

To be clear, I very much favour using Yarn for *applications*, but I don't think it's right for *libraries*.

I'd love to find a lightweight approach to allow Sanctuary developers to use Yarn if desired. I'd like `make setup` to run `yarn install` rather than `npm install` if `yarn` is available, and we could add __yarn.lock__ to __.gitignore__.

I very much appreciate the considerable work you put into this pull request, @svozza. Nevertheless, we must do what we believe to best for the project. Let's not deeply integrate Yarn at this stage.

PR #334, opened 2017-02-13
Static imports to aid bundle traversing tools

JAForbes:
Browserify require static import's in order to establish a dependency tree.

This PR addresses this issue by replacing e.g. `require(deps[0])` with `require('sanctuary-def')`

### DIFF
--- a/index.js
+++ b/index.js
@@ -207,18 +207,16 @@
 
   'use strict';
 
-  //  deps :: Array String
-  var deps = [
-    'sanctuary-def',
-    'sanctuary-type-classes',
-    'sanctuary-type-identifiers'
-  ];
-
   /* istanbul ignore else */
   if (typeof module === 'object' && typeof module.exports === 'object') {
-    module.exports = f(require(deps[0]), require(deps[1]), require(deps[2]));
+    module.exports = f(require('sanctuary-def'),
+                       require('sanctuary-type-classes'),
+                       require('sanctuary-type-identifiers'));
   } else if (typeof define === 'function' && define.amd != null) {
-    define(deps, f);
+    define(['sanctuary-def',
+            'sanctuary-type-classes',
+            'sanctuary-type-identifiers'],
+           f);
   } else {
     self.sanctuary = f(self.sanctuaryDef,
                        self.sanctuaryTypeClasses,


index.js
```diff
@@ -207,18 +207,11 @@
...
-  var deps = [
-    'sanctuary-def',
-    'sanctuary-type-classes',
-    'sanctuary-type-identifiers'
-  ];
-
   /* istanbul ignore else */
   if (typeof module === 'object' && typeof module.exports === 'object') {
-    module.exports = f(require(deps[0]), require(deps[1]), require(deps[2]));
+    module.exports = f(require('sanctuary-def'), require('sanctuary-type-classes'), require('sanctuary-type-identifiers'));
```
davidchambers: I expect the linter to complain about the length of this line. If so, let's use this instead:

```javascript
    module.exports = f(require('sanctuary-def'),
                       require('sanctuary-type-classes'),
                       require('sanctuary-type-identifiers'));
```

JAForbes:
```
    define(['sanctuary-def', 
      'sanctuary-type-classes', 
      'sanctuary-type-identifiers'], f);
```

Something like this too @davidchambers ?

JAForbes:
Also noticed this in the `Makefile`

`--rule 'max-len: [off]' \`

index.js
```diff
@@ -207,18 +207,15 @@
...
   if (typeof module === 'object' && typeof module.exports === 'object') {
-    module.exports = f(require(deps[0]), require(deps[1]), require(deps[2]));
+    module.exports = f(require('sanctuary-def'),
+                       require('sanctuary-type-classes'),
+                       require('sanctuary-type-identifiers'));
   } else if (typeof define === 'function' && define.amd != null) {
-    define(deps, f);
+    define(['sanctuary-def',
+      'sanctuary-type-classes',
+      'sanctuary-type-identifiers'], f);
```
davidchambers: My rule for line wrapping is that expressions at the same level semantically should be at the same level visually. I'd prefer this wrapping:

```javascript
define(['sanctuary-def',
        'sanctuary-type-classes',
        'sanctuary-type-identifiers'],
       f);
```

I seem to recall introducing `deps` to avoid the above, because `f` looked lonely. :cry:

@JAForbes, could you make this change and squash your commits while you're at it? I'll then merge this pull request and release v0.12.1. :)

davidchambers:
:deciduous_tree:

PR #344, opened 2017-02-17
Add S.complement

### LINKED ISSUE #342: complement :: (a -> Boolean) -> a -> Boolean
This is a beautiful implementation:

```javascript
//    complement :: (a -> Boolean) -> a -> Boolean
const complement = S.compose(S.not);
```

We tend not to use curried functions internally, so this is probably not the implementation we should choose. I do enjoy finding uses for partially applying `compose`, though. :)

cust0dian:
Fixes #342

### DIFF
--- a/index.js
+++ b/index.js
@@ -2560,6 +2560,8 @@
   //.
   //. Boolean "not".
   //.
+  //. See also [`complement`](#complement).
+  //.
   //. ```javascript
   //. > S.not(false)
   //. true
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -2577,6 +2579,8 @@
   //. Takes a unary predicate and a value of any type, and returns logical
   //. negation of applying predicate to the value.
   //.
+  //. See also (`not`)[#not]
```
cust0dian: Wasn't sure if this is a desired reference so left it in a separate commit for review, let me know if this good to be squashed into main commit

davidchambers:
Thank you for the pull request, @cust0dian. Very nice!

index.js
davidchambers: I think the reference is useful. I have a small request: please add a full stop (`.`) at the end of the line. :)

index.js
```diff
@@ -2572,6 +2574,25 @@
   }
   S.not = def('not', {}, [$.Boolean, $.Boolean], not);
 
+  //# complement :: (a -> Boolean) -> a -> Boolean
+  //.
+  //. Takes a unary predicate and a value of any type, and returns logical
+  //. negation of applying predicate to the value.
```
davidchambers: I'd like to see "the" used in two more places:

> Takes a unary predicate and a value of any type, and returns <ins>the</ins> logical negation of applying <ins>the</ins> predicate to the value.

Other than that, this description is excellent.

index.js
```diff
@@ -2572,6 +2574,25 @@
...
+  //.
+  //. See also [`not`](#not).
+  //.
+  //. ```javascript
+  //. > Number.isInteger('foo')
+  //. false
+  //.
+  //. > S.complement(Number.isInteger, 'foo')
+  //. true
+  //. ```
```
davidchambers: I like the choice of `Number.isInteger` as the predicate, but I'd like to pretend that its type is `Number -> Boolean` rather than `Any -> Boolean`. I suggest:

```javascript
> Number.isInteger(42)
true

> S.complement(Number.isInteger, 42)
false
```

cust0dian:
@davidchambers all your comments are addressed now ⚡️

PR #348, opened 2017-02-21
logic: add S.when and S.unless

### LINKED ISSUE #347: Add S.when and S.unless
Create versions of [R.when](http://ramdajs.com/0.23.0/docs/#when) and [R.unless](http://ramdajs.com/0.23.0/docs/#unless).

gabejohnson:
Closes #347

### DIFF
--- a/index.js
+++ b/index.js
@@ -2613,6 +2613,42 @@
   }
   S.ifElse = def('ifElse', {}, [Pred(a), Fn(a, b), Fn(a, b), a, b], ifElse);
 
+  //# when :: (a -> Boolean) -> (a -> a) -> a -> a
+  //.
+  //. Takes a unary predicate, a unary function, and a value of any type, and
+  //. returns the result of applying the function to the value if the value
+  //. satisfies the predicate; the value otherwise.
+  //.
+  //. ```javascript
+  //. > S.when(x => x >= 0, Math.sqrt, 16)
+  //. 4
+  //.
+  //. > S.when(x => x >= 0, Math.sqrt, -1)
+  //. -1
+  //. ```
+  function when(pred, f, x) {
+    return ifElse(pred, f, I, x);
+  }
+  S.when = def('when', {}, [Pred(a), Fn(a, a), a, a], when);
+
+  //# unless :: (a -> Boolean) -> (a -> a) -> a -> a
+  //.
+  //. Takes a unary predicate, a unary function, and a value of any type, and
+  //. returns the result of applying the function to the value if the value
+  //. does not satisfy the predicate; the value otherwise.
+  //.
+  //. ```javascript
+  //. > S.unless(x => x < 0, Math.sqrt, 16)
+  //. 4
+  //.
+  //. > S.unless(x => x < 0, Math.sqrt, -1)
+  //. -1
+  //. ```
+  function unless(pred, f, x) {
+    return ifElse(pred, I, f, x);
+  }
+  S.unless = def('unless', {}, [Pred(a), Fn(a, a), a, a], unless);
+
   //# allPass :: Array (a -> Boolean) -> a -> Boolean
   //.
   //. Takes an array of unary predicates and a value of any type
--- a/test/unless.js
+++ b/test/unless.js
@@ -1,0 +1,19 @@
+'use strict';
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+
+
+test('unless', function() {
+
+  eq(typeof S.unless, 'function');
+  eq(S.unless.length, 3);
+  eq(S.unless.toString(), 'unless :: (a -> Boolean) -> (a -> a) -> a -> a');
+
+  function lt0(x) { return x < 0; }
+
+  eq(S.unless(lt0, Math.sqrt, 16), 4);
+  eq(S.unless(lt0, Math.sqrt, -1), -1);
+
+});
--- a/test/when.js
+++ b/test/when.js
@@ -1,0 +1,19 @@
+'use strict';
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+
+
+test('when', function() {
+
+  eq(typeof S.when, 'function');
+  eq(S.when.length, 3);
+  eq(S.when.toString(), 'when :: (a -> Boolean) -> (a -> a) -> a -> a');
+
+  function gte0(x) { return x >= 0; }
+
+  eq(S.when(gte0, Math.sqrt, 16), 4);
+  eq(S.when(gte0, Math.sqrt, -1), -1);
+
+});


davidchambers:
If you replace "Implements" with "Closes" in the description, @gabejohnson, the related issue will automatically be closed when this pull request is merged. :)

See [Closing Issues via Pull Requests][1] for details.


[1]: https://github.com/blog/1506-closing-issues-via-pull-requests

index.js
```diff
@@ -2613,6 +2613,42 @@
...
+  //. returns the result of applying the function to the value if the value
+  //. satisfies the predicate; the value otherwise.
+  //.
+  //. ```javascript
+  //. > S.when(x => x < 0, Math.abs, -1)
+  //. 1
+  //.
+  //. > S.when(x => x < 0, Math.abs, 4)
+  //. 4
+  //. ```
```
davidchambers: The problem with these examples is that `Math.abs(x)` is more direct than `S.when(x => x < 0, Math.abs, x)`. Perhaps we should mirror the `unless` examples:

```javascript
> S.when(x => x >= 0, Math.sqrt, 16)
4

> S.when(x => x >= 0, Math.sqrt, -1)
-1
```

index.js
```diff
@@ -2613,6 +2613,42 @@
...
+  //. returns the result of applying the function to the value if the value
+  //. does not satisfy the predicate; the value otherwise.
+  //.
+  //. ```javascript
+  //. > S.unless(x => x < 0, Math.sqrt, -1)
+  //. -1
+  //.
+  //. > S.unless(x => x < 0, Math.sqrt, 16)
+  //. 4
+  //. ```
```
davidchambers: I have a slight preference for showing the example that applies the `a -> b` function first; the failure case second. What do you think?

index.js
```diff
@@ -2613,6 +2613,42 @@
   }
   S.ifElse = def('ifElse', {}, [Pred(a), Fn(a, b), Fn(a, b), a, b], ifElse);
 
+  //# when :: (a -> Boolean) -> (a -> b) -> a -> b
```
safareli: Signature is not quite correct.

We have:
```haskell
when :: (a -> Boolean) -> (a -> b) -> a -> b
```

If second function is for example from Int to String then you can't give back String in if predicate fails:
```js
// :: (Int -> Boolean) -> (Int -> String) -> Int -> String
when(a => a < 0,ai => `got ${a}`, -1) // will get error as `Int` is not `String`
```

safareli: btw I think it would be nice to use property based testing instead of just having cople hardcoded test cases, to avoid such issues.

safareli: in ramda type is:
`(a → Boolean) → (a → a) → a → a`
So i think it we should go this way or:
`(a → Boolean) → (a → b) → a → Either a b`

davidchambers: Great catch, Irakli! Returning `Either a b` seems nice, but if we do so the functions will not be sugar for `ifElse` with `I` in second or third position. The `Either a b` version may also be worth adding to the library, but we should wait to discover a compelling use case. :)

rjmk: I suspect `(a -> Boolean) -> (a -> a) -> a -> a` will not prove very useful (and that the Ramda version is probably often used not according to its type signature) but am happy to give it a go and see if I'm proved wrong

davidchambers: @JAForbes, on Gitter:

> I use `R.when` all the time, particularly with streams

Could you provide an example, James? Seeing a real use case would be helpful, I think.

safareli: I'm not sure on which version would be the most useful, but we should choose one which is "correct"

davidchambers: > I'm not sure on which version would be the most useful, but we should choose one which is "correct"

Absolutely! The pull request as it currently stands is not correct. I'm currently leaning towards the Ramda version, but I have a suspicion that I'm missing some deeper truth. This is a vague statement, I realize. ;)

JAForbes: Sure.  It's helpful when you have some transformation you want to do to a stream in certain cases.

E.g. say we want to `concat` some warning messages in a UI if a form field is empty

```
form.map( 
  when(
     view( fields, any( isEmpty) )
     ,over( warnings, concat(['Please fill in all form fields'])
   )
)
```

As we emit form objects in our stream, we'll conditionally add validation messages if necessary.

Notice there's no need for an explicit else case.  The else case is to simply leave the structure as is.  You could use `ifElse( ..., identity)`, but it feels cumbersome.

safareli: Type for `ifElse(_, identity)` would be `(a → Boolean) → (a → a) → a → a`  so looks like it still useful

cust0dian: > we should choose one which is "correct"

Just my 2¢:

The only type-correct signature of `ifElse` is the one currently documented:

```haskell
ifElse :: (a -> Bool) -> (a -> b) -> (a -> b) -> a -> b
--                       ^------^    ^------^
--                           \          /
--                     have to be the same type
```

So when we fix one of the "branches" with `id :: a -> a`, we get:

```haskell
-- a == b
ifElse :: (a -> Bool) -> (a -> a) -> (a -> a) -> a -> a

-- with `id` as one of the "branches"
when :: (a -> Bool) -> (a -> a) -> a -> a
```

To represent a case where we might run "when" function or leave value as is, we'd need to have @safareli's type with `Either` and not rely on `ifElse` as implementation since they have different semantics.

davidchambers: What is wrong with `when :: (a -> a) -> a -> a`, @cust0dian? I see no problem with implementing this in terms of `ifElse`, since there's no requirement that `a` and `b` be replaced with *different* types.

gabejohnson: So I'll change the type signatures to:
```javascript
when :: (a -> Boolean) -> (a -> a) -> a -> a
```
and
```javascript
unless :: (a -> Boolean) -> (a -> a) -> a -> a
```
> The Either a b version may also be worth adding to the library, but we should wait to discover a compelling use case. :)

Agreed.

gabejohnson: Just saw #350. I'll wait on any changes until that discussion resolves.

davidchambers: > Just saw #350. I'll wait on any changes until that discussion resolves.

I don't think `tagBy` invalidates `ifElse`. I think we should add `tagBy` but keep `ifElse`. If we're keeping `ifElse` we should add `when` and `unless` with type `(a -> Boolean) -> (a -> a) -> a -> a`.

cust0dian: > What is wrong with `when :: (a -> a) -> a -> a`, @cust0dian?

@davidchambers sorry, not sure I follow that signature: `when :: (a -> a) -> a -> a` looks like `id` to me 😄 

Did you mean `when :: (a -> Bool) -> (a -> a) -> a -> a`? In that case there's nothing wrong with it, I was just commenting that it's the only correct type if we follow definitions of `ifElse` and `id`

davidchambers: Oops, I did indeed forget the `(a -> Boolean)`. :)

> I was just commenting that it's the only correct type if we follow definitions of `ifElse` and `id`

Yes, you're right.

gabejohnson: Okay. Then I'll update the signature and make the other changes you requested @davidchambers.

index.js
gabejohnson: It's your boat 😃 . Consistency's important.

index.js
gabejohnson: I think I'll change the test cases as well. It's always good to make sure your examples don't break.

davidchambers: The examples are actually checked by [doctest][1]. ;)


gabejohnson: Oops! Do you want me to change them back?

davidchambers: I was just letting you know that you can sleep peacefully knowing that the examples in the readme and on the website are always correct. :)

PR #354, opened 2017-02-24
Replace Type constructors in env with actual Types

Avaq:
Based on sanctuary-js/sanctuary-def#123

This fixes the inconsistent type of `env`, allowing it to be passed into `$.test`. We should release this as a patch.

### DIFF
--- a/test/env.js
+++ b/test/env.js
@@ -1,0 +1,22 @@
+'use strict';
+
+var $ = require('sanctuary-def');
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+
+
+//  Type :: Type
+var Type = $.NullaryType(
+  'sanctuary/Type',
+  '',
+  function(x) { return S.type(x) === 'sanctuary-def/Type'; }
+);
+
+test('env', function() {
+
+  eq(typeof S.env, 'object');
+  eq($.test([], $.Array(Type), S.env), true);
+
+});
--- a/index.js
+++ b/index.js
@@ -349,13 +349,13 @@
   var defaultEnv = Z.concat($.env, [
     $.FiniteNumber,
     $.NonZeroFiniteNumber,
-    $Either,
+    $Either($.Unknown, $.Unknown),
     Fn($.Unknown, $.Unknown),
     $.GlobalRegExp,
     $.NonGlobalRegExp,
     $.Integer,
-    $Maybe,
-    $.Pair,
+    $Maybe($.Unknown),
+    $.Pair($.Unknown, $.Unknown),
     $.RegexFlags,
     $.ValidDate,
     $.ValidNumber


test/env.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+var S = require('..');
+var $ = require('sanctuary-def');
+var type = require('sanctuary-type-identifiers');
+
+var eq = require('./internal/eq');
+
+
+var $Type = $.NullaryType(
```
Avaq: It would be nice if `sanctuary-def` exports `Type`, so I don't have to create it myself.

Avaq:
If `sanctuary-def` exports `Type`, we would also be able to use it in the [definition of `Options`](https://github.com/sanctuary-js/sanctuary/pull/354/files#diff-168726dbe96b3ce427e7fedce31bb0bcR365). For now I don't think we should go for that, as it may be considered "breaking".

davidchambers:
Could you update this pull request's destination branch to the temporary `0.12.x` branch I just created?

test/env.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+var S = require('..');
+var $ = require('sanctuary-def');
+var type = require('sanctuary-type-identifiers');
+
+var eq = require('./internal/eq');
```
davidchambers: Let's group the imports like so:

```javascript
var $ = require('sanctuary-def');
var type = require('sanctuary-type-identifiers');

var S = require('..');

var eq = require('./internal/eq');
```

test/env.js
davidchambers: Could we name this `Type` rather than `$Type`? Regardless, let's add a type signature. :)

test/env.js
```diff
@@ -0,0 +1,21 @@
...
+var $Type = $.NullaryType(
+  'sanctuary/Type',
+  '',
+  function(x) { return type(x) === 'sanctuary-def/Type'; }
+);
+
+test('env', function() {
+
+  eq(typeof S.env, 'object');
+  eq($.test([], $.Array($Type), S.env), true);
```
davidchambers: This is an elegant assertion!

test/env.js
```diff
@@ -0,0 +1,21 @@
...
+var $ = require('sanctuary-def');
+var type = require('sanctuary-type-identifiers');
+
+var eq = require('./internal/eq');
+
+
+var $Type = $.NullaryType(
+  'sanctuary/Type',
+  '',
+  function(x) { return type(x) === 'sanctuary-def/Type'; }
```
davidchambers: Did you consider using `S.type` here?

test/env.js
Avaq: :zap:

test/env.js
Avaq: :zap:

test/env.js
Avaq: :zap:

PR #356, opened 2017-02-27
foldable: add S.takeWhile and S.dropWhile

### LINKED ISSUE #349: takeWhile and dropWhile
The Ramda functions [takeWhile](http://ramdajs.com/docs/#takeWhile) and [dropWhile](http://ramdajs.com/docs/#dropWhile) should both have the type signature:
```javascript
(a -> Boolean) -> List a -> List a
```

I mentioned `takeUntil` and `dropUntil` in the Gitter room, but thinking on it more they don't bring much to the table (the way I was thinking of them):

```javascript
const takeUntil = pred => S.takeWhile(S.complement(pred));
```

gabejohnson:
Closes #349

### DIFF
--- a/index.js
+++ b/index.js
@@ -1039,6 +1039,70 @@
       [Fn(a, $.Boolean), m(a), m(a)],
       Z.filterM);
 
+  //# takeWhile :: (Foldable f, Alternative f) => (a -> Boolean) -> f a -> f a
+  //.
+  //. Discards the first inner value which does not satisfy the predicate, and
+  //. all subsequent inner values.
+  //.
+  //. ```javascript
+  //. > S.takeWhile(S.odd, [3, 3, 3, 7, 6, 3, 5, 4])
+  //. [3, 3, 3, 7]
+  //.
+  //. > S.takeWhile(S.even, [3, 3, 3, 7, 6, 3, 5, 4])
+  //. []
+  //. ```
+  function Array$takeWhile(pred, xs) {
+    var idx = 0;
+    while (idx < xs.length && pred(xs[idx])) idx += 1;
+    return xs.slice(0, idx);
+  }
+
+  function takeWhile(pred, xs) {
+    if (Array.isArray(xs)) return Array$takeWhile(pred, xs);
+    var done = false;
+    function takeWhileReducer(xs, x) {
+      return !done && pred(x) ? append(x, xs) : (done = true, xs);
+    }
+    return Z.reduce(takeWhileReducer, Z.empty(xs.constructor), xs);
+  }
+  S.takeWhile =
+  def('takeWhile',
+      {f: [Z.Foldable, Z.Alternative]},
+      [Pred(a), f(a), f(a)],
+      takeWhile);
+
+  //# dropWhile :: (Foldable f, Alternative f) => (a -> Boolean) -> f a -> f a
+  //.
+  //. Retains the first inner value which does not satisfy the predicate, and
+  //. all subsequent inner values.
+  //.
+  //. ```javascript
+  //. > S.dropWhile(S.odd, [3, 3, 3, 7, 6, 3, 5, 4])
+  //. [6, 3, 5, 4]
+  //.
+  //. > S.dropWhile(S.even, [3, 3, 3, 7, 6, 3, 5, 4])
+  //. [3, 3, 3, 7, 6, 3, 5, 4]
+  //. ```
+  function Array$dropWhile(pred, xs) {
+    var idx = 0;
+    while (idx < xs.length && pred(xs[idx])) idx += 1;
+    return xs.slice(idx);
+  }
+
+  function dropWhile(pred, xs) {
+    if (Array.isArray(xs)) return Array$dropWhile(pred, xs);
+    var done = false;
+    function dropWhileReducer(xs, x) {
+      return !done && pred(x) ? xs : (done = true, append(x, xs));
+    }
+    return Z.reduce(dropWhileReducer, Z.empty(xs.constructor), xs);
+  }
+  S.dropWhile =
+  def('dropWhile',
+      {f: [Z.Foldable, Z.Alternative]},
+      [Pred(a), f(a), f(a)],
+      dropWhile);
+
   //. ### Combinator
 
   //# I :: a -> a
--- a/test/takeWhile.js
+++ b/test/takeWhile.js
@@ -1,0 +1,22 @@
+'use strict';
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+
+
+test('takeWhile', function() {
+
+  eq(typeof S.takeWhile, 'function');
+  eq(S.takeWhile.length, 2);
+  eq(S.takeWhile.toString(), 'takeWhile :: (Foldable f, Alternative f) => (a -> Boolean) -> f a -> f a');
+
+  eq(S.takeWhile(S.odd, [3, 3, 3, 7, 6, 3, 5, 4]), [3, 3, 3, 7]);
+  eq(S.takeWhile(S.even, [3, 3, 3, 7, 6, 3, 5, 4]), []);
+  eq(S.takeWhile(S.odd, []), []);
+
+  eq(S.takeWhile(S.odd, S.Just(1)), S.Just(1));
+  eq(S.takeWhile(S.even, S.Just(1)), S.Nothing);
+  eq(S.takeWhile(S.odd, S.Nothing), S.Nothing);
+
+});
--- a/test/dropWhile.js
+++ b/test/dropWhile.js
@@ -1,0 +1,22 @@
+'use strict';
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+
+
+test('dropWhile', function() {
+
+  eq(typeof S.dropWhile, 'function');
+  eq(S.dropWhile.length, 2);
+  eq(S.dropWhile.toString(), 'dropWhile :: (Foldable f, Alternative f) => (a -> Boolean) -> f a -> f a');
+
+  eq(S.dropWhile(S.odd, [3, 3, 3, 7, 6, 3, 5, 4]), [6, 3, 5, 4]);
+  eq(S.dropWhile(S.even, [3, 3, 3, 7, 6, 3, 5, 4]), [3, 3, 3, 7, 6, 3, 5, 4]);
+  eq(S.dropWhile(S.odd, []), []);
+
+  eq(S.dropWhile(S.odd, S.Just(1)), S.Nothing);
+  eq(S.dropWhile(S.even, S.Just(1)), S.Just(1));
+  eq(S.dropWhile(S.odd, S.Nothing), S.Nothing);
+
+});


gabejohnson:
I put these two functions near `filter` as I think they deal with `Foldable`, take a predicate and "reduce" the collection (in the case of `Array`).

index.js
```diff
@@ -1039,6 +1039,78 @@
       [Fn(a, $.Boolean), m(a), m(a)],
       Z.filterM);
 
+  //# takeWhile :: (Foldable f, Monoid (f a)) =>  (a -> Boolean) -> f a -> f a
```
davidchambers: Nit: There's an extra space after the `=>`.

index.js
```diff
@@ -1039,6 +1039,78 @@
...
+  //.
+  //. Returns the first elements for which the predicate returns true.
```
davidchambers: This doesn't sound quite right to me for a non-array argument. How about the following?

_Discards the first inner value which does not satisfy the predicate, and all subsequent inner values._

index.js
```diff
@@ -1039,6 +1039,78 @@
...
+  //. > S.takeWhile(S.K(true), S.Just(1))
+  //. Just(1)
+  //.
+  //. > S.takeWhile(S.K(false), S.Just(1))
+  //. Nothing
+  //. ```
+  function takeWhile(p, xs) {
+    if (Array.isArray(xs)) {
+      var result = [];
+      if (xs.length === 0) return result;
```
svozza: Probably makes more sense to create the result array after checking the length of `xs`.

index.js
```diff
@@ -1039,6 +1039,78 @@
       [Fn(a, $.Boolean), m(a), m(a)],
       Z.filterM);
 
+  //# takeWhile :: (Foldable f, Monoid (f a)) =>  (a -> Boolean) -> f a -> f a
+  //.
+  //. Returns the first elements for which the predicate returns true.
+  //.
+  //. ```javascript
+  //. > S.takeWhile(c => c < 'c', ['a', 'b', 'c', 'd', 'e'])
+  //. ['a', 'b']
```
davidchambers: `c => c < 'c'` is slightly confusing due to `c` being used in two ways. `n => n < 3` with `[1, 2, 3, 4, 5]` would be clearer. Even better, though, would be to use an example which demonstrates how this differs from `filter`:

```javascript
> S.takeWhile(S.odd, [3, 3, 3, 7, 6, 3, 5, 4])
[3, 3, 3, 7]
```

index.js
```diff
@@ -1039,6 +1039,78 @@
...
+  //.
+  //. > S.takeWhile(c => c < 'a', ['a', 'b', 'c'])
+  //. []
+  //.
+  //. > S.takeWhile(S.K(true), S.Just(1))
+  //. Just(1)
+  //.
+  //. > S.takeWhile(S.K(false), S.Just(1))
+  //. Nothing
```
davidchambers: Why not use a predicate which references its argument?

```javascript
> S.takeWhile(S.odd, S.Just(1))
Just(1)

> S.takeWhile(S.odd, S.Just(2))
Nothing
```

I don't think we should include these examples, though, as they would be better expressed with `filter`.

index.js
```diff
@@ -1039,6 +1039,78 @@
...
+  //. > S.dropWhile(S.K(true), S.Just(1))
+  //. Nothing
+  //.
+  //. > S.dropWhile(S.K(false), S.Just(1))
+  //. Just(1)
+  //. ```
+  function dropWhile(p, xs) {
+    if (Array.isArray(xs)) {
+      if (xs.length === 0) return [];
+      for (var i = 0; i < xs.length && p(xs[i]); i += 1);
```
svozza: I think `idx` is preferred over `i`.

index.js
```diff
@@ -1039,6 +1039,78 @@
...
+  //. > S.takeWhile(c => c < 'a', ['a', 'b', 'c'])
+  //. []
+  //.
+  //. > S.takeWhile(S.K(true), S.Just(1))
+  //. Just(1)
+  //.
+  //. > S.takeWhile(S.K(false), S.Just(1))
+  //. Nothing
+  //. ```
+  function takeWhile(p, xs) {
```
davidchambers: Let's `:%s/\<p\>/pred/gc` for consistency with existing code.

index.js
```diff
@@ -1039,6 +1039,78 @@
...
+    if (Array.isArray(xs)) {
+      var result = [];
+      if (xs.length === 0) return result;
+      for (var i = 0, x; i < xs.length && p(x = xs[i]); i += 1) result.push(x);
```
davidchambers: Let's `:%s/\<i\>/idx/gc` for consistency with existing code.

svozza: Interestingly, I did a search on the code and we have and neither of our `for` loops use `idx`.

index.js
davidchambers: This line is unnecessary.

index.js
```diff
@@ -1039,6 +1039,78 @@
...
+  //.
+  //. > S.dropWhile(S.K(true), S.Just(1))
+  //. Nothing
+  //.
+  //. > S.dropWhile(S.K(false), S.Just(1))
+  //. Just(1)
+  //. ```
+  function dropWhile(p, xs) {
+    if (Array.isArray(xs)) {
+      if (xs.length === 0) return [];
```
davidchambers: This line is unnecessary.

index.js
```diff
@@ -1039,6 +1039,78 @@
...
+      for (var i = 0; i < xs.length && p(xs[i]); i += 1);
+      return xs.slice(i);
+    }
+    var drop = true;
+    function dropWhileReducer(xs, x) {
+      return drop && p(x) ? xs : (drop = false, append(x, xs));
```
davidchambers: Is this a bit clearer, do you think?

```javascript
return (drop = drop && p(x)) ? xs : append(x, xs);
```

test/takeWhile.js
```diff
@@ -0,0 +1,32 @@
...
+      return a < b;
+    };
+  }
+
+  eq(S.takeWhile(lt(1), [1, 2, 3, 4, 5]), []);
+  eq(S.takeWhile(lt(2), [1, 2, 3, 4, 5]), [1]);
+  eq(S.takeWhile(lt(3), [1, 2, 3, 4, 5]), [1, 2]);
+  eq(S.takeWhile(lt(4), [1, 2, 3, 4, 5]), [1, 2, 3]);
+  eq(S.takeWhile(lt(5), [1, 2, 3, 4, 5]), [1, 2, 3, 4]);
+  eq(S.takeWhile(lt(6), [1, 2, 3, 4, 5]), [1, 2, 3, 4, 5]);
```
davidchambers: Let's change these to demonstrate and assert that `takeWhile` is different from `filter`.

index.js
davidchambers: We don't currently have a single `for` loop which uses an index variable. :)

svozza: Yeah, I guess they're all in `sanctuary-type-classes` now.

index.js
cust0dian: Not sure if a good idea, but could we borrow generators' nomenclature and go with `done` here? In that case this reads clearer to me:

```js
// takeWhile
return !done && p(x) ? append(x, xs) : (done = true, xs);

// dropWhile
return !done && p(x) ? xs : (done = true, append(x, xs));
```

index.js
gabejohnson: I was trying to show an example of using a non-array. If we're not allowing strings, I can't think of a good example.

davidchambers: I think it's okay to provide array examples only.

index.js
gabejohnson: I think @cust0dian has the clearest statement using `done` and the assignment in the binary expression at the end. `done = true` is very explicit.

davidchambers: I like the use of `done`. :)

index.js
gabejohnson: This is what I get for submitting a PR while on cold medicine :sneezing_face:

test/takeWhile.js
gabejohnson: That's fine. I'll have to keep the tests using `Maybe` to maintain code coverage.

gabejohnson:
This should be ready to go.

index.js
```diff
@@ -1039,6 +1039,66 @@
...
+  //. > S.takeWhile(S.even, [3, 3, 3, 7, 6, 3, 5, 4])
+  //. []
+  //. ```
+  function takeWhile(pred, xs) {
+    if (Array.isArray(xs)) {
+      for (var idx = 0, result = [], x;
+           idx < xs.length && pred(x = xs[idx]);
+           idx += 1) result.push(x);
+      return result;
+    }
```
davidchambers: Let's move this to a separate function:

```javascript
function Array$takeWhile(pred, xs) {
  ...
}
function takeWhile(pred, xs) {
  if (Array.isArray(xs)) return Array$takeWhile(pred, xs);
  ...
}
S.takeWhile =
def(...);
```

davidchambers: Could we use `xs.slice(idx)` rather than push repeatedly?

index.js
```diff
@@ -1039,6 +1039,66 @@
       [Fn(a, $.Boolean), m(a), m(a)],
       Z.filterM);
 
+  //# takeWhile :: (Foldable f, Monoid (f a)) => (a -> Boolean) -> f a -> f a
```
davidchambers: I think we should replace `Monoid (f a)` with `Alternative f` for the reason given in sanctuary-js/sanctuary-type-classes#37.

safareli: btw here `Alternative f` constraint is used to express partiality
https://github.com/guaraqe/total-alternative#readme
https://hackage.haskell.org/package/total-alternative-0.1.0.1/docs/src/Data-List-Alternative.html#head

index.js
gabejohnson: You'd prefer `Array$takeWhile` to a generic `findIndex`? The definition of `Array$takeWhile` and `Array$dropWhile` will be nearly identical save the arguments to `slice`. We don't have to expose it.

gabejohnson: Plus we can just remove it once support for IE is dropped 😄.

gabejohnson: I took the liberty of going the `findIndex` route. I can change it if you feel strongly.

index.js
```diff
@@ -1039,6 +1039,62 @@
       [Fn(a, $.Boolean), m(a), m(a)],
       Z.filterM);
 
+  function Array$findIndex(pred, xs) {
+    for (var idx = 0; idx < xs.length && pred(xs[idx]); idx += 1);
+    return idx;
+  }
```
davidchambers: I would expect `Array$findIndex(x => x === 42, [1, 2, 3])` to evaluate to `-1` rather than to `0`. It seems this function isn't quite `Array#findIndex` after all.

gabejohnson: My mistake. Even returning `-1` it would still be incorrect. I was naming it `findIndex` but it would really have been `findLastIndex` if returning `-1` for failure.

I'll make `Array$takeWhile` and `Array$dropWhile` instead.

index.js
```diff
@@ -1039,6 +1039,68 @@
...
+  //.
+  //. ```javascript
+  //. > S.takeWhile(S.odd, [3, 3, 3, 7, 6, 3, 5, 4])
+  //. [3, 3, 3, 7]
+  //.
+  //. > S.takeWhile(S.even, [3, 3, 3, 7, 6, 3, 5, 4])
+  //. []
+  //. ```
+  function Array$takeWhile(pred, xs) {
+    for (var idx = 0; idx < xs.length && pred(xs[idx]); idx += 1);
```
davidchambers: My preference is to only declare block-scoped variables in loops (even though we don't *actually* have block scoping without `let`). Since we refer to `idx` outside the loop I'd like to declare it outside the loop:

```javascript
var idx = 0;
for (; idx < xs.length && pred(xs[idx]); idx += 1) {}
```

It would then make sense to switch to a `while` loop, as it would remove the awkward empty block:

```javascript
var idx = 0;
while (idx < xs.length && pred(xs[idx])) idx += 1;
```

davidchambers:
I left one final comment, @gabejohnson. Please also rebase your branch. Let me know once you have done so (as I don't receive a notification when you `--force` push). I'll then merge this pull request. :)

index.js
gabejohnson: The empty block is unnecessary. It can be replaced by a semicolon. But I'll make the change nonetheless.

gabejohnson:
@davidchambers It should be good to merge now.

index.js
davidchambers: I think we updated the type signature without switching from `Z.empty` to `Z.zero`. I'll rectify this in the pull request I'll soon open for sanctuary-type-classes.

davidchambers: Actually, given our `concat`-based implementation I think `Monoid (f a)` is the correct constraint after all. We could have a `chain`-based `takeWhileM` (to match `filterM`) but it seems excessive to provide two versions of a function which is only used occasionally.

davidchambers: sanctuary-js/sanctuary-type-classes#69

PR #357, opened 2017-02-28
string: add S.splitOnRegex

### LINKED ISSUE #339: Add splitOnRegex
Create a regex version of [splitOn](https://github.com/sanctuary-js/sanctuary#spliton--string---string---arraystring).

jordrake:
Closes #339 

Would appreciate some feedback please, especially on the description.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4163,7 +4163,7 @@
   //. Returns the substrings of its second argument separated by occurrences
   //. of its first argument.
   //.
-  //. See also [`joinWith`](#joinWith).
+  //. See also [`joinWith`](#joinWith) and [`splitOnRegex`](#splitOnRegex).
   //.
   //. ```javascript
   //. > S.splitOn('::', 'foo::bar::baz')
@@ -4174,6 +4174,49 @@
   }
   S.splitOn =
   def('splitOn', {}, [$.String, $.String, $.Array($.String)], splitOn);
+
+  //# splitOnRegex :: GlobalRegExp -> String -> Array String
+  //.
+  //. Takes a pattern and a string, and returns the result of splitting the
+  //. string at every non-overlapping occurrence of the pattern.
+  //.
+  //. Properties:
+  //.
+  //.   - `forall s :: forall s :: String, t :: String.
+  //.     S.joinWith(s, S.splitOnRegex(S.regex('g', S.regexEscape(s)), t)) = t
+  //.
+  //. See also [`splitOn`](#splitOn).
+  //.
+  //. ```javascript
+  //. > S.splitOnRegex(/[,;][ ]*/g, 'foo, bar, baz')
+  //. ['foo', 'bar', 'baz']
+  //.
+  //. > S.splitOnRegex(/[,;][ ]*/g, 'foo;bar;baz')
+  //. ['foo', 'bar', 'baz']
+  //. ```
+  function splitOnRegex(pattern, s) {
+    return withRegex(pattern, function() {
+      var result = [];
+      var lastIndex = 0;
+      var match;
+      while ((match = pattern.exec(s)) != null) {
+        if (pattern.lastIndex === lastIndex && match[0] === '') {
+          if (pattern.lastIndex === s.length) return result;
+          pattern.lastIndex += 1;
+        } else {
+          result.push(s.slice(lastIndex, match.index));
+          lastIndex = match.index + match[0].length;
+        }
+      }
+      result.push(s.slice(lastIndex));
+      return result;
+    });
+  }
+  S.splitOnRegex =
+  def('splitOnRegex',
+      {},
+      [$.GlobalRegExp, $.String, $.Array($.String)],
+      splitOnRegex);
 
   return S;
 
--- a/test/splitOnRegex.js
+++ b/test/splitOnRegex.js
@@ -1,0 +1,59 @@
+'use strict';
+
+var jsc = require('jsverify');
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+
+
+test('splitOnRegex', function() {
+
+  eq(typeof S.splitOnRegex, 'function');
+  eq(S.splitOnRegex.length, 2);
+  eq(S.splitOnRegex.toString(), 'splitOnRegex :: GlobalRegExp -> String -> Array String');
+
+  eq(S.splitOnRegex(/b/g, 'abc'), ['a', 'c']);
+  eq(S.splitOnRegex(/d/g, 'abc'), ['abc']);
+  eq(S.splitOnRegex(/b(?=c)/g, 'abc abd'), ['a', 'c abd']);
+  eq(S.splitOnRegex(/b(?!c)/g, 'abc abd'), ['abc a', 'd']);
+
+  eq(S.splitOnRegex(/\s+/g, 'foo  bar baz   qux'), ['foo', 'bar', 'baz', 'qux']);
+  eq(S.splitOnRegex(/\s+/g, 'foo  bar baz\tqux\nquux'), ['foo', 'bar', 'baz', 'qux', 'quux']);
+  eq(S.splitOnRegex(/\s+/g, 'foobar'), ['foobar']);
+  eq(S.splitOnRegex(/([:;])\1/g, 'foo::bar:;baz;;quux'), ['foo', 'bar:;baz', 'quux']);
+
+  eq(S.splitOnRegex(/./g, 'a\rb\nc'), ['', '\r', '\n', '']);
+  eq(S.splitOnRegex(/[^]/g, 'a\nb'), ['', '', '', '']);
+
+  eq(S.splitOnRegex(/^/g, 'foo\nbar\nbaz\n'), ['foo\nbar\nbaz\n']);
+  eq(S.splitOnRegex(/^/gm, 'foo\nbar\nbaz\n'), ['foo\n', 'bar\n', 'baz\n']);
+
+  eq(S.splitOnRegex(/foo/g, 'FOObar'), ['FOObar']);
+  eq(S.splitOnRegex(/foo/gi, 'FOObar'), ['', 'bar']);
+
+  eq(S.splitOnRegex(/(?:)/g, ''), []);
+  eq(S.splitOnRegex(/(?:)/g, 'a'), ['a']);
+  eq(S.splitOnRegex(/(?:)/g, 'ab'), ['a', 'b']);
+  eq(S.splitOnRegex(/(?:)/g, 'abc'), ['a', 'b', 'c']);
+
+  eq(S.splitOnRegex(/[^]/g, ''), ['']);
+  eq(S.splitOnRegex(/[^]/g, 'a'), ['', '']);
+  eq(S.splitOnRegex(/[^]/g, 'ab'), ['', '', '']);
+  eq(S.splitOnRegex(/[^]/g, 'abc'), ['', '', '', '']);
+
+  eq(S.splitOnRegex(/.*/g, ''), []);
+  eq(S.splitOnRegex(/.*/g, 'hello'), ['']);
+  eq(S.splitOnRegex(/./g, ''), ['']);
+  eq(S.splitOnRegex(/./g, 'hello'), ['', '', '', '', '', '']);
+
+  jsc.assert(jsc.forall(jsc.asciistring, function(t) {
+    var min = 0;
+    var max = t.length;
+    var i = jsc.random(min, max);
+    var j = jsc.random(min, max);
+    var s = t.slice(Math.min(i, j), Math.max(i, j));
+    return S.joinWith(s, S.splitOnRegex(S.regex('g', S.regexEscape(s)), t)) === t;
+  }), {tests: 1000});
+
+});


test/splitOnRegex.js
```diff
@@ -0,0 +1,21 @@
...
+  eq(typeof S.splitOnRegex, 'function');
+  eq(S.splitOnRegex.length, 2);
+  eq(S.splitOnRegex.toString(), 'splitOnRegex :: RegExp -> String -> Array String');
+
+  eq(S.splitOnRegex(/\s+/, 'foo  bar baz   qux'), ['foo', 'bar', 'baz', 'qux']);
+  eq(S.splitOnRegex(/\s+/, 'foo  bar baz\tqux\nquux'), ['foo', 'bar', 'baz', 'qux', 'quux']);
+  eq(S.splitOnRegex(/b(?=c)/, 'abcde'), ['a', 'cde']);
+  eq(S.splitOnRegex(/\s+/, 'foobar'), ['foobar']);
+  eq(S.splitOnRegex(/b/, 'abc'), ['a', 'c']);
+  eq(S.splitOnRegex(/.*/, 'hello'), ['', '']);
```
svozza: Should probably have a text for an empty string.

jordrake: Done, also added a case for no match.

test/splitOnRegex.js
```diff
@@ -0,0 +1,26 @@
...
+
+test('splitOnRegex', function() {
+
+  eq(typeof S.splitOnRegex, 'function');
+  eq(S.splitOnRegex.length, 2);
+  eq(S.splitOnRegex.toString(), 'splitOnRegex :: RegExp -> String -> Array String');
+
+  eq(S.splitOnRegex(/b/, 'abc'), ['a', 'c']);
+  eq(S.splitOnRegex(/d/, 'abc'), ['abc']);
+  eq(S.splitOnRegex(/b(?=c)/, 'abcde'), ['a', 'cde']);
```
cust0dian: This doesn't seem much different to the `/b/` regex. Maybe apply it to something like `abc abd`?

```js
 eq(S.splitOnRegex(/b(?=c)/, 'abc abd'), ['a', 'c abd']);
```

index.js
```diff
@@ -3983,6 +3983,26 @@
   S.splitOn =
   def('splitOn', {}, [$.String, $.String, $.Array($.String)], splitOn);
 
+  //# splitOnRegex :: RegExp -> String -> Array String
+  //.
+  //. Returns the substrings of its second argument separated by evaluated
+  //. matches of its first argument.
```
cust0dian: @jordrake my take at a better description would be something like:

> Takes a regex and a string and returns the substrings that do not match the regex.

but probably best to wait till @davidchambers can review the PR :)

davidchambers:
Another new contributor! Thank you for the pull request, @jordrake!

index.js
```diff
@@ -3983,6 +3983,26 @@
...
+  //.
+  //. See also [`splitOn`](#splitOn).
```
davidchambers: Let's add a reciprocal link.

index.js
```diff
@@ -3983,6 +3983,26 @@
...
+  //.
+  //. ```javascript
+  //. > S.splitOnRegex(/\s+/, 'foo  bar baz   qux')
+  //. ['foo', 'bar', 'baz', 'qux']
+  //. ```
```
davidchambers: I like to avoid examples which could better be written with some other Sanctuary function. In this case, [`words`][1] would be clearer. I'm struggling to think of a simple, vaguely realistic example which can't be written with `splitOn`.


[1]: https://sanctuary.js.org/#words

index.js
```diff
@@ -3983,6 +3983,26 @@
...
+  function splitOnRegex(pattern, s) {
+    return s.split(pattern);
+  }
+  S.splitOnRegex =
+  def('splitOnRegex',
+    {},
+    [$.RegExp, $.String, $.Array($.String)],
+    splitOnRegex);
```
davidchambers: We could reuse `splitOn` here rather than define `splitOnRegex`, since the implementations are the same.

index.js
davidchambers: I like @cust0dian's suggestion, though it could be misinterpreted. Given `/x/` and `'abc'`, the substrings that do not match the pattern are `'abc'`, `'ab'`, `'bc'`, `'a'`, `'b'`, `'c'`, and `''` (perhaps).

Another option:

_Takes a pattern and a string, and returns the result of splitting the string at every non-overlapping occurrence of the pattern (removing the matched substrings)._

cust0dian: @davidchambers would "longest substrings" be technically correct?

> Takes a regex and a string and returns the **longest** substrings that do not match the regex.

Which, after rereading [what `String#split` does](http://www.ecma-international.org/ecma-262/7.0/index.html#sec-regexp.prototype-@@split), actually not exactly the full description:

> If the regular expression contains capturing parentheses, then each time *separator* is matched the results (including any undefined results) of the capturing parentheses are spliced into the output array.

i.e. this would be surprising to me:

```js
> S.splitOnRegex(/(b|c)/, 'abcd')
['a', 'b', '', 'c', 'd']
```

index.js
cust0dian: Maybe something like more complicated `words`?

```js
> S.splitOnRegex(/\W+/, 'foo, bar (baz?), qux')
['foo', 'bar', 'baz', 'qux']
```

index.js
jordrake: I considered this especially since ramda has it in a single function as well. I just assumed that this decision had already been made because of the wording of the related issue and didn't want to step any anybody's toes with my first contribution 😀

Happy to change it to fit this.

index.js
jordrake: To me, that seems equally contrived in the opposite direction as it seems simpler to understand that by just using match:

```js
>  "foo, bar (baz?), qux".match(/(\w+)/g);
["foo", "bar", "baz", "qux"]
```

jordrake: Using StackOverflow to come up with some realistic examples people have searched for under this functionality:

- Dates (handling several seperators) http://stackoverflow.com/questions/3559883/javascript-split-regex-question
- File paths that handle backslash and forward slash.

davidchambers: These are both good suggestions. My reservation about date parsing is that if one is dealing with different separators (`[-/.]`) one is likely to also be dealing with different ordering (YYYYMMDD, DDMMYYYY, MMDDYYYY). My reservation about file paths is that the pattern is a bit noisy (`/[/\\]/`).

How about the following? It's less realistic than your suggested scenarios, but it shows the value of `splitOnRegex` over `splitOn`, and the pattern is quite readable.

```javascript
> S.splitOnRegex(/[,;][ ]*/, 'foo, bar, baz')
['foo', 'bar', 'baz']

> S.splitOnRegex(/[,;][ ]*/, 'foo;bar;baz')
['foo', 'bar', 'baz']
```

index.js
davidchambers: To be clear, I'm suggesting reusing the internal `splitOn` function as the implementation for `S.splitOnRegex`, without touching `S.splitOn`. We'd have something like this:

```javascript
function splitOn(separator, s) {
  return s.split(separator);
}
S.splitOn =
def('splitOn', {}, [$.String, $.String, $.Array($.String)], splitOn);

S.splitOnRegex =
def('splitOnRegex', {}, [$.RegExp, $.String, $.Array($.String)], splitOn);
```

index.js
davidchambers: Thank you for mentioning capturing groups, @cust0dian. They do make this function more difficult to accurately describe. Perhaps the best option is to provide an incomplete one-sentence description and a link to [MDN][1]. What do you think?


[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split

cust0dian: @davidchambers I'd say then it should be made completely clear that this is just a "safe" wrapper, i.e.

> Type-checked version of `String#split` that accepts RegEx as a separator.
>
> Takes a pattern and a string, and returns the result of splitting the string at every non-overlapping occurrence of the pattern (removing the matched substrings, depending on a pattern).
>
> [Full documentation on MDN.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split)

davidchambers: Looks good to me, Serg. :)

index.js
jordrake: Ah I see, I'm fine with that too. Out of curiosity, is there a reason behind why there are two separate functions? Are we trying to avoid polymorphic parameters? 

Thanks again for all the feedback.

davidchambers: Parametric polymorphism (e.g. `Semigroup a => a -> a -> a`) is wonderful, but we avoid ad hoc polymorphism (e.g. `(RegExp | String) -> String -> Array String`) as it's [simpler][1] to have two functions which each do one thing than to have one function which does two things.


[1]: https://www.infoq.com/presentations/Simple-Made-Easy

jordrake: Awesome. I've not spent much time in statically typed languages and never in typed functional programming languages so these are interesting topics for me.

I appreciate the patience and will action these comments tomorrow.

jordrake:
I think I've addressed all the comments now 😃

index.js
```diff
@@ -3983,6 +3983,31 @@
   S.splitOn =
   def('splitOn', {}, [$.String, $.String, $.Array($.String)], splitOn);
 
+  //# splitOnRegex :: RegExp -> String -> Array String
+  //.
+  //. Type-checked version of String#split that accepts RegEx as a separator.
```
davidchambers: I would like to make `String#split` itself a link to the MDN documentation:

    Type-checked version of the regular expression mode of [`String#split`][].

We'd need to include a reference for ``[`String#split`]`` at the bottom of the document.

index.js
```diff
@@ -3983,6 +3983,31 @@
...
+  //.
+  //. Takes a pattern and a string, and returns the result of splitting the
+  //. string at every non-overlapping occurrence of the pattern (possibly
+  //. removing the matched substrings, depending on a pattern).
```
davidchambers: > possibly removing the matched substrings

It's simpler to state that the function *always* replaces each match with a (possibly empty) splice containing the captured substrings. If the pattern is `/(-)/`, say, it will *appear* that the matches are not removed:

```javascript
> 'foo-bar-baz'.split(/(-)/)
['foo', '-', 'bar', '-', 'baz']
```

We just happen to be replacing `'-'` with the splice `['-']`.

Changing the number of capturing groups is instructive:

```javascript
> 'foo-bar-baz'.split(/-/)
['foo', 'bar', 'baz']

> 'foo-bar-baz'.split(/(((-)))/)
['foo', '-', '-', '-', 'bar', '-', '-', '-', 'baz']
```

I think we should remove the parenthesized clause and rely on MDN to describe the treatment of capturing groups.

index.js
```diff
@@ -3983,6 +3983,31 @@
...
+  //. ['foo', 'bar', 'baz']
+  //.
+  //. > S.splitOnRegex(/[,;][ ]*/, 'foo;bar;baz')
+  //. ['foo', 'bar', 'baz']
+  //. ```
+  S.splitOnRegex =
+  def('splitOnRegex',
+    {},
+    [$.RegExp, $.String, $.Array($.String)],
+    splitOn);
```
davidchambers: <del>Please indent the three lines above so that all four arguments to `def` are aligned.</del>

<ins>Please place all four arguments to `def` on the same line (line wrapping is not necessary in this case).</ins>

jordrake: Ah yeah, changing from splitOnRegex to splitOn will have brought me just under the linting rule.

davidchambers:
@jordrake, let me know if there's anything I can do to help here. I'm happy to make the final few changes myself if you're busy with other things.

jordrake:
@davidchambers Incredibly sorry about the delay, life got in the way. I should be able to pick this up in next few days.

davidchambers:
No worries, Jordan. Life tends to do that.

I look forward to seeing an update, but we're under no time pressure whatsoever. :)

jordrake:
Made changes as suggested in #339 (shamelessly copy and pasted because if it ain't broke...). 
Areas I'd like to highlight for comment include the new description, unit tests and the property tests.

index.js
```diff
@@ -3983,6 +3983,47 @@
   S.splitOn =
   def('splitOn', {}, [$.String, $.String, $.Array($.String)], splitOn);
 
+
```
davidchambers: Nit: Please remove this line so that there's just a single empty line between functions.

index.js
```diff
@@ -3983,6 +3983,47 @@
...
+  //# splitOnRegex :: RegExp -> String -> Array String
+  //.
+  //. Takes a pattern and a string, and returns the result of splitting the
+  //. string at every non-overlapping occurrence of the pattern
```
davidchambers: Nit: Please add a full stop (`.`) at the end of the sentence.

test/splitOnRegex.js
```diff
@@ -0,0 +1,52 @@
...
+  eq(S.splitOnRegex(/.*/, 'hello'), ['']);
+  eq(S.splitOnRegex(/./, ''), ['']);
+  eq(S.splitOnRegex(/./, 'hello'), ['', '', '', '', '', '']);
+  
+
+  jsc.assert(jsc.forall(jsc.nestring, function(s) {
+    return equals(S.splitOnRegex(S.regex('', S.regexEscape(s)), s))(['', ''])
+        && equals(S.splitOnRegex(/[^]/, s).length)(s.length + 1)
+        && equals(S.splitOnRegex(/(?:)/, s))(S.splitOn('', s));
+  }), {tests: 1000});
```
davidchambers: I imagined a test for this property:

  - `forall s :: String, t :: String. S.joinWith(s, S.splitOnRegex(S.regex('', S.regexEscape(s)), t)) = t`

Ideally the algorithm for choosing values for `s` and `t` would result in `s` being a substring of `t` in a large number of cases, with `s` appearing multiple times in `t` in at least a handful of cases. See __test/splitOn.js__ for inspiration. :)

test/splitOnRegex.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+
+var jsc = require('jsverify');
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+var equals = require('./internal/equals');
+
+test('splitOnRegex', function() {
```
davidchambers: Nit: Two empty lines after the final import, please. :)

jordrake:
Nits picked and added that property test.

test/splitOnRegex.js
```diff
@@ -0,0 +1,61 @@
...
+  eq(S.splitOnRegex(/^.*$/, 'foo\nbar'), ['foo\nbar']);
+
+  eq(S.splitOnRegex(/foo/, 'FOObar'), ['FOObar']);
+  eq(S.splitOnRegex(/foo/i, 'FOObar'), ['', 'bar']);
+
+  eq(S.splitOnRegex(/b/, 'abc'), ['a', 'c']);
+  eq(S.splitOnRegex(/(?:b)/, 'abc'), ['a', 'c']);
+
+  eq(S.splitOnRegex(/(?:)/, ''), []);
+  eq(S.splitOnRegex(/(?:)/, 'hello'), ['h', 'e', 'l', 'l', 'o']);
```
davidchambers: In place of these two unit tests and the `S.splitOnRegex(/(?:)/, s) = S.splitOn('', s)` property test, I think a few unit tests would suffice:

```javascript
eq(S.splitOnRegex(/(?:)/, ''), []);
eq(S.splitOnRegex(/(?:)/, 'a'), ['a']);
eq(S.splitOnRegex(/(?:)/, 'ab'), ['a', 'b']);
eq(S.splitOnRegex(/(?:)/, 'abc'), ['a', 'b', 'c']);
```

test/splitOnRegex.js
```diff
@@ -0,0 +1,61 @@
...
+  }), {tests: 1000});
+
+  jsc.assert(jsc.forall(jsc.asciistring, function(t) {
+    var min = 0;
+    var max = t.length;
+    var i = jsc.random(min, max);
+    var j = jsc.random(min, max);
+    var s = t.slice(Math.min(i, j), Math.max(i, j));
+    return S.joinWith(s, S.splitOnRegex(S.regex('', S.regexEscape(s)), t)) === t;
+  }), {tests: 1000});
```
davidchambers: Nice! Let's also add the property to the description using the same format we use for `joinWith`. :)

test/splitOnRegex.js
davidchambers: Unit tests would suffice for the `/[^]/` behaviour too:

```javascript
eq(S.splitOnRegex(/[^]/, ''), ['']);
eq(S.splitOnRegex(/[^]/, 'a'), ['', '']);
eq(S.splitOnRegex(/[^]/, 'ab'), ['', '', '']);
eq(S.splitOnRegex(/[^]/, 'abc'), ['', '', '', '']);
```

test/splitOnRegex.js
```diff
@@ -0,0 +1,61 @@
...
+  eq(S.splitOnRegex(/\s+/, 'foo  bar baz   qux'), ['foo', 'bar', 'baz', 'qux']);
+  eq(S.splitOnRegex(/\s+/, 'foo  bar baz\tqux\nquux'), ['foo', 'bar', 'baz', 'qux', 'quux']);
+  eq(S.splitOnRegex(/\s+/, 'foobar'), ['foobar']);
+  eq(S.splitOnRegex(/;(;)?/, 'foo;;bar;baz'), ['foo', 'bar', 'baz']);
+
+  eq(S.splitOnRegex(/./, 'a\rb\nc'), ['', '\r', '\n', '']);
+  eq(S.splitOnRegex(/[^]/, 'a\nb'), ['', '', '', '']);
+
+  eq(S.splitOnRegex(/^.*$/m, 'foo\nbar'), ['', '\n', '']);
+  eq(S.splitOnRegex(/^.*$/, 'foo\nbar'), ['foo\nbar']);
```
davidchambers: I think these examples would be a nice test of `m` functionality:

```javascript
eq(S.splitOnRegex(/^/, 'foo\nbar\nbaz\n'), ['foo\nbar\nbaz\n']);
eq(S.splitOnRegex(/^/m, 'foo\nbar\nbaz\n'), ['foo\n', 'bar\n', 'baz\n']);
```

`/^/m` demonstrates a use case: splitting a string into lines without losing the trailing `\n` characters.

test/splitOnRegex.js
```diff
@@ -0,0 +1,61 @@
...
+
+  eq(S.splitOnRegex(/foo/, 'FOObar'), ['FOObar']);
+  eq(S.splitOnRegex(/foo/i, 'FOObar'), ['', 'bar']);
+
+  eq(S.splitOnRegex(/b/, 'abc'), ['a', 'c']);
+  eq(S.splitOnRegex(/(?:b)/, 'abc'), ['a', 'c']);
```
davidchambers: I'm not sure what this is testing, since `/b/` and `/(?:b)/` should be interchangeable. Unless I'm overlooking something, we could remove these two test cases.

test/splitOnRegex.js
```diff
@@ -0,0 +1,61 @@
...
+
+  eq(S.splitOnRegex(/b/, 'abc'), ['a', 'c']);
+  eq(S.splitOnRegex(/d/, 'abc'), ['abc']);
+  eq(S.splitOnRegex(/b(?=c)/, 'abc abd'), ['a', 'c abd']);
+  eq(S.splitOnRegex(/b(?!c)/, 'abc abd'), ['abc a', 'd']);
+
+  eq(S.splitOnRegex(/\s+/, 'foo  bar baz   qux'), ['foo', 'bar', 'baz', 'qux']);
+  eq(S.splitOnRegex(/\s+/, 'foo  bar baz\tqux\nquux'), ['foo', 'bar', 'baz', 'qux', 'quux']);
+  eq(S.splitOnRegex(/\s+/, 'foobar'), ['foobar']);
+  eq(S.splitOnRegex(/;(;)?/, 'foo;;bar;baz'), ['foo', 'bar', 'baz']);
```
davidchambers: I like the fact that we're testing a capturing group, but it bothers me slightly that a capturing group is not required in this case. I'm struggling to think of a good example, but we could use something like this:

```javascript
eq(S.splitOnRegex(/([:;])\1/, 'foo::bar:;baz;;quux'), ['foo', 'bar:;baz', 'quux']);
```

jordrake:
Updated :)

davidchambers:
Thank you for persevering with this pull request, Jordan! I'm sorry that an issue labelled [good first contribution][1] was far from straightforward.

There are two things I would love you to do before we merge this pull request:

  - `git rebase master`; and
  - make the change I'll suggest shortly (in a separate comment).


[1]: https://github.com/sanctuary-js/sanctuary/labels/good%20first%20contribution

index.js
```diff
@@ -3983,6 +3983,51 @@
...
+  //. ['foo', 'bar', 'baz']
+  //.
+  //. > S.splitOnRegex(/[,;][ ]*/, 'foo;bar;baz')
+  //. ['foo', 'bar', 'baz']
+  //. ```
+  function splitOnRegex(pattern, s) {
+    var $pattern = regex(
+      'g' + (pattern.ignoreCase ? 'i' : '') + (pattern.multiline ? 'm' : ''),
+      pattern.source
+    );
```
davidchambers: It occurred to me that it would be semantically accurate for the type of the first argument to be `GlobalRegExp` rather than simply `RegExp`. We are, after all, "splitting […] at *every* non-overlapping occurrence of the pattern".

If we make this change we'll be able to use `withRegex` to avoid cloning `pattern`:

```javascript
return withRegex(pattern, function() {
  var result = [];
  ...
  return result;
});
```

jordrake:
Done both those actions. I appreciate the patience and feedback and even though a lot of this was copying from comments/feedback because of the complexity I learnt a lot.

index.js
```diff
@@ -4175,6 +4175,49 @@
...
   def('splitOn', {}, [$.String, $.String, $.Array($.String)], splitOn);
 
+  //# splitOnRegex :: GlobalRegExp -> String -> Array String
+  //.
+  //. Takes a pattern and a string, and returns the result of splitting the
+  //. string at every non-overlapping occurrence of the pattern.
+  //.
+  //. Properties:
+  //.
+  //.   - `forall s :: forall s :: String, t :: String.
```
Avaq: Is this `forall s :: forall s` intentional, and if so, what does it mean?

davidchambers: Good :eyes:! Fixed in #400.

davidchambers: The line below is missing a closing `` ` ``. Fixed in #405.

PR #359, opened 2017-03-05
maybe, either: define "fantasy-land/concat" conditionally

### LINKED ISSUE #346: Type-class constraints on applied types not handled
I'm not sure if "applied types" is the right term, so I'll just convey the problem by example:

```js
> S.concat(S.Nothing, S.Just(1))
Just(1)

> S.concat(S.Just(1), S.Just(1))
TypeError: Semigroup.methods.concat(...) is not a function
```

This happens because `sanctuary-type-classes` does not have access to the information that `Maybe Number` is not a Semigroup. It only knows that `Maybe` implements `concat`, and therefore must be a Semigroup. The fact that `Maybe#concat` dispatches to `concat` on the inner-type is currently an implementation detail.

davidchambers:
Closes #346

Before:

```javascript
S.concat(S.Just(1), S.Just(1));
// ! TypeError: Semigroup.methods.concat(...) is not a function
```

After:

```javascript
S.concat(S.Just(1), S.Just(1));
//
//   concat :: Semigroup a => a -> a -> a
//             ^^^^^^^^^^^    ^
//                            1
//
//   1)  Just(1) :: Maybe Number, Maybe FiniteNumber, Maybe NonZeroFiniteNumber, Maybe Integer, Maybe ValidNumber
//
//   ‘concat’ requires ‘a’ to satisfy the Semigroup type-class constraint; the value at position 1 does not.
```

Losing the ability to `filter(odd, Just(4))` is a real shame. This happens because `Just(4)` does not satisfy the requirements of Monoid as it does not satisfy the requirements of Semigroup.

It seems we need [`mfilter`][1]. Fantasy Land does not yet specify MonadPlus. We could define `mfilter` and hard-code support for the Maybe type, while also pushing to have MonadPlus added to Fantasy Land. With luck we would later be able to move the implementation from `mfilter` to `Maybe#fantasy-land/mplus`. Is this reasonable?

[1]: http://hackage.haskell.org/package/base-4.9.1.0/docs/Control-Monad.html#v:mfilter

### DIFF
--- a/index.js
+++ b/index.js
@@ -1021,12 +1021,6 @@
   //. ```javascript
   //. > S.filter(S.odd, [1, 2, 3, 4, 5])
   //. [1, 3, 5]
-  //.
-  //. > S.filter(S.odd, S.Just(9))
-  //. Just(9)
-  //.
-  //. > S.filter(S.odd, S.Just(4))
-  //. Nothing
   //. ```
   S.filter =
   def('filter',
@@ -1400,6 +1394,12 @@
     this.isNothing = tag === 'Nothing';
     this.isJust = tag === 'Just';
     if (this.isJust) this.value = value;
+
+    //  Add "fantasy-land/concat" method conditionally so that Just('abc')
+    //  satisfies the requirements of Semigroup but Just(123) does not.
+    if (this.isNothing || Z.Semigroup.test(this.value)) {
+      this['fantasy-land/concat'] = Maybe$prototype$concat;
+    }
   }
 
   //# Nothing :: Maybe a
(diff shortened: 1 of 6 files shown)


safareli:
MonadPlus is just Monad + Alternative with some extra laws(on which there still is [debate](https://wiki.haskell.org/MonadPlus_reform_proposal)), and even if it would be added to FL one couldn't distinguish between some `(Apply m, Chain m, Plus m) => m` and `(MonadPlus m) => m` as it's not adding any new methods.

so you can write:
```js
// :: (Alternative m, Monad m) => (a -> Boolean, m a) -> m a
// -- or:
// :: (Plus m, Chain m, Apply m) => (a -> Boolean, m a) -> m a
// --  but I think first one is much better
const mfilter =  (p,ma) => {
 const T = ma.constructor
 return ma.chain(a => p(a) ? T.of(a) : T.zero()
}
```

davidchambers:
Terrific suggestion, @safareli! The Sanctuary projects—sanctuary-type-classes in particular—would not be where they are today without your involvement. Thank you!

I just opened sanctuary-js/sanctuary-type-classes#37. Have a look. :)

safareli:
@davidchambers Welcome! ;)

davidchambers:
I rebased on #370 to confirm that values of type `Maybe a` will remain compatible with `filterM`. :)

Avaq:
LGTM - Seems tests experience IE10 iFrame resource error again. Probably some race condition.

PR #363, opened 2017-03-18
lint: use a Markdown parser to check link definitions

davidchambers:
This cryptic six-line Unix pipeline is present in many @sanctuary-js makefiles, and will soon find its way into sanctuary-either and sanctuary-maybe if we don't replace it.

There are several advantages to this replacement:

  - the code in the makefile is self descriptive;
  - the error output includes a list of failures and their locations; and
  - the code does not rely on Unix programs so should behave consistently on all platforms.

The disadvantage is that we have three more npm dependencies. At least I managed to avoid introducing another __.\*rc__ file!

I'd appreciate a review. Once someone has signed off on this change I'll merge this pull request and open similar pull requests for other @sanctuary-js projects.


Makefile
```diff
@@ -2,6 +2,7 @@ DOCTEST = node_modules/.bin/doctest --module commonjs --prefix .
 ESLINT = node_modules/.bin/eslint --config node_modules/sanctuary-style/eslint-es3.json --env es3
 ISTANBUL = node_modules/.bin/istanbul
 NPM = npm
+REMARK = node_modules/.bin/remark --frail --no-stdout
```
davidchambers:
```console
$ node_modules/.bin/remark --help 2>&1 | grep -- frail
  -f  --frail               exit with 1 on warnings

$ node_modules/.bin/remark --help 2>&1 | grep -- stdout
      --[no-]stdout         specify writing to stdout (on by default)
```

Makefile
```diff
@@ -47,13 +48,13 @@ lint:
...
-	      -e 's:\[.*\]\[\(.*\)\]:\1:' \
-	      -e '/0-9/d' \
-	| xargs -I '{}' sh -c "grep '^//[.] \[{}\]: ' index.js"
+	rm -f README.md
+	VERSION=0.0.0 make README.md
+	$(REMARK) \
+	  --use remark-lint-no-undefined-references \
+	  --use remark-lint-no-unused-definitions \
+	  -- README.md
+	git checkout README.md
```
davidchambers: The new approach is no shorter than the old approach, even though it contains no logic. :stuck_out_tongue_winking_eye:

Makefile
```diff
@@ -47,13 +48,13 @@ lint:
...
 	$(REMEMBER_BOWER) $(shell pwd)
-	@echo 'Checking for missing link definitions...'
-	grep -o '\[[^]]*\]\[[^]]*\]' index.js \
-	| sort -u \
-	| sed -e 's:\[\(.*\)\]\[\]:\1:' \
-	      -e 's:\[.*\]\[\(.*\)\]:\1:' \
-	      -e '/0-9/d' \
-	| xargs -I '{}' sh -c "grep '^//[.] \[{}\]: ' index.js"
+	rm -f README.md
+	VERSION=0.0.0 make README.md
```
svozza: Why version `0.0.0`?

davidchambers: We need to generate __README.md__ in order to check the link definitions in __index.js__. `VERSION` is not significant for this purpose, but I like to specify it as a reminder that it *is* required in order to generate a "valid" readme.

svozza: Cool

Issue #353, opened 2017-02-23
Add S.sort and S.sortBy


svozza:
Add Sanctuary version of [R.sort](http://ramdajs.com/0.23.0/docs/#sort) and [R.sortBy](http://ramdajs.com/0.23.0/docs/#sort).

davidchambers:
I think we should borrow heavily from Haskell here.

> <code>__sortBy__ :: (a -> a -> [Ordering][]) -> [a] -> [a]</code>
>
> The [`sortBy`][] function is the non-overloaded version of [`sort`][].

> <code>__sort__ :: [Ord][] a => [a] -> [a]</code>
>
> The [`sort`][] function implements a stable sorting algorithm. It is a special case of [`sortBy`][], which allows the programmer to supply their own comparison function.

Let's start with `sortBy`, which doesn't require a type class. We'll need to define `Ordering` via sanctuary-def. It will look something like this:

```javascript
//# Ordering :: Type
var Ordering = $.EnumType(
  'sanctuary/Ordering',
  readmeUrl('Ordering'),
  ['LT', 'EQ', 'GT']
);
```


[Ord]: http://hackage.haskell.org/package/base-4.9.1.0/docs/Data-Ord.html#t:Ord
[Ordering]: http://hackage.haskell.org/package/base-4.9.1.0/docs/Data-Ord.html#t:Ordering
[`sortBy`]: http://hackage.haskell.org/package/base-4.9.1.0/docs/Data-List.html#v:sortBy
[`sort`]: http://hackage.haskell.org/package/base-4.9.1.0/docs/Data-List.html#v:sort

sadasant:
@davidchambers thank you! I will make progress between today and tomorrow.

sadasant:
@davidchambers

So, what about `def('sortBy', {f: [Z.Applicative, Z.Foldable, Z.Monoid]}, [Fn(a, Fn(a, Ordering)), f(a), f(a)], sortBy);` ?

Then we could use something like this: `sortBy cmp = foldr (insertBy cmp) []`, or should we rather implement the better/harder merge sort that's over here? http://hackage.haskell.org/package/base-4.9.1.0/docs/src/Data.OldList.html#sortBy

Little question: Do we have foldr in sanctuary-js?

sadasant:
sortBy for review: https://github.com/sanctuary-js/sanctuary/pull/365

davidchambers:
[...]

PR #364, opened 2017-03-23
S.sort as in R.sort

sadasant:
For: https://github.com/sanctuary-js/sanctuary/issues/353

Please review \o/

### DIFF
--- a/index.js
+++ b/index.js
@@ -4030,6 +4030,25 @@
   S.splitOn =
   def('splitOn', {}, [$.String, $.String, $.Array($.String)], splitOn);
 
+  //# sort :: ((a, a) -> Number) -> Array a -> Array a
+  //.
+  //. Returns a copy of the list, sorted according to the comparator
+  //. function, which should accept two values at a time and return a
+  //. negative number if the first value is smaller, a positive number if
+  //. it's larger, and zero if they are equal. Please note that this is a
+  //. **copy** of the list. It does not modify the original.
+  //.
+  //. ```javascript
+  //. > S.sort(S.sub, [4, 2, 7, 5])
+  //. [2, 4, 5, 7]
+  //. ```
+  function sort(f, a) {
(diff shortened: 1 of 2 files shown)


davidchambers:
I left a comment in #353. Let's start with `sortBy` before deciding how `sort` should work.

index.js
```diff
@@ -4030,6 +4030,25 @@
...
+  //. negative number if the first value is smaller, a positive number if
+  //. it's larger, and zero if they are equal. Please note that this is a
+  //. **copy** of the list. It does not modify the original.
+  //.
+  //. ```javascript
+  //. > S.sort(S.sub, [4, 2, 7, 5])
+  //. [2, 4, 5, 7]
+  //. ```
+  function sort(f, a) {
+    return a.sort(f);
```
davidchambers: This mutates `a`. :scream:

Avaq: :scream_cat:

sadasant: It's alright! A new version is coming soon 👀💧

sadasant: I have a comment here: https://github.com/sanctuary-js/sanctuary/issues/353#issuecomment-288889085


So, what about `def('sortBy', {f: [Z.Applicative, Z.Foldable, Z.Monoid]}, [Fn(a, Fn(a, Ordering)), f(a), f(a)], sortBy);` ?

Then we could use something like this: `sortBy cmp = foldr (insertBy cmp) []`, or should we rather implement the better/harder merge sort that's over here? http://hackage.haskell.org/package/base-4.9.1.0/docs/src/Data.OldList.html#sortBy

Little question: Do we have foldr in sanctuary-js?

sadasant: sortBy for review: https://github.com/sanctuary-js/sanctuary/pull/365

davidchambers:
Let's add `sortBy` (#365) then revisit `sort`. :)

PR #365, opened 2017-03-24
sortBy for Arrays

sadasant:
Work in progress! Please review!


index.js
```diff
@@ -4030,6 +4037,83 @@
...
+  //. [2, 4, 5, 7]
+  //. ```
+  function sortBy(cmp, xs) {
+    // TODO: Should we separate foldr?
+    // sortBy cmp = foldr (insertBy cmp) []
+    if (xs.length <= 1) {
+        return xs;
+    }
+    var x = S.head(xs).value;
+    var xs_ = S.tail(xs).value;
```
davidchambers: `head` and `tail` can only operate on arrays (or strings). We cannot use these functions if we wish `sortBy` to be as general as its type signature suggests it is.

index.js
```diff
@@ -4030,6 +4037,83 @@
...
+  //.
+  //. ```javascript
+  //. > S.insertBy(S.compare, 6, [4, 2, 7, 5])
+  //. [4, 2, 6, 7, 5]
+  //. ```
+  function insertBy(cmp, x, ys) {
+    if (ys.length === 0) {
+        return S.prepend(x, ys);
+    }
+    var y = S.head(ys).value;
```
svozza: I don't think there's any need to use `head`, just access the index directly

index.js
```diff
@@ -4030,6 +4037,83 @@
...
+    var ys_ = S.tail(ys).value;
+    var result = cmp(x)(y);
+    if (result === 'GT') {
+        return S.prepend(y, insertBy(cmp, x, ys_));
```
svozza: Will this not just fail on larger lists because it's recursive?

sadasant: Yes, definitely. I'll make a non-recursive one today.

index.js
sadasant: Do we want it to be as general as the signature? If so, how should this be?
Otherwise, using slice is valid?

index.js
sadasant: Gotcha, waiting for feedback here before I go in direct index direction: https://github.com/sanctuary-js/sanctuary/pull/365#discussion_r107914151

index.js
safareli: It will.

davidchambers:
I suggest starting with something quite straightforward:

```javascript
//# sortBy :: (a -> a -> Ordering) -> Array a -> Array a
function sortBy(cmp) {
  return function(xs) {
    return xs.slice().sort(function(a, b) {
      var ord = cmp(a)(b);
      return ord === 'LT' ? -1 : ord === 'GT' ? 1 : 0;
    });
  };
}
```

```javascript
var LT = 'LT';
var EQ = 'EQ';
var GT = 'GT';

function lower(a) {
  return function(b) {
    var aa = a.toLowerCase();
    var bb = b.toLowerCase();
    return aa < bb ? LT : aa > bb ? GT : a < b ? LT : a > b ? GT : EQ;
  };
}

sortBy(lower)(['alice', 'Carin', 'bob', 'Bob']);
// => ['alice', 'Bob', 'bob', 'Carin']
```

sadasant:
@davidchambers doing it in some minutes.

sadasant:
@davidchambers I can submit insertBy separately, and foldr, if they are valuable. Are they?

index.js
gabejohnson: It either has to be as general as the signature or the signature has to change.

safareli: Try to sort linked List for example and this function will fail. Also for large list/array it will fail too.

index.js
gabejohnson: What should `insertBy` when `ys` is a `Maybe`?

sadasant: I'm not sure 😂

sadasant:
@davidchambers changes here: https://github.com/sanctuary-js/sanctuary/pull/365/commits/a625ef46b979a65b024769ed8066fcc188fca574

feedback welcome \o/ more like: please feedback 😂 

I was trying to make it a List instead of an Array, to sort strings, but then I thought it might not be wanted, so I left it for Arrays. Also, I didn't use the curried method you used, should I have?

Thanks for the time! (everyone)

sadasant:
Oh, I can also clean the git history after we polish this \o/

sadasant:
Any clue on what's `IE 10.0.0 (Windows 8 0.0.0)  get FAILED` ?

index.js
```diff
@@ -4030,6 +4037,39 @@
   S.splitOn =
   def('splitOn', {}, [$.String, $.String, $.Array($.String)], splitOn);
 
+  //# sortBy :: (a -> a -> Ordering) -> Array a -> Array a
+  //.
+  //. Returns a copy of the list, sorted according to the comparator
+  //. function, which should accept two values at a time and return a
+  //. negative number if the first value is smaller, a positive number if
+  //. it's larger, and zero if they are equal. Please note that this is a
+  //. **copy** of the list. It does not modify the original.
```
davidchambers: Let's make some changes to this description:

  - replace occurrences of "list" with "array";
  - replace mentions of negative numbers, zero, and positive numbers; and
  - remove mentions of copying (it should go without saying that Sanctuary will not mutate arguments).

sadasant: Thanks!

index.js
```diff
@@ -345,6 +345,13 @@
     function(typeRep) { return []; }
   );
 
+  //  Ordering :: Type
+  var Ordering = $.EnumType(
+    'sanctuary/Ordering',
+    readmeUrl('Ordering'),
+    ['LT', 'EQ', 'GT']
+  );
```
gabejohnson: Should `Ordering` be exposed to the user?

Pros:
1. Ability to change the values underlying the enumeration in the future without breaking.
2. Documentation of `Ordering`.

sadasant: Got it, doing it! :)

sadasant: Actually, before doing so, anyone else agrees?
@davidchambers @svozza

index.js
sadasant: Done! (ameded) https://github.com/sanctuary-js/sanctuary/pull/365/commits/acba0cfdb4160b400cd2dd8bcd92446001ed38dd#diff-168726dbe96b3ce427e7fedce31bb0bcR4048

index.js
davidchambers: I'm certainly in favour of documenting `Ordering`.

If we're to export `LT`, `EQ`, and `GT` it's probably a good idea to make their values `-1`, `0`, and `1` respectively for more convenient interoperability with `Array#sort`.

index.js
```diff
@@ -4030,6 +4037,41 @@
...
   def('splitOn', {}, [$.String, $.String, $.Array($.String)], splitOn);
 
+  //# sortBy :: (a -> a -> Ordering) -> Array a -> Array a
+  //.
+  //. Returns a copy of the array, sorted according to the comparator
+  //. function. The comparator function is expected to be curried,
+  //. and will receive two values at a time. The comparator must return
+  //. a value of type Ordering, which can be any of the following strings:
+  //. - 'LT': when the first value received goes **before** the second value.
+  //. - 'GT': when the first value received goes **after** the second value.
```
davidchambers: Rather than describe the behaviour in terms of before/after, I think it's sufficient to use less/greater than. For example, _the comparator should return `'LT'` if the first value is less than the second value_.

index.js
```diff
@@ -4030,6 +4037,41 @@
...
+  //. - 'GT': when both values should remain unchanged in the Array.
```
davidchambers: `s/GT/EQ/`

sadasant: *facepalms*, omw

index.js
```diff
@@ -4030,6 +4037,41 @@
...
+  //. ['alice', 'Bob', 'bob', 'Carin']
+  //. ```
+  function sortBy(cmp, xs) {
+    return xs.slice().sort(function(a, b) {
+      var ord = cmp(a)(b);
+      return ord === 'LT' ? -1 : ord === 'GT' ? 1 : 0;
+    });
+  }
+  S.sortBy = def('sortBy', {},
+    [Fn(a, Fn(a, Ordering)), $.Array(a), $.Array(a)], sortBy);
```
davidchambers: Let's go with this line wrapping:

```javascript
S.sortBy =
def('sortBy', {}, [Fn(a, Fn(a, Ordering)), $.Array(a), $.Array(a)], sortBy);
```

index.js
sadasant: Ordering values set to -1, 0 and 1: https://github.com/sanctuary-js/sanctuary/pull/365/files#diff-168726dbe96b3ce427e7fedce31bb0bcR352

Things look a lot better now!

index.js
```diff
@@ -4030,6 +4037,29 @@
...
+  //.
+  //. ```javascript
+  //. > function increasingOrder(a) { return function(b) { return a - b; }; }
+  //.
+  //. > S.sortBy(increasingOrder)([4, 2, 7, 5]);
+  //. [2, 4, 5, 7]
+  //. ```
+  function sortBy(cmp, xs) {
+    return xs.slice().sort(function(a, b) {
+      return cmp(a)(b);
```
sadasant: Any idea on how to make the comparison function optionally curried? as in `cmp(a, b)` and `cmp(a)(b)`?

index.js
```diff
@@ -4030,6 +4037,29 @@
...
+    });
```
davidchambers: Even nicer:

```javascript
return xs.slice().sort(uncurry2(cmp));
```

index.js
```diff
@@ -4030,6 +4037,29 @@
...
+  //.
+  //. Returns a copy of the array, sorted according to the comparator
+  //. function. The comparator function is expected to be curried,
+  //. and will receive two values at a time. The comparator must return:
+  //. - 'LT': if the first value is **less** than the second value.
+  //. - 'GT': if the first value is **greater** than the second value.
+  //. - 'EQ': if the first value is **equal** to the second value.
+  //.
+  //. ```javascript
+  //. > function increasingOrder(a) { return function(b) { return a - b; }; }
```
davidchambers: This comparator isn't valid, as it can return values other than `-1`, `0`, and `1`. It's not being caught by the type checker because sanctuary-def doesn't yet handle functions which take *curried* functions.

sadasant: Thanks for noticing!

index.js
davidchambers: Let's not introduce magic. If we want to support `((a, a) -> Ordering) -> Array a -> Array a` as well we should provide a separate function named `sortBy_`. The `_` suffix is used for this purpose (see, for example, [`reduce_`][1]).


[1]: https://sanctuary.js.org/#reduce_

davidchambers:
I've come up with an implementation based on #388:

```javascript
//# sortBy :: (Ord b, Applicative f, Foldable f, Monoid (f a)) => (a -> b) -> f a -> f a
function sortBy(f, foldable) {
  return Z.reduce(
    function(foldable, x) { return append(x, foldable); },
    Z.empty(foldable.constructor),
    Z.reduce(function(xs, x) {
      var fx = f(x);
      var idx = 0;
      for (; idx < xs.length; idx += 1) if (Z.lte(fx, f(xs[idx]))) break;
      xs.splice(idx, 0, x);
      return xs;
    }, [], foldable)
  );
}
S.sortBy =
def('sortBy',
    {b: [Z.Ord], f: [Z.Applicative, Z.Foldable, Z.Monoid]},
    [$.Function([a, b]), f(a), f(a)],
    sortBy);
```

I'd love to know what you think of this, @safareli. Note that I'm using an array as a mutable intermediate data structure. Is there a more elegant approach?

Avaq:
@davidchambers I remember recently coming across [this article](https://elvishjerricco.github.io/2017/03/23/applicative-sorting.html), where somebody explores the possibility to sort Traversables without the need for an intermediate list. In the mean time, somebody took the idea and made it into a [Haskell library](https://github.com/treeowl/sort-traversable). Maybe we can take inspiration from it?

davidchambers:
That article is interesting, @Avaq. I don't know how useful it will be in this case as it looks as though we need to define the Const type and a few other building blocks. I'd love to see an implementation!

Avaq:
> it looks as though we need to define the Const type

Yeah, plus this inexplicable `Mono` type. I would also enjoy seeing a JavaScript implementation, but I will agree that replacing the already available Array with two other types we first have to implement at little other gain than the elusive "elegance" may not really be worth it from a practical perspective. ;)

safareli:
One thing I spot in the implementation is that you call `f(xs[idx])` in loop.

davidchambers:
I'm sure the implementation I proposed could be improved. Please have a go if you'd like to, Irakli.

safareli:
``` diff
//# sortBy :: (Ord b, Applicative f, Foldable f, Monoid (f a)) => (a -> b) -> f a -> f a
function sortBy(f, foldable) {
  return Z.reduce(
-    function(foldable, x) { return append(x, foldable); },
+    function(foldable, x) { return append(x.val, foldable); },
    Z.empty(foldable.constructor),
    Z.reduce(function(xs, x) {
      var fx = f(x);
      var idx = 0;
-      for (; idx < xs.length; idx += 1) if (Z.lte(fx, f(xs[idx]))) break;
+      for (; idx < xs.length; idx += 1) if (Z.lte(fx, xs[idx].tag)) break;
-      xs.splice(idx, 0, x);
+      xs.splice(idx, 0, { val : x, tag : fx });
      return xs;
    }, [], foldable)
  );
}
S.sortBy =
def('sortBy',
    {b: [Z.Ord], f: [Z.Applicative, Z.Foldable, Z.Monoid]},
    [$.Function([a, b]), f(a), f(a)],
    sortBy);
```

davidchambers:
Very nice, Irakli!

davidchambers:
Superseded by #397

PR #367, opened 2017-04-02
update description to reflect the broader goal

davidchambers:
Sanctuary began life as a small library to accompany Ramda, providing safe replacements for unsafe Ramda functions such as `R.head :: Array a -> a?`.

Today, Sanctuary is a significantly larger and more ambitious project. It seems appropriate to update the description to reflect this.

_Refuge from <del>unsafe</del> <ins>unprincipled</ins> JavaScript_ – :thumbsup: / :thumbsdown: ?

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "sanctuary",
-  "description": "Refuge from unsafe JavaScript",
+  "description": "Refuge from unprincipled JavaScript",
   "license": "MIT",
   "repository": {
     "type": "git",
--- a/index.js
+++ b/index.js
@@ -3,7 +3,7 @@
  ####   ###   ####
 #####   ###########   sanctuary
 ########   ########   noun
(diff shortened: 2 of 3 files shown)


Avaq:
I'm not sure about this change. I might agree that it reflects the broader goal in a better way, but I also feel like it wouldn't resonate with people quite as much as the previous phrase: Everybody knows, and cares about the fact, that JavaScript is *unsafe*. "Unprincipled", however, is somewhat more vague and may not trigger the "I care about that too!" reaction we may hope for.

I also prefer the (in my opinion) stronger semantic relationship between "unsafe" and "sanctuary", compared to "unprincipled" and "sanctuary".

davidchambers:
Great points, Aldwin! You've convinced me that the current description is better. :)

One other point in favour of the incumbent: _unsafe_ has two syllables whereas _unprincipled_ has four.

Issue #351, opened 2017-02-23
Add S.groupBy


svozza:
Implement a Sanctuary version of [R.groupBy](http://ramdajs.com/0.23.0/docs/#groupBy) although probably closer to the Haskell version so we would have a type signature of:

```haskell
groupBy :: (a -> a -> Bool) -> [a] -> [[a]] 
```

sadasant:
Hi! What do you people think of:

```javascript
  function groupBy(fn, xs) {
    var acc = [[xs[0]]];
    for (var i = 1; i < xs.length; i += 1) {
      var a = acc[acc.length - 1][0];
      var b = xs[i];
      if (fn(a)(b)) {
        acc[acc.length - 1].push(b);
      } else {
        acc.push([b]);
      }
    }
    return acc;
  }
  S.groupBy =
  def('groupBy',
      {},
      [Fn(a, Fn(a, $.Boolean)), $.Array(a), $.Array($.Array(a))],
      groupBy);
```

Would it be better with Z.reduce? Or any other preference? ❤🙏

PR #371, opened 2017-04-03
groupBy as in Haskell

sadasant:
Should work as in Haskell!

Some notes:

- This can be do in many different ways! Please give feedback
  so it can be improved \o/

Please review 🙏🙇

Closes: #351

### DIFF
--- a/index.js
+++ b/index.js
@@ -3282,6 +3282,57 @@
   S.range =
   def('range', {}, [$.Integer, $.Integer, $.Array($.Integer)], range);
 
+  //# groupBy :: (a -> a -> Boolean) -> Array a -> Array (Array a)
+  //.
+  //. Splits its array argument into an array of arrays of equal,
+  //. adjacent elements. Equality is determined by the function
+  //. provided as the first argument. Its behaviour can be surprising
+  //. for functions that aren't reflexive, transitive, and symmetric
+  //. (see [equivalence][] relation).
+  //.
+  //. See also [`groupBy_`](#groupBy_).
+  //.
+  //. Properties:
+  //.
+  //.   - `forall f :: a -> a -> Boolean, xs :: Array a.
+  //.      S.join(S.groupBy(f, xs)) = xs`
+  //.
+  //. ```javascript
+  //. > S.groupBy(S.equals, [1, 1, 2, 1, 1])
+  //. [[1, 1], [2], [1, 1]]
+  //.
+  //. > S.groupBy(x => y => x + y === 0, [2, -3, 3, 3, 3, 4, -4, 4])
+  //. [[2], [-3, 3, 3, 3], [4, -4], [4]]
+  //. ```
+  function groupBy(f, xs) {
+    return groupBy_(uncurry2(f), xs);
+  }
+  S.groupBy =
+  def('groupBy',
+      {},
+      [Fn(a, Fn(a, $.Boolean)), $.Array(a), $.Array($.Array(a))],
+      groupBy);
+
+  //# groupBy_ :: ((a, a) -> Boolean) -> Array a -> Array (Array a)
+  //.
+  //. Variant of [`groupBy`](#groupBy) which takes an uncurried function.
+  function groupBy_(f, xs) {
+    if (xs.length === 0) return [];
+    var x0 = xs[0];         // :: a
+    var active = [x0];      // :: Array a
+    var result = [active];  // :: Array (Array a)
+    for (var idx = 1; idx < xs.length; idx += 1) {
+      var x = xs[idx];
+      if (f(x0, x)) active.push(x); else result.push(active = [x0 = x]);
+    }
+    return result;
+  }
+  S.groupBy_ =
+  def('groupBy_',
+      {},
+      [$.Function([a, a, $.Boolean]), $.Array(a), $.Array($.Array(a))],
+      groupBy_);
+
   //. ### Object
 
   //# prop :: Accessible a => String -> a -> b
@@ -4180,6 +4231,7 @@
 //. [`Z.traverse`]:     v:sanctuary-js/sanctuary-type-classes#traverse
 //. [`Z.zero`]:         v:sanctuary-js/sanctuary-type-classes#zero
 //. [`of`]:             v:fantasyland/fantasy-land#of-method
+//. [equivalence]:      https://en.wikipedia.org/wiki/Equivalence_relation
 //. [parseInt]:         https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt
 //. [sanctuary-def]:    v:sanctuary-js/sanctuary-def
 //. [thrush]:           https://github.com/raganwald-deprecated/homoiconic/blob/master/2008-10-30/thrush.markdown
--- a/test/groupBy.js
+++ b/test/groupBy.js
@@ -1,0 +1,37 @@
+'use strict';
+
+var jsc = require('jsverify');
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+
+
+test('groupBy', function() {
+
+  eq(typeof S.groupBy, 'function');
+  eq(S.groupBy.length, 2);
+  eq(S.groupBy.toString(), 'groupBy :: (a -> a -> Boolean) -> Array a -> Array (Array a)');
+
+  function productsOf3(x) {
+    return function(y) {
+      return x * y % 3 === 0;
+    };
+  }
+
+  function zeroSum(x) {
+    return function(y) {
+      return x + y === 0;
+    };
+  }
+
+  eq(S.groupBy(productsOf3, []), []);
+  eq(S.groupBy(productsOf3, [1, 2, 3, 4, 5, 6, 7, 8, 9]), [[1], [2, 3], [4], [5, 6], [7], [8, 9]]);
+  eq(S.groupBy(S.equals, [1, 1, 2, 1, 1]), [[1, 1], [2], [1, 1]]);
+  eq(S.groupBy(zeroSum, [2, -3, 3, 3, 3, 4, -4, 4]), [[2], [-3, 3, 3, 3], [4, -4], [4]]);
+
+  jsc.assert(jsc.forall('nat -> nat -> bool', 'array nat', function(f, xs) {
+    return S.equals(S.join(S.groupBy(f, xs)), xs);
+  }), {tests: 100});
+
+});
--- a/test/groupBy_.js
+++ b/test/groupBy_.js
@@ -1,0 +1,35 @@
+'use strict';
+
+var jsc = require('jsverify');
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+
+
+test('groupBy_', function() {
+
+  eq(typeof S.groupBy_, 'function');
+  eq(S.groupBy_.length, 2);
+  eq(S.groupBy_.toString(), 'groupBy_ :: ((a, a) -> Boolean) -> Array a -> Array (Array a)');
+
+  function productsOf3(a, b) {
+    return a * b % 3 === 0;
+  }
+
+  function zeroSum(x, y) {
+    return x + y === 0;
+  }
+
+  eq(S.groupBy_(productsOf3)([]), []);
+  eq(S.groupBy_(productsOf3)([1, 2, 3, 4, 5, 6, 7, 8, 9]), [[1], [2, 3], [4], [5, 6], [7], [8, 9]]);
+  eq(S.groupBy_(S.equals, [1, 1, 2, 1, 1]), [[1, 1], [2], [1, 1]]);
+  eq(S.groupBy_(zeroSum)([2, -3, 3, 3, 3, 4, -4, 4]), [[2], [-3, 3, 3, 3], [4, -4], [4]]);
+
+  jsc.assert(jsc.forall('nat -> nat -> bool', 'array nat', function(f, xs) {
+    return S.equals(S.join(S.groupBy_(function(x, y) {
+      return f(x)(y);
+    }, xs)), xs);
+  }), {tests: 100});
+
+});


index.js
```diff
@@ -3232,6 +3232,51 @@
...
+  //. The behavior of groupBy aims to be equal to Haskell's groupby,
+  //. rather than Ramda's groupBy.
+  //.
+  //. The behavior of groupBy can be surprising for functions that
+  //. aren't reflexive, transitive and symmetric
+  //. (see [equivalence][] relation).
+  //.
+  //. ```javascript
+  //. > S.groupBy(S.equals, [1, 1, 2, 3, 3, 4])
+  //. [[1, 1], [2], [3, 3], [4]]
```
Avaq: It would be nice to have an example that shows what happens if an equal value reoccurs after an unequal one.

sadasant: Hi! You mean as in `[1,1,2,1...]` ?

Avaq: Yes.

sadasant: Omw :)

index.js
```diff
@@ -3232,6 +3232,54 @@
...
+  //.
+  //. > S.groupBy(S.equals)('aabccd'.split(''))
+  //. [['a', 'a'], ['b'], ['c', 'c'], ['d']]
+  //.
+  //. > S.groupBy(S.equals)([1, 1, 2, 1, 1])
+  //. [[1, 1], [2], [1, 1]]
```
sadasant: @Avaq here ^ 😁

test/groupBy.js
```diff
@@ -0,0 +1,24 @@
...
+  var productsOf3 = S.curry2(function(a, b) {
+    return a * b % 3 === 0;
+  });
+
+  eq(S.groupBy(productsOf3, []), []);
+  eq(S.groupBy(productsOf3, [[]]), [[[]]]);
+  eq(S.groupBy(productsOf3, [1, 2, 3, 4, 5, 6, 7, 8, 9]), [[1], [2, 3], [4], [5, 6], [7], [8, 9]]);
+  eq(S.groupBy(S.equals, [1, 1, 2, 3, 3, 4]), [[1, 1], [2], [3, 3], [4]]);
+  eq(S.groupBy(S.equals)('aabccd'.split('')), [['a', 'a'], ['b'], ['c', 'c'], ['d']]);
+  eq(S.groupBy(S.equals)([1, 1, 2, 1, 1]), [[1, 1], [2], [1, 1]]);
```
sadasant: @Avaq here too :)

sadasant:
Tests failed because of Karma, the browsers couldn't start apparently.

index.js
```diff
@@ -3232,6 +3232,55 @@
...
+  //.
+  //. > var P3 = S.curry2(function(a, b) { return a * b % 3 === 0; });
+  //. > S.groupBy(P3, [1, 2, 3, 4, 5, 6, 7, 8, 9])
+  //. [[1], [2, 3], [4], [5, 6], [7], [8, 9]]
+  //. ```
+  function groupBy(fn, xs) {
+    if (xs[0] === undefined) {
```
davidchambers: What are we trying to ascertain here? If we're checking to see whether `xs` is `[]` let's use `xs.length === 0` instead.

index.js
```diff
@@ -3232,6 +3232,55 @@
...
+  //. > S.groupBy(S.equals, [1, 1, 2, 3, 3, 4])
+  //. [[1, 1], [2], [3, 3], [4]]
+  //.
+  //. > S.groupBy(S.equals)('aabccd'.split(''))
+  //. [['a', 'a'], ['b'], ['c', 'c'], ['d']]
+  //.
+  //. > S.groupBy(S.equals)([1, 1, 2, 1, 1])
+  //. [[1, 1], [2], [1, 1]]
+  //.
+  //. > var P3 = S.curry2(function(a, b) { return a * b % 3 === 0; });
```
davidchambers: We're allowed to use arrow functions in doctests. This could be `a => b => a * b % 3 === 0`.

test/groupBy.js
davidchambers: For consistency let's use the `S.groupBy(_, _)` form here rather than the `S.groupBy(_)(_)` form.

test/groupBy.js
```diff
@@ -0,0 +1,24 @@
...
+});
```
davidchambers: Nit: Please include a blank line before the closing `});`. :)

test/groupBy.js
```diff
@@ -0,0 +1,24 @@
...
+
+  var productsOf3 = S.curry2(function(a, b) {
+    return a * b % 3 === 0;
+  });
+
+  eq(S.groupBy(productsOf3, []), []);
+  eq(S.groupBy(productsOf3, [[]]), [[[]]]);
+  eq(S.groupBy(productsOf3, [1, 2, 3, 4, 5, 6, 7, 8, 9]), [[1], [2, 3], [4], [5, 6], [7], [8, 9]]);
+  eq(S.groupBy(S.equals, [1, 1, 2, 3, 3, 4]), [[1, 1], [2], [3, 3], [4]]);
+  eq(S.groupBy(S.equals)('aabccd'.split('')), [['a', 'a'], ['b'], ['c', 'c'], ['d']]);
```
davidchambers: I'd prefer to see `'aabccd'.split('')` written `['a', 'a', 'b', 'c', 'c', 'd']`. I prefer the test cases with numbers, though, as they're less noisy syntactically.

index.js
```diff
@@ -4091,6 +4140,7 @@
 //. [sanctuary-def]:    v:sanctuary-js/sanctuary-def
 //. [thrush]:           https://github.com/raganwald-deprecated/homoiconic/blob/master/2008-10-30/thrush.markdown
 //. [type identifier]:  v:sanctuary-js/sanctuary-type-identifiers
+//. [equivalence]:      https://en.m.wikipedia.org/wiki/Equivalence_relation
```
davidchambers: Nit: Please preserve alphabetical order.

index.js
```diff
@@ -3232,6 +3232,55 @@
   S.range =
   def('range', {}, [$.Integer, $.Integer, $.Array($.Integer)], range);
 
+  //# groupBy :: (a -> a -> Boolean) -> Array a -> Array (Array a)
+  //.
+  //. Splits its list argument into a list of lists of equal,
```
davidchambers: Let's replace _a list of lists_ with _an array of arrays_.

index.js
```diff
@@ -3232,6 +3232,55 @@
...
+  //. adjacent elements. The equality of the elements is tested with
+  //. a curried function provided by the user in the first argument.
+  //. The behavior of groupBy aims to be equal to Haskell's groupby,
+  //. rather than Ramda's groupBy.
```
davidchambers: Let's drop this sentence. I think comparing Sanctuary functions to Ramda functions is better done in a large table somewhere. :)

index.js
```diff
@@ -3232,6 +3232,55 @@
...
+  //.
+  //. The behavior of groupBy can be surprising for functions that
```
davidchambers: Nit: Let's write behaviour with a "u" for consistency with the rest of the documentation.

test/groupBy.js
sadasant: Going!

index.js
sadasant: Ohh thank you, I was getting crazy

sadasant:
@davidchambers thank for the reviews!!!

test/groupBy.js
sadasant: I'm removing this then. Should I leave it in the documentation? (as an array of strings?)

test/groupBy.js
sadasant: Should I leave examples using `)(` in the documentation?

sadasant: I'm removing `)(` from the documentation unless specified 👍

test/groupBy.js
sadasant: I'm removing string examples from the documentation unless specified 👍

index.js
```diff
@@ -3232,6 +3232,49 @@
...
+      var a = last[0];
+      var b = xs[i];
+      if (fn(a)(b)) {
+        last.push(b);
+      } else {
+        acc.push([b]);
+      }
+    }
+    return acc;
+  }
```
davidchambers: Most of our implementations are only a few lines long; I suspect we can come up with something shorter and clearer for `groupBy` if we put our minds to it. :)

Additionally, I'd like to see a few minor changes:

  - use `f` rather than `fn`;
  - use `idx` rather than `i`; and
  - use `if (xs.length === 0) return [];` (on a single line).

sadasant: Here's a smaller version, passing tests and lint:

```
function groupBy(f, xs) {
  var acc = xs.length ? [[xs[0]]] : [];
  for (var idx = 1; idx < xs.length; idx += 1) {
    var last = acc[acc.length - 1];
    var b = xs[idx];
    if (f(last[0])(b)) last.push(b);
    else acc.push([b]);
  }
  return acc;
}
```
We can also do this with reduce, but it's not smaller, and also slower.

Avaq: I also had a go. Optimized by taking out `.length` calls and random access on every iteration:

```js
function groupBy(f, xs) {
  var len = xs.length;
  if (len === 0) return [];
  var last = [xs[0]], acc = [last];
  for (var idx = 1; idx < len; idx++) {
    var a = last[0];
    var b = xs[idx];
    if (f(a)(b)) last.push(b);
    else acc.push(last = [b]);
  }
  return acc;
}
```

sadasant: That looks good too! :D

sadasant: I'm up for anything as long as people agree \o/

Avaq: Better yet - reuse `f(a)` and no need to access `last[0]` (untested concept):

```js
function groupBy(f, xs) {
  var len = xs.length;
  if (len === 0) return [];
  var x = xs[0], fx = f(x), last = [x], acc = [last];
  for (var idx = 1; idx < len; idx++) {
    x = xs[idx];
    if (fx(x)) last.push(x);
    else {
      fx = f(x);
      acc.push(last = [x]);
    }
  }
  return acc;
}
```

sadasant: Isn't that longer? and harder to read? At least to me x)

sadasant: @Avaq that last one also has lint problems:

```
/home/sadasant/code/github.com/sadasant/sanctuary/index.js
  3259:34  error  Unary operator '++' used         no-plusplus
  3261:7   error  Expected { after 'if' condition  curly
```

sadasant: @Avaq fixing those two linting details, it works :) @davidchambers whenever you have time, perhaps you have a better idea?

davidchambers: I like these suggestions. I had a play myself, because I enjoy this sort of thing. I think this version—inspired by the suggestions above—strikes a good balance between readability and efficiency:

```javascript
function groupBy(f, xs) {
  if (xs.length === 0) return [];
  var x0 = xs[0];         // :: a
  var active = [x0];      // :: Array a
  var result = [active];  // :: Array (Array a)
  for (var idx = 1; idx < xs.length; idx += 1) {
    var x = xs[idx];
    if (f(x0)(x)) active.push(x); else result.push(active = [x0 = x]);
  }
  return result;
}
```

`result.push(active = [x0 = x])` is a dense expression, but it reads nicely to me. We're beginning a new group, so we update `x0` (which keeps track of the first value of the active group). We assign a new singleton array to `active` (which keeps track of the active group). Finally, we append the new group to `result` (which accumulates the groups).

sadasant: @Avaq if you like that one, I can upload the changes. Or if you have a comment! Feel free.

sadasant: We could find people to vote based on 👍 and the one with more 👍 wins, just in case!

davidchambers: This is more of a benevolent dictatorship than a democracy, Daniel. :wink:

sadasant: ahahahah, ok I'll add yours later

index.js
```diff
@@ -3232,6 +3232,43 @@
...
+  //. ```
+  function groupBy(f, xs) {
+    if (xs.length === 0) return [];
+    var x0 = xs[0];         // :: a
+    var active = [x0];      // :: Array a
+    var result = [active];  // :: Array (Array a)
+    for (var idx = 1; idx < xs.length; idx += 1) {
+      var x = xs[idx];
+      if (f(x0)(x)) active.push(x); else result.push(active = [x0 = x]);
+    }
```
sadasant: @davidchambers done!

davidchambers:
Let's also define `groupBy_` with the following type:

```haskell
groupBy_ :: ((a, a) -> Boolean) -> Array a -> Array (Array a)
```

I suggest using the same approach we used for `encase2` and `encase2_`:

1.  Change the name of the implementation function from `groupBy` to `groupBy_`, and update it to take an uncurried binary function.

2.  Define `S.groupBy_` with `groupBy_` as its implementation function.

3.  Add a new `groupBy` implementation function:

    ```javascript
    function groupBy(f, xs) {
      return groupBy_(uncurry2(f), xs);
    }
    ```

index.js
```diff
@@ -3232,6 +3232,57 @@
...
+      var x = xs[idx];
+      if (f(x0, x)) active.push(x); else result.push(active = [x0 = x]);
+    }
+    return result;
+  }
+  S.groupBy_ =
+  def('groupBy_',
+      {},
+      [$.Function([a, a, $.Boolean]), $.Array(a), $.Array($.Array(a))],
+      groupBy_);
```
sadasant: @davidchambers here is it :)

davidchambers:
> groupBy for [[]] returns [[[]]], but in haskell it causes an error.

It doesn't for me:

```haskell
> groupBy (==) [[]]
[[[]]]
```

sadasant:
@davidchambers nice! I was trying with > or <, so perhaps it was that? thanks!

index.js
```diff
@@ -3232,6 +3232,57 @@
...
+  }
+  S.groupBy =
+  def('groupBy',
+      {},
+      [Fn(a, Fn(a, $.Boolean)), $.Array(a), $.Array($.Array(a))],
+      groupBy);
+
+  //# groupBy_ :: ((a, a) -> Boolean) -> Array a -> Array (Array a)
+  //.
+  //. Variant of [`groupBy`](#flip) which takes an uncurried function.
```
Avaq: `groupBy` links to `flip`.

sadasant: @Avaq nooooo! 🙈 thanks for spotting this!

sadasant:
(Exactly `+100 −0`! r/oddlysatisfying)

index.js
sadasant: @Avaq fixed it 👁

index.js
```diff
@@ -3232,6 +3232,57 @@
...
+  //.
+  //. Variant of [`groupBy`](#groupBy) which takes an uncurried function.
+  function groupBy_(f, xs) {
+    if (xs.length === 0) return [];
+    var x0 = xs[0];         // :: a
+    var active = [x0];      // :: Array a
+    var result = [active];  // :: Array (Array a)
+    for (var idx = 1; idx < xs.length; idx += 1) {
+      var x = xs[idx];
+      if (f(x0, x)) active.push(x); else result.push(active = [x0 = x]);
```
svozza: Could this not use the ternary operator instead or do we want to emphasise that it's a side effect?

```js
f(x0, x) ? active.push(x) : result.push(active = [x0 = x]);
```

sadasant: @svozza I think you're right, I'd rather use the ternary operator here. It's easier for me to see it than with the one liner if/else. Should I do the change or wait for David and Aldwin?

davidchambers: I prefer `if (...) ...; else ...;` for exactly the reason you suggested, Stefano.

index.js
```diff
@@ -3232,6 +3232,57 @@
...
+  //# groupBy :: (a -> a -> Boolean) -> Array a -> Array (Array a)
+  //.
+  //. Splits its array argument into an array of arrays of equal,
+  //. adjacent elements. The equality of the elements is tested with
+  //. a curried function provided by the user in the first argument.
+  //.
+  //. The behaviour of groupBy can be surprising for functions that
+  //. aren't reflexive, transitive and symmetric
+  //. (see [equivalence][] relation).
+
```
davidchambers: Please add `//.` to this line. Without this the resulting Markdown will not have a paragraph break here.

sadasant: Done!

test/groupBy.js
```diff
@@ -0,0 +1,24 @@
...
+  eq(typeof S.groupBy, 'function');
+  eq(S.groupBy.length, 2);
+  eq(S.groupBy.toString(), 'groupBy :: (a -> a -> Boolean) -> Array a -> Array (Array a)');
+
+  var productsOf3 = S.curry2(function(a, b) {
+    return a * b % 3 === 0;
+  });
+
+  eq(S.groupBy(productsOf3, []), []);
+  eq(S.groupBy(productsOf3, [[]]), [[[]]]);
```
davidchambers: This is a type error:

```
> import Data.List (groupBy)
> groupBy (\x -> \y -> x * y `mod` 3 == 0) [[]]

<interactive>:3:1:
    No instance for (Integral [t0]) arising from a use of ‘it’
    In a stmt of an interactive GHCi command: print it
```

sadasant: @davidchambers yes, I guess that was what I was receiving! Should we remove this example?

Avaq: I noticed that too. But in Sanctuary that would be caught by defining `productsOf3` with `def`. No way for `groupBy` to know whether it's passing an invalid argument into `f`.

davidchambers: I'm not suggesting `S.groupBy` should have caught this. I'm simply suggesting we should not include test cases which would not type check. ;)

sadasant: Removed those tests!

index.js
```diff
@@ -3232,6 +3232,57 @@
   S.range =
   def('range', {}, [$.Integer, $.Integer, $.Array($.Integer)], range);
 
+  //# groupBy :: (a -> a -> Boolean) -> Array a -> Array (Array a)
+  //.
+  //. Splits its array argument into an array of arrays of equal,
+  //. adjacent elements. The equality of the elements is tested with
+  //. a curried function provided by the user in the first argument.
```
davidchambers: Let's shorten the second sentence to this:

_Equality is determined by the function provided as the first argument._

index.js
```diff
@@ -3232,6 +3232,57 @@
...
+  //.
+  //. The behaviour of groupBy can be surprising for functions that
+  //. aren't reflexive, transitive and symmetric
+  //. (see [equivalence][] relation).
```
davidchambers: Let's make this part of the first paragraph. We can then drop "of groupBy" (relying on context). Finally, please add an Oxford comma after "transitive".

index.js
```diff
@@ -3232,6 +3232,57 @@
...
+  //.
+  //. See also [`groupBy_`](#groupBy_).
+  //.
+  //. ```javascript
+  //. > S.groupBy(S.equals, [1, 1, 2, 3, 3, 4])
+  //. [[1, 1], [2], [3, 3], [4]]
+  //.
+  //. > S.groupBy(S.equals, [1, 1, 2, 1, 1])
+  //. [[1, 1], [2], [1, 1]]
```
davidchambers: This example is strictly more informative than the preceding one. Let's drop the preceding one.

index.js
```diff
@@ -3232,6 +3232,57 @@
...
+  //.
+  //. > S.groupBy(a => b => a * b % 3 === 0, [1, 2, 3, 4, 5, 6, 7, 8, 9])
+  //. [[1], [2, 3], [4], [5, 6], [7], [8, 9]]
```
davidchambers: I like the inclusion of a "surprising" example. Let's avoid `%`, though, as simpler examples are available to us. I suggest this instead:

```javascript
> S.groupBy(x => y => x + y === 0, [2, -3, 3, 3, 3, 4, -4, 4])
[[2], [-3, 3, 3, 3], [4, -4], [4]]
```

index.js
```diff
@@ -4087,6 +4138,7 @@
 //. [`Z.traverse`]:     v:sanctuary-js/sanctuary-type-classes#traverse
 //. [`Z.zero`]:         v:sanctuary-js/sanctuary-type-classes#zero
 //. [`of`]:             v:fantasyland/fantasy-land#of-method
+//. [equivalence]:      https://en.m.wikipedia.org/wiki/Equivalence_relation
```
davidchambers: Let's not link to the mobile version.

test/groupBy.js
```diff
@@ -0,0 +1,23 @@
...
+
+test('groupBy', function() {
+
+  eq(typeof S.groupBy, 'function');
+  eq(S.groupBy.length, 2);
+  eq(S.groupBy.toString(), 'groupBy :: (a -> a -> Boolean) -> Array a -> Array (Array a)');
+
+  var productsOf3 = S.curry2(function(a, b) {
+    return a * b % 3 === 0;
+  });
```
davidchambers: We should avoid `S.curry2` here, as it will return a function which can be applied `(x)(y)`-style or `(x, y)`-style. Let's define a curried function manually:

```javascript
function productsOf3(x) {
  return function(y) {
    return x * y % 3 === 0;
  };
}
```

test/groupBy.js
```diff
@@ -0,0 +1,23 @@
...
+
+  eq(S.groupBy(productsOf3, []), []);
+  eq(S.groupBy(productsOf3, [1, 2, 3, 4, 5, 6, 7, 8, 9]), [[1], [2, 3], [4], [5, 6], [7], [8, 9]]);
+  eq(S.groupBy(S.equals, [1, 1, 2, 3, 3, 4]), [[1, 1], [2], [3, 3], [4]]);
+  eq(S.groupBy(S.equals, [1, 1, 2, 1, 1]), [[1, 1], [2], [1, 1]]);
+
+});
```
davidchambers: As a stretch goal you could define a property-based test to verify this property:

  - `forall f :: a -> a -> Boolean, xs :: Array a. S.join(S.groupBy(f, xs)) = xs`

Regardless, let's add this to the description of `S.groupBy`.

See the `S.regexEscape` description and tests for guidance.

sadasant: Hi! Why is this a stretch goal? Isn't it just simply adding these lines? (They go over the line width limit though).

```diff
  //. See also [`groupBy_`](#groupBy_).
  //.
+  //. Properties:
+  //.
+  //.   - `forall f :: a -> a -> Boolean, xs :: Array a. S.join(S.groupBy(f, xs)) = xs`
+  //.
```

davidchambers: The stretch goal is actually testing the property. See [__test/regexEscape.js__][1], for example.


[1]: https://github.com/sanctuary-js/sanctuary/blob/v0.12.2/test/regexEscape.js#L18-L24

sadasant: Doing it!

sadasant: @davidchambers this is working:

```
  jsc.assert(jsc.forall('nat -> nat -> bool', 'array nat', function(f, xs) {
    return S.equals(S.join(S.groupBy(f, xs)), xs);
  }), {tests: 1000});
```

Do you think that's ok? Or would you do something different?

sadasant: And this one for `groupBy_`:

```
  jsc.assert(jsc.forall('nat -> nat -> bool', 'array nat', function(f, xs) {
    return S.equals(S.join(S.groupBy_(function(x, y) {
      return f(x)(y);
    }, xs)), xs);
  }), {tests: 1000});
```

Just to ask again, please tell me if you would do it differently! \o/

They're expensive though, but that's expected, right?

```
  ✓ groupBy (8479ms)
  ✓ groupBy_ (5752ms)
```

davidchambers: Does `'nat -> nat -> bool'` generate a function which arbitrarily chooses between `true` and `false`? If so, that seems appropriate. We don't mind in this context whether the function is actually an equality function.

You could drop the number of tests from `1000` to `100` to avoid making `make test` a lot slower. :)

sadasant: Ok I'll drop the number! And yeah it does the job, here's a section of the `console.log(xs, f(1)(2))`:

```
 ] false
[ 18, 15 ] false
[ 18, 12, 27, 39 ] false
[ 23, 23, 26, 32, 0 ] true
[] true
[ 45, 34 ] false
[ 33, 7, 11 ] false
[ 25, 3, 16, 42, 30, 0 ] false
[] true
[ 25, 12 ] true
[ 14, 17, 7 ] true
[ 25, 12, 5, 15, 19 ] false
[ 4, 1 ] true
[ 40, 22 ] false
[ 4 ] true
[ 1 ] false
[ 38, 27 ] false
[] true
[] true
[ 4, 5, 1 ] true
[ 10, 11, 16, 14 ] true
[] true
[ 18, 13, 20 ] true
[ 0 ] false
[ 1, 0, 10 ] true
[ 8, 12, 5, 4 ] true
[ 3, 8, 12, 33, 33 ] true
[ 32, 35, 13, 7, 48, 27 ] false
[ 2, 3, 7 ] true
[] false
[ 22, 8, 27, 14 ] false
[ 14, 7, 1, 2 ] false
[] true
[ 25, 17, 8 ] true
[ 1, 10 ] true
```

sadasant: With 100 it drops to:

```
  ✓ groupBy (1147ms)
  ✓ groupBy_ (606ms)
```

Much better ^_^

index.js
```diff
@@ -3232,6 +3232,57 @@
   S.range =
   def('range', {}, [$.Integer, $.Integer, $.Array($.Integer)], range);
 
+  //# groupBy :: (a -> a -> Boolean) -> Array a -> Array (Array a)
+  //.
+  //. Splits its array argument into an array of arrays of equal,
+  //. adjacent elements. Equality is determined by the function
+  //. provided as the first argument. It's behaviour can be surprising
```
davidchambers: `s/It's/Its/`

davidchambers:
I'm sorry for letting this pull request languish, @sadasant! Please fix the typo I just spotted, squash your commits, `git rebase master`, update this pull request, and leave a comment to let me know that you've done so. I'll then merge these additions. :)

sadasant:
Thanks! I will let you know when it's done ✌

sadasant:
@davidchambers Rebased, fixed typo and squashed :)

sadasant:
@davidchambers woohooo!! :D I'm trying to find time to do something else, but these days have been intense 🙈

PR #377, opened 2017-04-08
documentation: list differences between Ramda and Sanctuary

davidchambers:
This began as a [comment][1] @Avaq posted in the Gitter room several months ago. I have made significant changes in order to match the tone of the documentation, and have added several points.

I'd love feedback from @CrossEye and @buzzdecafe. Let me know if any of my Ramda claims are inaccurate.

@Avaq and others, have I missed any philosophical differences between Ramda and Sanctuary? Also, please suggest improvements to the wording, grammar, punctuation, and formatting. I consider this a starting point and look forward to refining this section as a group. :)

[1]: https://gitter.im/sanctuary-js/sanctuary?at=585407a0058ca96737550316


scripts/version-urls
```diff
@@ -19,6 +19,6 @@ process.stdout.write(fs.readFileSync(process.argv[2], 'utf8').replace(
              '/tree/v' + deps[name] + parts.slice(1).join('');
     }
     process.stderr.write('Exact version not specified for ' + name + '\n');
-    process.exit(1);
+//  process.exit(1);
```
svozza: Was this intentional?

index.js
```diff
@@ -43,6 +43,72 @@
 //.
 //. Sanctuary is designed to work in Node.js and in ES5-compatible browsers.
 //.
+//. ## Ramda
+//.
+//. Sanctuary started as a complementary project to [Ramda][], aiming to
+//. provide some of Ramda's functionality in a "safer" way, by utilizing
```
svozza: How about _Sanctuary was conceived as a complementary project to [Ramda][], aiming to provide some of its functionality in a safer way_ to avoid repeating the word Ramda twice in quick succession.

scripts/version-urls
davidchambers: Yes, for the time being. We can't yet specify versions for `sanctuary-{either,maybe}` as these packages have not yet been published.

index.js
davidchambers: Good idea. I'll make this change. :)

davidchambers: Hmm. I think grammatically "its" refers to Sanctuary rather than Ramda. From context it may be apparent that we're referring to Ramda, but I'd prefer not to ask readers to make this switch. I'll give this sentence more thought.

davidchambers:
Do you have any thoughts on these additions, @Avaq?

index.js
```diff
@@ -43,6 +43,79 @@
 //.
 //. Sanctuary is designed to work in Node.js and in ES5-compatible browsers.
 //.
+//. ## Ramda
+//.
+//. Ramda provides a number of partial functions. These necessitate the use
```
Avaq: I'm not sure if all readers will know what "partial function" means. I learned what it means not too long ago, and consider it to be functional jargon. Maybe we should consider either linking it to its meaning, or use the phrase "functions which might return `null` or `undefined`".

index.js
```diff
@@ -43,6 +43,79 @@
...
+//. Using the Maybe type renders such guards (and null checks) unnecessary.
+//. Changing functions such as `R.head` to return Maybe values was proposed
+//. in [ramda/ramda#683][], but was considered too much of a stretch for
+//. JavaScript programmers. Sanctuary was released the following month,
+//. in January 2015, as a companion library to Ramda.
+//.
+//. In addition to broadening in scope in the years since its release,
+//. Sanctuary's philosophy has diverged from Ramda's in several respects.
+//.
+//. ### Totality
```
Avaq: Maybe we can link "partial" to here, and state that "totality" is its opposite?

index.js
```diff
@@ -43,6 +43,79 @@
...
+//.
+//. Every Sanctuary function is defined for every value in its domain. Ramda
```
Avaq: How about something like: *"The type of every Sanctuary function fully describes all of the possible input and output of the function"*?

index.js
```diff
@@ -43,6 +43,79 @@
...
+//. permissive as JavaScript is that there's no guarantee that garbage input
+//. will produce garbage output ([ramda/ramda#1413][]). Ramda performs ad hoc
+//. type checking in some such cases ([ramda/ramda#1419][]).
+//.
+//. Sanctuary can be configured to operate in garbage in, garbage out mode.
+//. Ramda cannot be configured to enforce its invariants.
+//.
+//. ### Variadic functions
+//.
+//. Ramda has a few variadic functions, and a few functions which take variadic
```
Avaq: How about *"Ramda has a few functions which take any number of arguments"*?

index.js
gabejohnson: Putting "partial function" in parenthesis (possibly linking to a more precise/comprehensive definition) would provide an education opportunity as well.

Something like "Ramda provides a number of functions which may return `null` or `undefined` ([partial functions](https://en.wikipedia.org/wiki/Partial_function))".

Same thing with [totality](https://en.wikipedia.org/wiki/Partial_function#Total_function) and [variadic functions](https://en.wikipedia.org/wiki/Variadic_function).

davidchambers: I completely agree with the sentiment, @Avaq. I'm also very much in favour of precise definitions. There are two problems with using the phrase "functions which might return `null` or `undefined`" to describe partial functions:

  - there are total functions which may return these values (e.g. [`S.I`][1]); and
  - there are partial functions which never return these values (e.g. [`R.divide`][2]).

I will make changes based on your various suggestions, Aldwin. I greatly appreciate your input!


[1]: https://sanctuary.js.org/#I
[2]: http://ramdajs.com/docs/#divide

davidchambers:
I've pushed several changes, including a rendered proof which can be viewed at <https://github.com/sanctuary-js/sanctuary/blob/davidchambers/ramda/README.md#ramda>.

index.js
```diff
@@ -43,6 +43,95 @@
...
+//. Every Sanctuary function is defined for every value which is a member of
+//. the function's input type. Such functions are known as [total functions][].
+//. Ramda, on the other hand, contains a number of [partial functions][].
+//.
+//. ### Information preservation
+//.
+//. Certain Sanctuary functions preserve more information than their Ramda
+//. counterparts. For example, one can determine the length of `xs` from
+//. the result of `S.tail(xs)` but not from the result of `R.tail(xs)`
+//. (since `R.tail([])` evaluates to `[]`).
```
gabejohnson: Perhaps giving an example of how to determine the length of `xs` from the result fo `S.tail(xs)` would be in order. I doubt it's immediately obvious to someone unfamiliar with `Maybe`.

Avaq: Maybe just:

```diff
- since `R.tail([])` evaluates to `[]`
+ `R.tail([[]])` equals `R.tail([])`, but `S.tail([[]])` does not equal `S.tail([])`
```

davidchambers: Thanks for raising this issue, @gabejohnson.

I like your suggestion, @Avaq, although I find your choice of "head" confusing. Let's use `'foo'` rather than `[]`, giving _since `R.tail(['foo'])` equals `R.tail([])`, but `S.tail(['foo'])` does not equal `S.tail([])`_.

Avaq: But `R.tail(['foo'])` **does not** equal `R.tail([])`. The point of using `[[]]` is to show that in Ramda, different inputs can lead to the same output, therefore we don't have total information preservation.

Avaq: Ah. I confused `R.tail` for `R.last`. Your suggestion is a more clear example indeed.

index.js
```diff
@@ -43,6 +43,95 @@
...
+//. for transducers.
+//.
+//. ### Modularity
+//.
+//. Whereas Ramda has no dependencies, Sanctuary has a modular design:
+//. [sanctuary-def][] provides type checking, [sanctuary-type-classes][]
+//. provides Fantasy Land functions and type classes, and [sanctuary-maybe][]
+//. and [sanctuary-either][] will soon provide the Maybe and Either data types.
+//. Not only does this approach reduce the complexity of Sanctuary itself, but
+//. it allows these components to be reused in other contexts.
```
Avaq: Are we waiting for these packages to exist before merging this PR? That could take a while. Why don't we just leave that one sentence out for now, and add back it once it's relevant. That we we can merge this PR (after removing the rendered proof) and have our users enjoy the new source of information.

index.js
```diff
@@ -43,6 +43,95 @@
...
+//.
+//. ### Totality
+//.
+//. Every Sanctuary function is defined for every value which is a member of
+//. the function's input type. Such functions are known as [total functions][].
+//. Ramda, on the other hand, contains a number of [partial functions][].
+//.
+//. ### Information preservation
+//.
+//. Certain Sanctuary functions preserve more information than their Ramda
```
tomkel: What information is retained?

index.js
```diff
@@ -43,6 +43,95 @@
...
+//. Sanctuary has no variadic functions, nor any functions which take variadic
+//. functions as arguments. Sanctuary provides two "lift" functions, each with
+//. a helpful type signature:
+//.
+//.     S.lift2 :: Apply f => (a -> b -> c) -> f a -> f b -> f c
+//.     S.lift3 :: Apply f => (a -> b -> c -> d) -> f a -> f b -> f c -> f d
+//.
+//. ### Implicit context
+//.
+//. Ramda provides a handful of functions for working with `this`. Sanctuary
```
tomkel: Which set of functions?

index.js
davidchambers: A couple of examples:

  - `S.tail(xs)` evaluates to a value which indicates whether `xs` is `[]`; `R.tail([])` and `R.tail(['foo'])` both evaluate to `[]`.

  - `S.stripPrefix('https://', url)` evaluates to a value which indicates whether `url` begins with `'https://'`; `R.replace(/^https:[/][/]/, '', url)` does not.

index.js
davidchambers: [`R.bind`][1] and [`R.invoke`][2] spring to mind. There may be others which I'm forgetting.


[1]: http://ramdajs.com/docs/#bind
[2]: http://ramdajs.com/docs/#invoker

index.js
tomkel: Thanks for clarifying. Maybe this paragraph could be expanded with a short explanation of the fact that a Maybe retains more information? In its current form the paragraph is not very enlightening.

index.js
tomkel: Same as above, maybe consider adding 
>such as  `R.bind` and `R.invoke`.

?

davidchambers: Good idea. I've updated this section.

index.js
davidchambers: I've reworked this paragraph. Let me know if you find it clearer now. :)

tomkel: Yes, very enlightening. Thank you :)

index.js
```diff
@@ -43,6 +43,143 @@
...
+//. ways to apply `R.reduce` to `R.add`, `0`, and `xs`:
+//.
+//.   - `R.reduce (R.add) (0) (xs)`
+//.   - `R.reduce (R.add) (0, xs)`
+//.   - `R.reduce (R.add, 0) (xs)`
+//.   - `R.reduce (R.add, 0, xs)`
+//.
+//. Ramda supports all these forms because curried functions enable partial
+//. application, one of the library's tenets, but `f(x)(y)(z)` is considered
+//. too unfamiliar and too unattractive to appeal to JavaScript programmers.
```
davidchambers: Is this accurate, @CrossEye?

index.js
```diff
@@ -43,6 +43,144 @@
...
+//.   - `R.reduce (R.__, 0, xs) (R.add)`
+//.   - `R.reduce (R.add, R.__, xs) (0)`
+//.   - `R.reduce (R.__, 0) (R.add) (xs)`
+//.   - `R.reduce (R.__, 0) (R.add, xs)`
+//.   - `R.reduce (R.__, R.__, xs) (R.add) (0)`
+//.   - `R.reduce (R.__, R.__, xs) (R.add, 0)`
+//.
+//. ### Variadic functions
+//.
+//. Ramda provides several functions with take any number of arguments. These
```
jazen: Typo `with` -> `which`

davidchambers: I must have read this sentence a dozen times without noticing that. Well spotted, @jazen. Would you like to open a pull request to receive credit for having sharp eyes?

jazen: Sure, thanks 😄

PR #382, opened 2017-04-13
array: generalize S.pluck

### LINKED ISSUE #381: S.pluck documentation is incorrect
The documentation for [`pluck`](https://github.com/sanctuary-js/sanctuary/blob/0853d0a592233fd602e8167760e33812d1870fa8/index.js#L3044-L3045) states that it is equivalent to `map(prop(k), xs)`. This is only the case if `xs` is an `Array`.

Either the documentation should be changed, or `pluck` should be updated to handle Functors in general (matching [`R.pluck`](http://ramdajs.com/docs/#pluck)'s behavior).

Incidentally, `R.pluck` works for Arrays and indices as well as Objects and keys (for obvious reasons).

davidchambers:
Fixes #381

### DIFF
--- a/index.js
+++ b/index.js
@@ -3209,7 +3209,7 @@
   }
   S.find = def('find', {}, [Pred(a), $.Array(a), $Maybe(a)], find);
 
-  //# pluck :: Accessible a => String -> Array a -> Array b
+  //# pluck :: (Accessible a, Functor f) => String -> f a -> f b
   //.
   //. Combines [`map`](#map) and [`prop`](#prop). `pluck(k, xs)` is equivalent
   //. to `map(prop(k), xs)`.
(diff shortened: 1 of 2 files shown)


test/pluck.js
```diff
@@ -10,13 +10,16 @@ test('pluck', function() {
 
   eq(typeof S.pluck, 'function');
   eq(S.pluck.length, 2);
-  eq(S.pluck.toString(), 'pluck :: Accessible a => String -> Array a -> Array b');
+  eq(S.pluck.toString(), 'pluck :: (Accessible a, Functor f) => String -> f a -> f b');
 
   throws(function() { S.pluck('a', [{a: 1}, {b: 2}]); },
          TypeError,
-         '‘pluck’ expected object at index 1 to have a property named ‘a’; {"b": 2} does not');
+         '‘prop’ expected object to have a property named ‘a’; {"b": 2} does not');
```
svozza: Could be a bit confusing for users trying to debug an issue and also leaks implementation details.

davidchambers: Indeed. This is why we are not currently using `prop` internally.

Given the definition of `pluck` (_Combines `map` and `prop`._) people should not be particularly surprised by references to `prop` in error messages.

svozza: > Given the definition of pluck (Combines map and prop.)

Yeah, good point.

index.js
```diff
@@ -3209,27 +3209,25 @@
...
+  //. Just(1)
   //. ```
   function pluck(key, xs) {
-    return xs.map(function(x, idx) {
-      if (key in Object(x)) return x[key];
-      throw new TypeError('‘pluck’ expected object at index ' + idx +
-                          ' to have a property named ‘' + key + '’; ' +
-                          Z.toString(x) + ' does not');
-    });
+    return Z.map(function(obj) { return prop(key, obj); }, xs);
```
gabejohnson: Given that `prop(key)` is curried, would this be better written?
```js
return Z.map(prop(key), xs);
```

davidchambers: `S.prop` is curried, but here [`prop`][1] refers to the internal implementation which is not curried.


[1]: https://github.com/sanctuary-js/sanctuary/blob/v0.12.2/index.js#L3129-L3133

gabejohnson: That's right. Looks good then.

davidchambers:
I was about to merge this pull request when I noticed your `update-pluck` branch in the network graph, @gabejohnson. I like your approach as it avoids mentioning `prop` in error messages. I've updated this pull request. Would you mind giving it one more look?

test/pluck.js
davidchambers: We no longer mention `prop`. :)

gabejohnson:
@davidchambers my apologies. I wrote that while eating lunch last week and promptly forgot about it. It works for me. 😄

Issue #385, opened 2017-04-18
partition :: (a -> Boolean) -> Array a -> Pair (Array a) (Array a)


davidchambers:
Haskell provides [`partition`][1] in Data.List, but it's just one of many [`partition`][2] functions. Can we generalize? The first step might be to accept any Foldable:

```haskell
partition :: Foldable f => (a -> Boolean) -> f a -> Pair (Array a) (Array a)
```

Could we go one step further by parameterizing `Array` in the return type? Would adding a Monoid constraint to `f` be sufficient?

[1]: http://hackage.haskell.org/package/base-4.9.1.0/docs/Data-List.html#v:partition
[2]: https://www.haskell.org/hoogle/?hoogle=partition

gabejohnson:
Can you give some examples of the behavior under different `Foldable` and `Monoid` combinations?

gabejohnson:
Nevermind. I just looked at the second link. Sounds good to me!

davidchambers:
I was imagining something like this:

```javascript
> S.partition(S.odd, [1, 2, 3])
[[1, 3], [2]]

> S.partition(S.odd, Cons(1, Cons(2, Cons(3, Nil))))
[Cons(1, Cons(3, Nil)), Cons(2, Nil)]
```

Does this seem reasonable?

safareli:
``` purescript 
a :: (Monoid m, Foldable f)
  => (a -> Boolean) -> (a -> m) -> f a -> Pair m m
  
c :: (Monoid m, Monoid m', Foldable f)
  => (a -> Boolean) -> (a -> m) ->  (a -> m') -> f a -> Pair m m'
  
c :: (Monoid m, Monoid m',  Foldable f)
  => (a -> Either m m') -> f a -> Pair m m'
  
d :: (Monoid m,  Foldable f)
  => (a -> Either m m) -> f a -> Pair m m
  ```


I think `c` is best from this 4.

If we go farther [`These`](https://hackage.haskell.org/package/these-0.7.3/docs/Data-These.html) type could be used instead of `Either`. it will allow to either return `m'` or `m` or both `m` and `m'`. with this one could for example accumulate Min and Max in one fold this way:
``` js 
// in this case  -- partition :: (Int -> These Min Max) -> List Int -> Pair Min Max
partition(a => These(Min(a), Max(b)), as)
```

davidchambers:
[...]

PR #386, opened 2017-04-19
array: add S.partition

davidchambers:
As @safareli demonstrated in https://github.com/sanctuary-js/sanctuary/issues/385#issuecomment-294993056, this is just one of several `partition` functions from which we can choose. Is this the best choice? I don't know. I do appreciate how nicely it works with `tagBy` (#379).

### DIFF
--- a/index.js
+++ b/index.js
@@ -3456,6 +3456,30 @@
   }
   S.find = def('find', {f: [Z.Foldable]}, [Pred(a), f(a), $Maybe(a)], find);
 
+  //# partition :: (Applicative f, Foldable f, Semigroup (f b), Semigroup (f c)) => (a -> Either b c) -> f a -> Pair (f b) (f c)
+  //.
+  //. Partitions a [Foldable][] into a [`Pair`][$.Pair] of Foldables, the first
+  //. comprising all the Left values returned by the given function, the second
+  //. comprising all the Right values returned by the given function.
+  //.
+  //. ```javascript
+  //. > S.partition(S.tagBy(S.even), [1, 2, 3, 4, 5])
+  //. [[1, 3, 5], [2, 4]]
+  //. ```
+  function partition(f, foldable) {
+    var empty = Z.empty(foldable.constructor);
+    return Z.reduce(function(pair, x) {
+      return either(function(x) { return [append(x, pair[0]), pair[1]]; },
+                    function(x) { return [pair[0], append(x, pair[1])]; },
+                    f(x));
+    }, [empty, empty], foldable);
+  }
+  S.partition =
+  def('partition',
+      {f: [Z.Applicative, Z.Foldable, Z.Semigroup]},
+      [Fn(a, $Either(b, c)), f(a), $.Pair(f(b), f(c))],
+      partition);
+
   //# pluck :: (Accessible a, Functor f) => String -> f a -> f b
   //.
   //. Combines [`map`](#map) and [`prop`](#prop). `pluck(k, xs)` is equivalent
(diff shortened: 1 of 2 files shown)


gabejohnson:
@davidchambers do you have any use cases in mind where you would _not_ use a partially applied `tagBy` as the first argument? I worry about the ergonomics. If the use of `partition` requires the use of `tagBy` in most cases I'd rather see this function named something else and have `partition` be less general.

index.js
```diff
@@ -3209,6 +3209,30 @@
...
+  //. Partitions a [Foldable][] into a [`Pair`][$.Pair] of Foldables, the first
+  //. comprising all the Left values returned by the given function, the second
+  //. comprising all the Right values returned by the given function.
+  //.
+  //. ```javascript
+  //. > S.partition(S.tagBy(S.even), [1, 2, 3, 4, 5])
+  //. [[1, 3, 5], [2, 4]]
+  //. ```
+  function partition(f, foldable) {
+    var empty = Z.empty(foldable.constructor);
```
safareli: I think foldable is not necessarily monoid

index.js
```diff
@@ -3209,6 +3209,30 @@
   }
   S.find = def('find', {}, [Pred(a), $.Array(a), $Maybe(a)], find);
 
+  //# partition :: (Applicative f, Foldable f, Semigroup (f b), Semigroup (f c)) => (a -> Either b c) -> f a -> Pair (f b) (f c)
```
safareli: If we say that `f a` and `f b` Semigroups then maybe just say that `f` is `Alternative`

test/partition.js
```diff
@@ -0,0 +1,17 @@
...
+var eq = require('./internal/eq');
+
+
+test('partition', function() {
+
+  eq(typeof S.partition, 'function');
+  eq(S.partition.length, 2);
+  eq(S.partition.toString(), 'partition :: (Applicative f, Foldable f, Semigroup f) => (a -> Either b c) -> f a -> Pair (f b) (f c)');
+
+  eq(S.partition(S.tagBy(S.even), []), [[], []]);
```
safareli: You might want have list as input but want to produce two `Sets` which you can now

davidchambers:
Thank you for your feedback, Irakli. When I have some time I'll have a go at changing this to the even more general function you favour. :)

safareli:
Welcome

gabejohnson:
@davidchambers `partition` can be written in terms of `filter`. So this might be a good candidate for a `Z` function.

gabejohnson:
The signature would be `partition :: Filterable f => (a -> Boolean) -> f a -> Pair (f a) (f a)`

davidchambers:
Great point, Gabe! I'll submit a pull request. :)

davidchambers:
I made a change (locally) to sanctuary-type-classes with your proposed `partition` function, Gabe. The implementation could not be simpler:

```javascript
function partition(pred, filterable) {
  return [filter(pred, filterable), reject(pred, filterable)];
}
```

Now that we have `filter` and `reject` I no longer think `partition` is a good idea. Code expressed directly in terms of `filter` and `reject` is clearer, if a little repetitive:

```javascript
const yeses = S.filter (pred) (xs);
const nopes = S.reject (pred) (xs);
```

```javascript
const [yeses, nopes] = S.partition (pred) (xs);
```

With `partition` one must remember the order in which the results are returned (the type signature provides no indication). Furthermore, introducing `Pair` at this time is questionable given that its meaning may soon change (sanctuary-js/sanctuary-pair#3). Lastly, I'm concerned that people will assume that `partition` traverses the structure once rather than twice, and will write less declarative code in pursuit of supposed efficiency gains.

PR #387, opened 2017-04-28
sanctuary-def@0.10.1

davidchambers:
<https://github.com/sanctuary-js/sanctuary-def/releases/tag/v0.10.0>

### DIFF
--- a/test/internal/sanctuary.js
+++ b/test/internal/sanctuary.js
@@ -5,14 +5,14 @@ var $ = require('sanctuary-def');
 var S = require('../..');
 
 
-//  UnaryType :: String -> (Type -> Type)
+//  UnaryType :: String -> Type
 function UnaryType(typeIdent) {
   return $.UnaryType(
     typeIdent,
     '',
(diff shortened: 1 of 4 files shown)


test/internal/sanctuary.js
```diff
@@ -17,8 +17,8 @@ function UnaryType(typeIdent) {
 
 //  env :: Array Type
 var env = S.env.concat([
-  UnaryType('sanctuary/Compose'),
-  UnaryType('sanctuary/Identity')
+  UnaryType('sanctuary/Compose')($.Unknown),
+  UnaryType('sanctuary/Identity')($.Unknown)
```
Avaq: It looks like the purpose of `UnaryType` is to mock a type for the environment. So I think you may as well apply `$.Unknown` to the type inside the `UnaryType` function, changing its signature to `String -> Type`.

davidchambers:
<https://github.com/sanctuary-js/sanctuary-def/releases/tag/v0.10.1>

PR #388, opened 2017-04-30
sanctuary-type-classes@5.0.0

davidchambers:
<https://github.com/sanctuary-js/sanctuary-type-classes/releases/tag/v5.0.0>

This pull request adds four new functions: `lt`, `lte`, `gt`, and `gte`. It also replaces the internal Ord type class with the one defined in sanctuary-type-classes, making `min` and `max` even more general.

This pull request defines `fantasy-land/lte` for Maybe and Either. These additions will need to be applied to sanctuary-js/sanctuary-maybe#3 and sanctuary-js/sanctuary-either#3 respectively.

Note that `S.equals(0, -0)` will evaluate `true` as a result of this change. Since `-0` is neither less than nor greater than `0`, it must be equal to `0` in order to satisfy the [Ord][1] totality law.

@sadasant, you should be able to resume work on #365 once this pull request has been merged. :)

[1]: https://github.com/fantasyland/fantasy-land#ord

### DIFF
--- a/index.js
+++ b/index.js
@@ -247,6 +247,11 @@
     };
   }
 
+  //  negativeZero :: Number -> Boolean
+  function negativeZero(n) {
+    return n === 0 && 1 / n === -Infinity;
+  }
+
   //  typeEq :: String -> a -> Boolean
   function typeEq(typeIdent) {
     return function(x) {
@@ -276,17 +281,6 @@
     'sanctuary/Accessible',
     [],
     function(x) { return x != null; }
-  );
-
-  //  Ord :: TypeClass
-  var Ord = Z.TypeClass(
-    'sanctuary/Ord',
-    [],
-    function(x) {
-      return $.String._test(x) ||
-             $.ValidDate._test(x) ||
-             $.ValidNumber._test(x);
-    }
   );
 
   //  readmeUrl :: String -> String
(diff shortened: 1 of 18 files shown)


davidchambers:
Thank you for reminding me about this, Gabe. I'll suffix these functions and add the *real* `lt`, `lte`, `gt`, and `gte`. :)

davidchambers:
Which of these (equivalent) implementations do you prefer?

:tada: Option 1

```javascript
Either.prototype['fantasy-land/lte'] = function(other) {
  return this.isLeft ?
    other.isRight || Z.lte(this.value, other.value) :
    other.isRight && Z.lte(this.value, other.value);
};
```

:heart: Option 2

```javascript
Either.prototype['fantasy-land/lte'] = function(other) {
  return this.isLeft === other.isLeft ?
    Z.lte(this.value, other.value) :
    this.isLeft;
};
```

index.js
```diff
@@ -554,16 +548,181 @@
...
+  //. > S.lte_([1, 2, 3], [1, 2, 3])
+  //. true
+  //.
+  //. > S.lte_([1, 2, 3], [1, 2, 4])
+  //. true
+  //.
+  //. > S.lte_([1, 2, 3], [1, 2])
+  //. false
+  //. ```
+  S.lte_ = def('lte_', {a: [Z.Ord]}, [a, a, $.Boolean], Z.lte);
```
gabejohnson: We need these functions for type checking, but part of me feels like they shouldn't be curried. The unprefixed variants are provided for partial application. My understanding is that these were intended to be used in place of operators.

davidchambers: We *could* change the type of the suffixed functions to `Ord a => Pair a a -> Boolean`; otherwise we would need to make changes to sanctuary-def to make currying optional.

davidchambers: sanctuary-js/sanctuary-def#147 provides a way to prevent `S.lte` and friends from being applied to two arguments at once. Preventing `S.lte_` and friends from being partially applied is not so easy.

gabejohnson: Maybe we should just wait on that restriction then.

davidchambers:
This pull request should not be merged until #391 has been merged, to avoid getting the library into an inconsistent state (`S.lt` and friends would be "flipped" but `S.sub` would not).

index.js
```diff
@@ -2149,7 +2343,35 @@
...
+  //. > S.lte_(S.Right(0), S.Left(10))
+  //. false
+  //.
+  //. > S.lte_(S.Right(0), S.Right(1))
+  //. true
+  //.
+  //. > S.lte_(S.Right(1), S.Right(0))
+  //. false
+  //. ```
+  Either.prototype['fantasy-land/lte'] = function(other) {
```
Avaq: Shouldn't this be implemented conditionally, similarly to how we did Semigroup on Maybe? Or do all JavaScript values have `Ord`, as is the case with `Setoid`?

davidchambers: > Shouldn't this be implemented conditionally, similarly to how we did Semigroup on Maybe?

You're quite right. Excellent spotting! :eyes:

PR #389, opened 2017-05-02
array: generalize S.find

davidchambers:
This leads me to believe that sanctuary-type-classes should ensure consistent enumeration order in [`Object$prototype$reduce`][1] (by sorting the keys). Do others agree that this is desirable?

```haskell
> find (const True) (Map.fromList [("x", 0), ("y", 1)])
Just 0

> find (const True) (Map.fromList [("y", 1), ("x", 0)])
Just 0
```

[1]: https://github.com/sanctuary-js/sanctuary-type-classes/blob/v5.0.0/index.js#L862-L867

### DIFF
--- a/index.js
+++ b/index.js
@@ -3434,11 +3434,11 @@
   S.joinWith =
   def('joinWith', {}, [$.String, $.Array($.String), $.String], joinWith);
 
-  //# find :: (a -> Boolean) -> Array a -> Maybe a
+  //# find :: Foldable f => (a -> Boolean) -> f a -> Maybe a
   //.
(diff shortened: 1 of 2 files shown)


gabejohnson:
[`Object$prototype$equals`](https://github.com/sanctuary-js/sanctuary-type-classes/blob/1db4db2966bea57e0c6cbab64ca382f2e8a3720c/index.js#L815-L820) and [`Object$prototype$lte`](https://github.com/sanctuary-js/sanctuary-type-classes/blob/1db4db2966bea57e0c6cbab64ca382f2e8a3720c/index.js#L823-L835) already do this. If we're going to be consistent, I don't think we really have a choice.

PR #396, opened 2017-05-21
sanctuary-type-classes@5.2.0

davidchambers:
<https://github.com/sanctuary-js/sanctuary-type-classes/releases/tag/v5.2.0>

<del>I'd like to wait until sanctuary-js/sanctuary-type-classes#56 is merged before releasing `sanctuary-type-classes@5.2.0`, at which point I will update this pull request.</del>

### DIFF
--- a/index.js
+++ b/index.js
@@ -307,6 +307,7 @@
 
   //  :: Type -> Type -> Type
   var p = $.BinaryTypeVariable('p');
+  var s = $.BinaryTypeVariable('s');
 
   //  eitherTypeIdent :: String
   var eitherTypeIdent = 'sanctuary/Either';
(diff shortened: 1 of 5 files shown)


gabejohnson:
`S.I` is equivalent to `Z.id(Function)`. What are some reasons _not_ to define `S.I` in terms of `Z.id`?

davidchambers:
> What are some reasons *not* to define `S.I` in terms of `Z.id`?

The one that occurs to me is function hoisting. Because Sanctuary's internal functions are defined via function declarations we are free to use one internal function in the body of another without worrying about ordering. Replacing `function I(x) { return x; }` with `var I = Z.id(Function);` would make ordering significant.

gabejohnson:
It doesn't appear that `I` is used internally anywhere. Thus `S.I` could be defined
```js
S.I = def('I', {}, [a, a], Z.id(Function));
```
But I suppose there's currently no benefit in doing so.

index.js
```diff
@@ -1492,23 +1503,27 @@
...
   //. ```
-  function compose(f, g, x) {
-    return f(g(x));
-  }
-  S.compose = def('compose', {}, [Fn(b, c), Fn(a, b), a, c], compose);
+  S.compose =
+  def('compose',
+      {s: [Z.Semigroupoid]},
+      [s(b, c), s(a, b), s(a, c)],
+      Z.compose);
```
gabejohnson: This is a breaking change. I didn't realize the function is currently ternary. It was surprising.

davidchambers:
> It doesn't appear that `I` is used internally anywhere.

I just used `I` in #397. ;)

index.js
davidchambers: My policy is to accept as many arguments as possible. :)

PR #397, opened 2017-05-21
array: add S.sort and S.sortBy

### LINKED ISSUE #353: Add S.sort and S.sortBy

davidchambers:
Closes #353

This takes the `sortBy` implementation from <https://github.com/sanctuary-js/sanctuary/pull/365#issuecomment-298272599>, applies @safareli's improvements from <https://github.com/sanctuary-js/sanctuary/pull/365#issuecomment-300780512>, and further improves the algorithm by using binary search.

`sort` is trivially derived via `sortBy(I)`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3606,6 +3606,94 @@
       [$.Function([a, a, $.Boolean]), $.Array(a), $.Array($.Array(a))],
       groupBy_);
 
+  //# sort :: (Ord a, Applicative m, Foldable m, Monoid (m a)) => m a -> m a
+  //.
+  //. Performs a [stable sort][] of the elements of the given structure, using
+  //. [`Z.lte`][] for comparisons.
+  //.
+  //. Properties:
+  //.
+  //.   - `S.sort(S.sort(m)) = S.sort(m)` (idempotence)
+  //.
+  //. See also [`sortBy`](#sortBy).
+  //.
+  //. ```javascript
+  //. > S.sort(['foo', 'bar', 'baz'])
+  //. ['bar', 'baz', 'foo']
+  //.
+  //. > S.sort([S.Left(4), S.Right(3), S.Left(2), S.Right(1)])
+  //. [Left(2), Left(4), Right(1), Right(3)]
+  //. ```
+  function sort(m) {
+    return sortBy(I, m);
+  }
+  S.sort =
+  def('sort',
+      {a: [Z.Ord], m: [Z.Applicative, Z.Foldable, Z.Monoid]},
+      [m(a), m(a)],
+      sort);
+
+  //# sortBy :: (Ord b, Applicative m, Foldable m, Monoid (m a)) => (a -> b) -> m a -> m a
+  //.
+  //. Performs a [stable sort][] of the elements of the given structure, using
+  //. [`Z.lte`][] to compare the values produced by applying the given function
+  //. to each element of the structure.
+  //.
+  //. Properties:
+  //.
+  //.   - `S.sortBy(f, S.sortBy(f, m)) = S.sortBy(f, m)` (idempotence)
+  //.
+  //. See also [`sort`](#sort).
+  //.
+  //. ```javascript
+  //. > S.sortBy(S.prop('rank'), [
+  //. .   {rank: 7, suit: 'spades'},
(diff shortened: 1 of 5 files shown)


test/sort.js
```diff
@@ -0,0 +1,34 @@
...
+  var count = 0;
+  Z.lte = function(x, y) { count += 1; return lte(x, y); };
+
+  eq(S.sort(S.range(0, 100)), S.range(0, 100));
+  console.log('sort(range(0, 100)):', count);
+  count = 0;
+
+  eq(S.sort(S.reverse(S.range(0, 100))), S.range(0, 100));
+  console.log('sort(reverse(range(0, 100))):', count);
+  count = 0;
```
davidchambers: The results are interesting. The counts show the number of calls to `Z.lte`.

Naive implementation:

>     sort(range(0, 100)): 4950
>     sort(reverse(range(0, 100))): 99

Binary search:

>     sort(range(0, 100)): 480
>     sort(reverse(range(0, 100))): 573

The larger the array the worse the naive implementation performs in the "unlucky" case.

index.js
```diff
@@ -3580,6 +3580,77 @@
...
+  //.
+  //. > S.sortBy(S.prop('y'), [{x: 2, y: 8}, {x: 1, y: 9}, {x: 3, y: 7}])
+  //. [{x: 3, y: 7}, {x: 2, y: 8}, {x: 1, y: 9}]
+  //. ```
+  function sortBy(f, m) {
+    var M = m.constructor;
+    return Z.reduce(
+      function(m, r) { return Z.concat(m, Z.of(M, r.x)); },
+      Z.empty(M),
+      Z.reduce(function(xs, x) {
```
safareli: as we are mutating `xs` i think we can just use `for` instead reduce here. it will make intent clearer.

index.js
```diff
@@ -3580,6 +3580,77 @@
...
+      sort);
+
+  //# sortBy :: (Ord b, Applicative m, Foldable m, Monoid (m a)) => (a -> b) -> m a -> m a
+  //.
+  //. Sorts a structure by comparing the values produced by applying the given
+  //. function to each element of the structure in turn.
+  //.
+  //. Properties:
+  //.
+  //.   - `S.sortBy(f, S.sortBy(f, m)) = S.sortBy(f, m)` (idempotence)
```
safareli: Even tho by stating that sort is `idempotent` one can conclude that sort is stable, but it would be nicer to also state it in the description.

test/sort.js
safareli: nice!

index.js
davidchambers: We are folding `m`, which is not necessarily an array.

We could, though, replace the outer `reduce`:

```javascript
function sortBy(f, m) {
  var rs = Z.reduce(function(xs, x) {
    var fx = f(x);
    var lower = 0;
    var upper = xs.length;
    while (lower < upper) {
      var idx = Math.floor((lower + upper) / 2);
      if (Z.lte(fx, xs[idx].fx)) upper = idx; else lower = idx + 1;
    }
    xs.splice(lower, 0, {x: x, fx: fx});
    return xs;
  }, [], m);

  var M = m.constructor;
  var result = Z.empty(M);
  for (var idx = 0; idx < rs.length; idx += 1) {
    result = Z.concat(result, Z.of(M, rs[idx].x));
  }
  return result;
}
```

I prefer the current approach. :)

index.js
davidchambers: The binary search algorithm was actually unstable due to the handling of equal values. I made a tweak to correct this. Please review the change. :)

Thanks for mentioning this, Irakli!

index.js
safareli: > We are folding m, which is not necessarily an array.

Oh yes missed that.

But why you prefer current approach instead of using `for` for intermediate array?

davidchambers: My preference is not strong. I'm happy to switch to the `for`-loop version I posted above if you'd prefer that. Would you like me to do so?

safareli: I would vote for using `for`, but have no strong opinion too :d

PR #399, opened 2017-05-22
sanctuary-type-identifiers@2.0.1

davidchambers:
<https://github.com/sanctuary-js/sanctuary-type-identifiers/releases/tag/v2.0.0>

Commit message:

> This commit updates the return type of `S.type`. Those wishing to access unparsed type identifiers can use sanctuary-type-identifiers directly.

### DIFF
--- a/index.js
+++ b/index.js
@@ -484,18 +484,29 @@
 
   //. ### Classify
 
-  //# type :: Any -> String
+  //# type :: Any -> { namespace :: Maybe String, name :: String, version :: Integer }
   //.
-  //. Returns the [type identifier][] of the given value.
+  //. Returns the result of parsing the [type identifier][] of the given value.
   //.
   //. ```javascript
   //. > S.type(S.Just(42))
-  //. 'sanctuary/Maybe'
+  //. {namespace: Just('sanctuary'), name: 'Maybe', version: 0}
   //.
   //. > S.type([1, 2, 3])
-  //. 'Array'
+  //. {namespace: Nothing, name: 'Array', version: 0}
   //. ```
-  S.type = def('type', {}, [$.Any, $.String], type);
+  S.type =
+  def('type',
+      {},
+      [$.Any,
+       $.RecordType({namespace: $Maybe($.String),
+                     name: $.String,
+                     version: $.Integer})],
+      function(x) {
+        var r = type.parse(type(x));
+        r.namespace = toMaybe(r.namespace);
+        return r;
+      });
 
   //# is :: TypeRep a -> Any -> Boolean
   //.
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -483,18 +483,29 @@
 
   //. ### Classify
 
-  //# type :: Any -> String
+  //# type :: Any -> { namespace :: Maybe String, name :: String, version :: Integer }
```
davidchambers: To avoid the awkwardness of built-in types forever being stuck at version zero, we could use `Maybe PositiveInteger` rather than `Integer`. I don't like the idea of an explicit `@0` being represented as `Nothing`, though.

Avaq: I think awkward is better than incorrect. It seems the root of the problem lies in the fact that `type.parse` doesn't preserve whether the version was given or not.

In the case of [type identifier derivation](https://github.com/fluture-js/concurrify/blob/2db86183a03e5d8a8a3f3d0161c6705ab6b0da2d/index.js#L64-L68), that's actually nice because we don't have to do the whole `+ (o.version ? '@' + o.version : '')` thing.

We could change the output of `type.parse` to use `Nullable Integer`, preserving the absence of a version, but that would mean extra null checks of the kind I showed above. Another idea is to add something like `type.isNative :: Any -> Boolean` or `type.hasVersion :: String -> Boolean`. I'm weighing more towards the former though.

davidchambers: Okay. Let's stick with `Integer` here.

Avaq:
I'm can't really follow what's going on here. I think it's a work in progress, no?

davidchambers:
> I think [this is] a work in progress, no?

I was debugging sanctuary-js/sanctuary-type-identifiers#13 in a roundabout manner, as Sauce Labs currently provides my only access to IE8 (I temporarily added IE8 to our test matrix in <https://github.com/sanctuary-js/sanctuary/pull/399/commits/fa75e10dba7805459436b75bd06ae7c924773427>).

I'll remove the noise from this pull request shortly. Stay tuned. :)

davidchambers:
<https://github.com/sanctuary-js/sanctuary-type-identifiers/releases/tag/v2.0.1>

index.js
```diff
@@ -484,18 +484,29 @@
...
+  //. {namespace: Nothing, name: 'Array', version: 0}
   //. ```
-  S.type = def('type', {}, [$.Any, $.String], type);
+  S.type =
+  def('type',
+      {},
+      [$.Any,
+       $.RecordType({namespace: $Maybe($.String),
+                     name: $.String,
+                     version: $.Integer})],
```
Avaq: How come this type is accepted by type variables? Because `StrMap ???` is a valid type?

davidchambers: `{namespace: Just('fluture-js'), name: 'Future', version: 2}` is a member of [`$.Object`][1]. ;)


[1]: https://github.com/sanctuary-js/sanctuary-def#Object

PR #400, opened 2017-05-22
style: tweaks

davidchambers:
In the interest of minimizing noise I'm opening a long-running "tweaks" pull request which I'll merge shortly before the next release.

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,8 +15,8 @@
 //. [![Gitter](https://img.shields.io/gitter/room/badges/shields.svg)](https://gitter.im/sanctuary-js/sanctuary)
 //.
 //. Sanctuary is a JavaScript functional programming library inspired by
-//. [Haskell][] and [PureScript][]. It's stricter than [Ramda][], and provides
-//. a similar suite of functions.
+//. [Haskell][] and [PureScript][]. It's stricter than [Ramda][], and
+//. provides a similar suite of functions.
 //.
 //. Sanctuary promotes programs composed of simple, pure functions. Such
 //. programs are easier to comprehend, test, and maintain &ndash; they are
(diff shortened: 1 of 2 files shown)


davidchambers:
Please review the first commit (which updates __index.js__). I've temporarily included a second commit (which updates __README.md__) in case you're interested in seeing the rendered version.

index.js
```diff
@@ -178,7 +178,7 @@
 //. //
 //. //   The value at position 1 is not a member of ‘FiniteNumber’.
 //. //
-//. //   See https://github.com/sanctuary-js/sanctuary-def/tree/v0.11.0#FiniteNumber for information about the sanctuary-def/FiniteNumber type.
+//. //   See https://github.com/sanctuary-js/sanctuary-def/tree/v0.11.1#FiniteNumber for information about the sanctuary-def/FiniteNumber type.
```
svozza: How come this line is so long compared to all the rest?

davidchambers: By "the rest", do you mean the rest of the lines in this error message or the rest of the lines in the file?

svozza: Sorry, the latter.

davidchambers: We're faithfully transcribing the error message, which is not line-wrapped (although perhaps it should be).

svozza: Ah, of course. Makes sense.

index.js
```diff
@@ -397,39 +397,33 @@
   //. const $ = require('sanctuary-def');
   //. const type = require('sanctuary-type-identifiers');
   //.
-  //. //    identityTypeIdent :: String
-  //. const identityTypeIdent = 'my-package/Identity';
-  //.
   //. //    Identity :: a -> Identity a
-  //. function Identity(x) {
+  //. const Identity = function Identity(x) {
   //.   if (!(this instanceof Identity)) return new Identity(x);
```
gabejohnson: Why the assignment here? Is it simply to emphasize that `Identity` is an object before adding a property to it? Or is it to avoid redefinition later?

davidchambers: Why `const Identity =`, do you mean? I wanted to align the identifier with the type signature. ;)

gabejohnson: Ah, yes. That does nice.

PR #402, opened 2017-05-30
sanctuary-type-classes@6.0.0

davidchambers:
<https://github.com/sanctuary-js/sanctuary-type-classes/releases/v6.0.0>

### DIFF
--- a/index.js
+++ b/index.js
@@ -279,6 +279,7 @@
   //  Accessible :: TypeClass
   var Accessible = Z.TypeClass(
     'sanctuary/Accessible',
+    readmeUrl('accessible-pseudotype'),
     [],
     function(x) { return x != null; }
   );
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -279,6 +279,7 @@
   //  Accessible :: TypeClass
   var Accessible = Z.TypeClass(
     'sanctuary/Accessible',
+    readmeUrl('accessible-pseudotype'),
     [],
     function(x) { return x != null; }
   );
```
davidchambers: We should define this type in sanctuary-def. The term _pseudotype_ is archaic, and I would rather link to the documentation of a sanctuary-def `Type` value than to a paragraph which makes Accessible appear special in some way.

davidchambers: sanctuary-js/sanctuary-def#156

PR #413, opened 2017-06-14
array: generalize S.reverse

davidchambers:
:warning: This is a breaking change, as [`reverse`][1] will no longer operate on strings.

This pull request inches us closer to a goal I stated in #411: removing the [List][2] type.

[1]: https://github.com/sanctuary-js/sanctuary/tree/v0.13.2#reverse
[2]: https://github.com/sanctuary-js/sanctuary/tree/v0.13.2#list

### DIFF
--- a/index.js
+++ b/index.js
@@ -376,6 +376,12 @@
 
   /* eslint-disable indent */
 
+  //  checkTypes :: Boolean
+  var checkTypes = opts.checkTypes;
+
+  //  env :: Array Type
+  var env = opts.env;
+
   var S = {};
 
   //# create :: { checkTypes :: Boolean, env :: Array Type } -> Module
@@ -429,10 +435,10 @@
   //.
   //. See also [`env`](#env).
   S.create =
-  $.create({checkTypes: opts.checkTypes, env: defaultEnv})('create',
-                                                           {},
-                                                           [Options, $.Object],
-                                                           createSanctuary);
+  $.create({checkTypes: checkTypes, env: defaultEnv})('create',
+                                                      {},
+                                                      [Options, $.Object],
+                                                      createSanctuary);
 
   //# env :: Array Type
   //.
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -3576,6 +3573,34 @@
...
+  //.
+  //. > S.reverse(Cons(1, Cons(2, Cons(3, Nil))))
+  //. Cons(3, Cons(2, Cons(1, Nil)))
+  //.
+  //. > S.pipe([S.splitOn(''), S.reverse, S.joinWith('')], 'abc')
+  //. 'cba'
+  //. ```
+  function reverse(foldable) {
+    //  Fast path for arrays.
+    if (Array.isArray(foldable)) return foldable.slice().reverse();
```
davidchambers: I'm excited about this approach. We could do something similar if and when we define `length :: Foldable f => f a -> Integer`. :)

test/internal/List.js
```diff
@@ -0,0 +1,105 @@
+'use strict';
```
davidchambers: This file was largely copied from [sanctuary-type-classes][1].


[1]: https://github.com/sanctuary-js/sanctuary-type-classes/blob/v6.0.0/test/List.js

index.js
davidchambers: #414

test/internal/List.js
gabejohnson: This is fine for the time being. I really think a sanctuary-list package makes sense though.

index.js
gabejohnson: I agree. I'd call the function `count` or `size` though. Think of the 🌳 🌳 🌳

test/internal/List.js
davidchambers: I completely agree. It'll be satisfying to remove this file one day. :)

PR #414, opened 2017-06-14
array: add S.size

davidchambers:
This is equivalent to Haskell's [`length`][1] function.

I quite often find myself using `S.prop('length')` as `Array a -> Integer`; it will be convenient to be able to use <del>`S.length`</del> <ins>`S.size`</ins> in these situations instead. :)

[1]: http://hackage.haskell.org/package/base-4.9.1.0/docs/Data-Foldable.html#v:length

### DIFF
--- a/test/size.js
+++ b/test/size.js
@@ -1,0 +1,32 @@
+'use strict';
+
+var S = require('./internal/sanctuary');
+
+var List = require('./internal/List');
+var eq = require('./internal/eq');
+
+
+var Cons = List.Cons;
+var Nil = List.Nil;
+
+
+test('size', function() {
+
+  eq(typeof S.size, 'function');
+  eq(S.size.length, 1);
+  eq(S.size.toString(), 'size :: Foldable f => f a -> Integer');
+
+  eq(S.size([]), 0);
+  eq(S.size(['foo']), 1);
+  eq(S.size(['foo', 'bar']), 2);
+  eq(S.size(['foo', 'bar', 'baz']), 3);
+
+  eq(S.size(Nil), 0);
+  eq(S.size(Cons('foo', Nil)), 1);
+  eq(S.size(Cons('foo', Cons('bar', Nil))), 2);
+  eq(S.size(Cons('foo', Cons('bar', Cons('baz', Nil)))), 3);
+
+  eq(S.size(S.Nothing), 0);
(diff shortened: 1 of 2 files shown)


davidchambers:
Thanks for raising the question of naming, Gabe. I took the name from Haskell (and Ramda), but we're not forced to follow Haskell's lead.

Let's indicate our preferences via emoji:

  - :+1: = `length`
  - :smile: = `size`
  - :heart: = `count`

davidchambers:
@Avaq? @safareli? @scott-christopher? Do you have preferences?

Avaq:
Oh - I had cast my vote using a mobile app. Guess it didn't work. I prefer `size`, because JavaScript has conditioned me to treat `.length` like a Number.

test/size.js
```diff
@@ -0,0 +1,32 @@
...
+
+  eq(S.size([]), 0);
+  eq(S.size(['foo']), 1);
+  eq(S.size(['foo', 'bar']), 2);
+  eq(S.size(['foo', 'bar', 'baz']), 3);
+
+  eq(S.size(Nil), 0);
+  eq(S.size(Cons('foo', Nil)), 1);
+  eq(S.size(Cons('foo', Cons('bar', Nil))), 2);
+  eq(S.size(Cons('foo', Cons('bar', Cons('baz', Nil)))), 3);
```
davidchambers: Now that #413 has been merged we're able to include tests for lists. :)

safareli:
the function could also be moved in `s-type-classes`

davidchambers:
Good point, Irakli! It would be nice to decide upon a rule we can apply to tell us whether a particular function warrants inclusion in sanctuary-type-classes. Does anyone have a suggestion?

safareli:
If a function is polymorphic and it's input have some algebraic constraints, maybe.

gabejohnson:
It's interesting that given the definition of `reduce` for `Pair`, `size(Pair(1, 2)) === 1`.

davidchambers:
I like your suggestion, Irakli. :)

I will open a pull request to define `size` in sanctuary-type-classes. There are no doubt some other `S` functions whose implementations should move to `Z` too.

davidchambers:
Superseded by sanctuary-js/sanctuary-type-classes#61

PR #415, opened 2017-06-14
list: remove special treatment of -0 in access and slice functions

davidchambers:
Commit message:

> Now that `0` and `-0` are considered equal they should be interchangeable. Currently, though, `S.slice(0, -0)` and `S.slice(-0, 0)` are not equivalent.

I don't know how to feel about `S.toString(-0)`, but I'm fairly sure removing the special treatment of `-0` in `slice` and friends is a good idea.


davidchambers:
I'd :heart: a review of these changes. :)

Avaq:
:eyes:

Avaq:
Since we've removed the possibility to express *the last `n` list items* using `slice`, should we add a reference to `takeLast` in slice's documentation, guiding those who are used to using slice for that job?

And to continue that thought: maybe we can make `slice` more consistent by changing the signature to `NonZeroInteger -> NonZeroInteger -> List a -> Maybe (List a)`, and referring people to `take` if they wish to express *the first `n` list items*.

davidchambers:
> Since we've removed the possibility to express *the last `n` list items* using `slice`, should we add a reference to `takeLast` in slice's documentation, guiding those who are used to using slice for that job?

Good idea. It's not *quite* accurate to state that it's impossible to take the last `n` items using `slice`. `S.slice(-3, -0, xs)` no longer works, but one can still use `S.slice(-3, xs.length, xs)`.

> And to continue that thought: maybe we can make `slice` more consistent by changing the signature to `NonZeroInteger -> NonZeroInteger -> List a -> Maybe (List a)`, and referring people to `take` if they wish to express *the first `n` list items*.

This belongs in the realm of linters, I believe. I would love a linting tool to tell me that `S.slice(0, 3)` is better written `S.take(3)`, but I would be very surprised for `S.slice(0, 3)` to be a type error. In fact, my mental model tells me that `S.take` is equivalent to `S.slice(0)`, which in turn leads me to assume that `take` is implemented in terms of `slice`.

Avaq:
> one can still use `S.slice(-3, xs.length, xs)`

I hadn't considered that. This possibility invalidates most of my reason for suggesting the `NonZeroInteger` change. I agree with all you said. We could add the references to `take` and `takeLast` *just* because they're similar, but for no other reason. :)

PR #418, opened 2017-06-21
remove functions which take uncurried functions

### LINKED ISSUE #416: remove functions which take uncurried functions
We have several functions which take uncurried functions as arguments:

```console
$ grep '//# .*_ :: .*,' index.js
  //# reduce_ :: Foldable f => ((b, a) -> b) -> b -> f a -> b
  //# flip_ :: ((a, b) -> c) -> b -> a -> c
  //# on_ :: ((b, b) -> c) -> (a -> b) -> a -> a -> c
  //# encase2_ :: ((a, b) -> c) -> a -> b -> Maybe c
  //# encase3_ :: ((a, b, c) -> d) -> a -> b -> c -> Maybe d
  //# encaseEither2_ :: (Error -> l) -> ((a, b) -> r) -> a -> b -> Either l r
  //# encaseEither3_ :: (Error -> l) -> ((a, b, c) -> r) -> a -> b -> c -> Either l r
  //# groupBy_ :: ((a, a) -> Boolean) -> Array a -> Array (Array a)
```

Thanks to #289 we now have [`curry2`][1], [`curry3`][2], [`curry4`][3], and [`curry5`][4]. These provide a way to make uncurried functions compatible with functions such as [`encase2`][5] which require a *curried* function:

```javascript
S.encase2(S.curry2(f)) = S.encase2_(f)
```

We could remove all the `_`-suffixed functions listed above. Not only would this provide a small simplification now, it would save us from adding similar `_`-suffixed functions in future. Thoughts?

[1]: https://github.com/sanctuary-js/sanctuary/tree/v0.13.2#curry2
[2]: https://github.com/sanctuary-js/sanctuary/tree/v0.13.2#curry3
[3]: https://github.com/sanctuary-js/sanctuary/tree/v0.13.2#curry4
[4]: https://github.com/sanctuary-js/sanctuary/tree/v0.13.2#curry5
[5]: https://github.com/sanctuary-js/sanctuary/tree/v0.13.2#encase2

davidchambers:
Closes #416


gabejohnson:
Do you think exposing the `uncurry` functions has merit? They wouldn't be necessary for Sanctuary functions, but I _do_ have it in my personal utilities and use it on occasion.

davidchambers:
> Do you think exposing the `uncurry` functions has merit?

They're certainly useful functions in some contexts, but I worry that exporting these functions would muddy our message a little. Sanctuary loves curried functions. As a bonus, Sanctuary functions also behave like uncurried functions, so these need never be uncurryied. In order for uncurrying to be necessary, the user would need to be providing a curried function not defined by `def` or one of the `S.curry[2-5]` functions to a function which requires an uncurried function. I don't think we should be in the business of helping to glue together functions which exist outside our sanctuary and which don't share our philosophy. I oppose the addition of Promise-specific functions for the same reason.

tomkel:
Also, you can just keep around a reference to the uncurried version of the function, before you called `S.curry` on it.

masaeedu:
@davidchambers 

> As a bonus, Sanctuary functions also behave like uncurried functions, so these need never be uncurryied.

Worth noting that this is no longer true.

PR #420, opened 2017-07-04
lint: ensure merge commit summary will not exceed 72 characters

davidchambers:
Commit summary wrapping (as seen in <https://github.com/sanctuary-js/sanctuary-def/commit/9a1d1a6dc16646fcbf97b75eae66ddd6a5e31c1a>) is unattractive. To avoid it, I check the lengths of my branch names using `wc -c`. I'd like to automate this check. :)

If this pull request is merged I plan to see it in use for several weeks before making the same change to other Sanctuary projects. It would be efficient to make any necessary tweaks *before* duplicating this code.

### DIFF
--- a/scripts/commit-message
+++ b/scripts/commit-message
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+set -e
+
+fail() {
+  (echo && sed 's/^/! /' <<<"$1" && echo) >&2
+  exit 1
+}
+
+# Find the number of the most recently opened pull request merged so far.
+# Add 1 to this to get the smallest possible number of the next pull request,
+# then add an arbitrary buffer to allow for issues and pull requests opened
+# since the creation of the most recently opened pull request merged so far.
+# This provides an estimate of the number that will appear in the summary of
+# the merge commit so its length can be taken into account.
+message="$(
+  printf 'Merge pull request #%s from %s/%s'                                  \
+         "$(git log --grep '^Merge pull request #[1-9]' --pretty=format:%s    \
+            | awk '{ print substr($4, 2) }'                                   \
+            | sort --numeric-sort --reverse                                   \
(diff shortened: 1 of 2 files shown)


scripts/commit-message
```diff
@@ -0,0 +1,29 @@
...
+
+fail() {
+  (echo && sed 's/^/! /' <<<"$1" && echo) >&2
+  exit 1
+}
+
+# Find the number of the most recently opened pull request merged so far.
+# Add 1 to this to get the smallest possible number of the next pull request,
+# then add an arbitrary buffer to allow for issues and pull requests opened
+# since the creation of the most recently opened pull request merged so far.
```
Avaq: I don't understand.

Avaq: I should be more specific; You describe an algorithm, but I don't understand its purpose.

davidchambers: The maximum length of the branch name is a function of the number of digits in the pull request number and the number of characters in the source repository's username (e.g. `lucasschejtman`). If our search reveals…

    Merge pull request #999 from sanctuary-js/davidchambers/type-classes

we know that the next pull request number will be at least `1000`, so we expect it to be four characters in length.

Avaq: A suggestion to make the purpose of script more clear:

```diff
# Add 1 to this to get the smallest possible number of the next pull request,
# then add an arbitrary buffer to allow for issues and pull requests opened
# since the creation of the most recently opened pull request merged so far.
+# This gives an estimation of the number that will appear in the merge commit
+# message, allowing us to assert its length.
```

And possibly extract the part which actually does this, so that the comment relates directly to what follows:

```sh
# <the_comment>
pr_number_estimation=$(git log --grep ...)

message="... $pr_number_estimation ..."
```

PR #424, opened 2017-07-16
relax constraints on predicates given to S.get, S.gets, and S.parseJson

davidchambers:
Commit message:

> These functions facilitate safe interactions with untrusted values. To be maximally useful they should each take a predicate which safely handles any input (even inconsistently typed input).

### DIFF
--- a/index.js
+++ b/index.js
@@ -3600,7 +3600,7 @@
   }
   S.props = def('props', {}, [$.Array($.String), a, b], props);
 
-  //# get :: (b -> Boolean) -> String -> a -> Maybe c
+  //# get :: (Any -> Boolean) -> String -> a -> Maybe b
   //.
   //. Takes a predicate, a property name, and an object and returns Just the
   //. value of the specified object property if it exists and the value
@@ -3617,6 +3617,12 @@
   //.
   //. > S.get(S.is(Number), 'x', {})
   //. Nothing
+  //.
+  //. > S.get($.test([], $.Array($.Number)), 'x', {x: [1, 2, 3]})
+  //. Just([1, 2, 3])
+  //.
+  //. > S.get($.test([], $.Array($.Number)), 'x', {x: [1, 2, 3, null]})
+  //. Nothing
   //. ```
   function get(pred, key, x) {
     var obj = toObject(x);
@@ -3626,9 +3632,9 @@
     }
     return Nothing;
   }
-  S.get = def('get', {}, [$.Predicate(b), $.String, a, $Maybe(c)], get);
+  S.get = def('get', {}, [$.Predicate($.Any), $.String, a, $Maybe(b)], get);
 
-  //# gets :: (b -> Boolean) -> Array String -> a -> Maybe c
+  //# gets :: (Any -> Boolean) -> Array String -> a -> Maybe b
   //.
   //. Takes a predicate, a property path (an array of property names), and
   //. an object and returns Just the value at the given path if such a path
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -3638,15 +3638,21 @@
   //.
   //. > S.get(S.is(Number), 'x', {})
   //. Nothing
+  //.
```
JAForbes: The existing example are great.  But maybe a little conservative/sensible.  I think we can pre-emptively answer some questions about the behaviour of `Any` in conjunction with the `Accessible` constraint by demonstrating a completely non sensical input both within and outside an object.  Perhaps something like:

```js
S.get( S.is(Number), 'x', { x: 'purple' } ) 
//=> Nothing

S.get( S.is(Number), 'x', 'purple' ) 
// => Type Error
// value 'purple' does not satisfy the Accessible type class.
```

index.js
```diff
@@ -4037,27 +4043,30 @@
   S.parseInt =
   def('parseInt', {}, [$.Integer, $.String, $Maybe($.Integer)], parseInt_);
 
-  //# parseJson :: (a -> Boolean) -> String -> Maybe b
+  //# parseJson :: (Any -> Boolean) -> String -> Maybe a
```
JAForbes: This may be beyond the scope of this PR.  But this change could really benefit new users who want to dive in and maybe aren't sure how to bridge the gap from unsafe javascript to sanctuary.  Perhaps functions that accept `Any` and safely return a known type (via Maybe/Either or a default value) could be bundled together in a section: "Migrating from Unsafe Javascript" or "Integrating with an existing Javascript codebase", or something to that affect.  It need not be at the top of the file, but it could maybe be linked to quite early on to reassure users that there is an on ramp.

I'm thinking all the `parse` functions, `get`, `gets`, and potentially others.

index.js
```diff
@@ -3621,7 +3621,7 @@
   }
   S.props = def('props', {a: [Accessible]}, [$.Array($.String), a, b], props);
 
-  //# get :: Accessible a => (b -> Boolean) -> String -> a -> Maybe c
+  //# get :: Accessible a => (Any -> Boolean) -> String -> a -> Maybe b
   //.
   //. Takes a predicate, a property name, and an object and returns Just the
```
JAForbes: This might be hard to articulate without it becoming distracting, but it'd be nice to explain why in this instance a predicate is used instead of relying on a type definition. 

The reason I bring it up is, I've seen quite a few predicate based schema validation libraries.  Very few have their own sophisticated type system like sanctuary-def.  It might be confusing / surprising that a sanctuary function is relying on predicates alone.

*But* I think it does make sense.  It doesn't force a user of sanctuary to understand how to define their own types.  You can use `S.test` to make use of types as predicates.  And functions like `get` are helpful when you aren't completely sure what the input will be, e.g. at the edges of a system.  There are probably more reasons I'm not thinking of.  But I think it's worth while to elaborate on the subject.

index.js
davidchambers: `'purple'` is *meant* to satisfy the [Accessible][1] constraint but you are right to point out that it does not. It just occurred to me that we might be able to remove the Accessible constraint by applying `Object` to arguments in a few places. I'll investigate this possibility.


[1]: https://github.com/sanctuary-js/sanctuary/tree/v0.13.2#accessible-pseudotype

index.js
davidchambers: I :heart: this idea!

index.js
davidchambers: #425

JAForbes: > 'purple' is meant to satisfy the Accessible constraint

Yeah that's what I'm saying.  That behaviour is logical to me but could perhaps be illustrated in the examples further.

PR #425, opened 2017-07-16
remove Accessible type class

davidchambers:
I'm excited by the possibility of removing the Accessible type class. Let me know if you can think of any reasons not to do so. I have a feeling you will have a lot to say about this pull request, @Avaq.


davidchambers:
> My only concern is IE8 support, of which I once more forgot whether we aim to provide it.

We target ES5, which specified `Object.create`. :)

Issue #254, opened 2016-09-11
TS typings?


tycho01:
Gitter:
> @tycho01: afaict, there have not been attempts yet to write up TypeScript definitions for Sanctuary, correct? Has this just not been on the radar so far or might it be problematic with their type system?
> @davidchambers: More the latter I would say, @tycho01. I don't think it's currently possible to express types such as `Functor f => (a -> b) -> f a -> f b` in TypeScript. That said, @gcanti is doing things with Flow in https://github.com/gcanti/flow-static-land that I didn't think possible, so there's hope. :)

I was thinking it might be useful to put this out here, so we could consolidate questions/issues  and perhaps check for solutions with the TS team.

Potentially relevant:
* [TsMonad](https://github.com/cbowdon/TsMonad)
* SO: [functor in typescript](http://stackoverflow.com/questions/32970061/functor-in-typescript)

davidchambers:
I'm very interested to see this. It's not something I plan to work on at this stage as I'm focused on the existing modules (`sanctuary`, `sanctuary-def`, and the nascent `sanctuary-type-classes`). I'm happy to be involved in the exploration if someone would like to lead it. :)

tycho01:
Edit: some complications have been listed out by Isiah Meadows [here](https://esdiscuss.org/topic/es8-proposal-optional-static-typing#content-8):

> First, no type checker (TypeScript, Flow, or any other) can fully check the core language (most notably bind, apply, call, and Object.assign).

> Second, they both still are missing some pretty significant features required for typing common JavaScript idioms (higher kinded types for Fantasy Land and Ramda users, variadic generics for bind, call, and apply, n-ary unions for Object.assign and similar, etc.).

mnn:
Oh, that's unfortunate. I was quite excited about this library. Sadly, without (even bad) typings I cannot use it (employer won't pay me for writing typings). :cry:

davidchambers:
@mnn, I bet others would contribute if you were to start writing the TypeScript type definitions. :smile:

mnn:
Well, I am pretty sure I am not an ideal candidate - started learning TypeScript very recently and my knowledge of Sanctuary is limited to a brief browsing of your docs. :expressionless:

But I'll try! :smile: Started working on it over there - https://github.com/mnn/typings-sanctuary.

davidchambers:
That looks like a good start, @mnn!

jonahx:
[...]

PR #431, opened 2017-07-30
TypeScript support

davidchambers:
:construction: :construction_worker_woman: :construction:

Partially addresses #254

This is a work in progress, but as people are already providing useful feedback it makes sense to open a pull request for ongoing discussion.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,405 @@
+interface AwaitingTwo<A, B, R> {
+  (a: A, b: B): R
+  (a: A): (b: B) => R
+}
+
+interface AwaitingThree<A, B, C, R> {
+  (a: A, b: B, c: C): R
+  (a: A, b: B): (c: C) => R
+  (a: A): AwaitingTwo<B, C, R>
+}
+
+
+export interface Maybe<A> {
+  'fantasy-land/equals': (other: Maybe<A>) => boolean
+  'fantasy-land/map': <B>(f: (A) => B) => Maybe<B>
+  'fantasy-land/ap': <B>(other: Maybe<(A) => B>) => Maybe<B>
+  'fantasy-land/chain': <B>(f: (A) => Maybe<B>) => Maybe<B>
+  'fantasy-land/alt': (other: Maybe<A>) => Maybe<A>
+  'fantasy-land/reduce': <B>(f: (B, A) => B, x: B) => B
+  'fantasy-land/traverse': <B>(typeRep: TypeRep, f: (A) => Applicative<B>) => Applicative<Maybe<B>>
+}
+
+export interface Either<A, B> {
+  'fantasy-land/equals': (other: Either<A, B>) => boolean
+  'fantasy-land/map': <C>(f: (B) => C) => Either<A, C>
+  'fantasy-land/bimap': <C, D>(f: (A) => C, g: (B) => D) => Either<C, D>
+  'fantasy-land/ap': <C>(other: Either<A, (B) => C>) => Either<A, C>
+  'fantasy-land/chain': <C>(f: (B) => Either<A, C>) => Either<A, C>
+  'fantasy-land/alt': (other: Either<A, B>) => Either<A, B>
+  'fantasy-land/reduce': <C>(f: (C, B) => C, x: C) => C
+  'fantasy-land/traverse': <C>(typeRep: TypeRep, f: (B) => Applicative<C>) => Applicative<Either<A, C>>
+}
+
+
+type ValidNumber = number;
+type FiniteNumber = number;
+type NonZeroFiniteNumber = number;
+type Integer = number;
+type NonNegativeInteger = number;
+
+
+interface TypeRep {}
+
+
+interface Setoid<A> {
+  'fantasy-land/equals': (other: Setoid<A>) => boolean
+}
+interface Ord<A> extends Setoid<A> {
+  'fantasy-land/lte': (other: Ord<A>) => boolean
+}
+interface Semigroupoid<A, B> {
+  'fantasy-land/compose': <C>(other: Semigroupoid<B, C>) => Semigroupoid<A, C>
+}
+interface Category<A> extends Semigroupoid<A, A> {
+  constructor: {
+    'fantasy-land/id': () => Category<A>
+  }
+}
+interface Semigroup<A> {
+  'fantasy-land/concat': (other: Semigroup<A>) => Semigroup<A>
+}
+interface Monoid<A> extends Semigroup<A> {
+  constructor: {
+    'fantasy-land/empty': () => Monoid<A>
+  }
+}
+interface Functor<A> {
+  'fantasy-land/map': <B>(f: (A) => B) => Functor<B>
+}
+interface Bifunctor<A, C> extends Functor<C> {
+  'fantasy-land/bimap': <B, D>(f: (A) => B, g: (C) => D) => Bifunctor<B, D>
+}
+interface Profunctor<B, C> extends Functor<C> {
+  'fantasy-land/promap': <A, D>(f: (A) => B, g: (C) => D) => Profunctor<A, D>
+}
+interface Apply<A> extends Functor<A> {
+  'fantasy-land/ap': <B>(other: Apply<(A) => B>) => Apply<B>
+}
+interface ApplicativeTypeRep<A> {
+  'fantasy-land/of': (value: A) => Applicative<A>
+}
+interface Applicative<A> extends Apply<A> {
+  constructor: ApplicativeTypeRep<A>
+}
+interface Chain<A> extends Apply<A> {
+  'fantasy-land/chain': <B>(f: (A) => Chain<B>) => Chain<B>
+}
+interface ChainRec<A> extends Chain<A> {
+  constructor: {
+    'fantasy-land/chainRec': <B, C>(f: ($0: (A) => C, $1: (B) => C, $2: A) => ChainRec<C>, x: A) => ChainRec<B>
+  }
+}
+interface Monad<A> extends Applicative<A>, Chain<A> {}
+interface Alt<A> extends Functor<A> {
+  'fantasy-land/alt': (other: Alt<A>) => Alt<A>
+}
+interface PlusTypeRep<A> {
+  'fantasy-land/zero': () => Plus<A>
+}
+interface Plus<A> extends Alt<A> {
+  constructor: PlusTypeRep<A>
+}
+interface Alternative<A> extends Applicative<A>, Plus<A> {
+  constructor: ApplicativeTypeRep<A> & PlusTypeRep<A>
+}
+interface Foldable<A> {
+  'fantasy-land/reduce': <B>(f: (B, A) => B, x: B) => B
+}
+interface Traversable<A> extends Functor<A>, Foldable<A> {
+  'fantasy-land/traverse': <B>(typeRep: TypeRep, f: (A) => Applicative<B>) => Applicative<Traversable<B>>
+}
+interface Extend<A> extends Functor<A> {
+  'fantasy-land/extend': <B>(f: ($0: Extend<A>) => B) => Extend<B>
+}
+interface Comonad<A> extends Extend<A> {
+  'fantasy-land/extract': () => A
+}
+interface Contravariant<A> {
+  'fantasy-land/contramap': <B>(f: (B) => A) => Contravariant<B>
+}
+
+
+export const Maybe: TypeRep;
+export const Nothing: Maybe<any>;
+export function Just<A>(value: A): Maybe<A>;
+
+export const Either: TypeRep;
+export function Left<A, B>(value: A): Either<A, B>;
+export function Right<A, B>(value: B): Either<A, B>;
+
+//  Classify
+
+export function type(x: any): {
+  namespace: Maybe<string>
+  name: string
+  version: NonNegativeInteger
+};
+
+export function is(typeRep: TypeRep, x: any): boolean;
+export function is(typeRep: TypeRep): (x: any) => boolean;
+
+//  Showable
+
+export function toString(x: any): string;
+
+//  Fantasy Land
+
+export function equals<A>(x: number, y: number): boolean;
+export function equals<A>(x: number): (y: number) => boolean;
+export function equals<A>(x: Setoid<A>, y: Setoid<A>): boolean;
+export function equals<A>(x: Setoid<A>): (y: Setoid<A>) => boolean;
+
+export function lt<A>(x: Date): (y: Date) => boolean;
+export function lt<A>(x: number): (y: number) => boolean;
+export function lt<A>(x: string): (y: string) => boolean;
+export function lt<A>(x: Ord<A>): (y: Ord<A>) => boolean;
+
+export function lt_<A>(x: Date, y: Date): boolean;
+export function lt_<A>(x: Date): (y: Date) => boolean;
+export function lt_<A>(x: number, y: number): boolean;
+export function lt_<A>(x: number): (y: number) => boolean;
+export function lt_<A>(x: string, y: string): boolean;
+export function lt_<A>(x: string): (y: string) => boolean;
+export function lt_<A>(x: Ord<A>, y: Ord<A>): boolean;
+export function lt_<A>(x: Ord<A>): (y: Ord<A>) => boolean;
+
+export function lte<A>(x: Date): (y: Date) => boolean;
+export function lte<A>(x: number): (y: number) => boolean;
+export function lte<A>(x: string): (y: string) => boolean;
+export function lte<A>(x: Ord<A>): (y: Ord<A>) => boolean;
+
+export function lte_<A>(x: Date, y: Date): boolean;
+export function lte_<A>(x: Date): (y: Date) => boolean;
+export function lte_<A>(x: number, y: number): boolean;
+export function lte_<A>(x: number): (y: number) => boolean;
+export function lte_<A>(x: string, y: string): boolean;
+export function lte_<A>(x: string): (y: string) => boolean;
+export function lte_<A>(x: Ord<A>, y: Ord<A>): boolean;
+export function lte_<A>(x: Ord<A>): (y: Ord<A>) => boolean;
+
+export function gt<A>(x: Date): (y: Date) => boolean;
+export function gt<A>(x: number): (y: number) => boolean;
+export function gt<A>(x: string): (y: string) => boolean;
+export function gt<A>(x: Ord<A>): (y: Ord<A>) => boolean;
+
+export function gt_<A>(x: Date, y: Date): boolean;
+export function gt_<A>(x: Date): (y: Date) => boolean;
+export function gt_<A>(x: number, y: number): boolean;
+export function gt_<A>(x: number): (y: number) => boolean;
+export function gt_<A>(x: string, y: string): boolean;
+export function gt_<A>(x: string): (y: string) => boolean;
+export function gt_<A>(x: Ord<A>, y: Ord<A>): boolean;
+export function gt_<A>(x: Ord<A>): (y: Ord<A>) => boolean;
+
+export function gte<A>(x: Date): (y: Date) => boolean;
+export function gte<A>(x: number): (y: number) => boolean;
+export function gte<A>(x: string): (y: string) => boolean;
+export function gte<A>(x: Ord<A>): (y: Ord<A>) => boolean;
+
+export function gte_<A>(x: Date, y: Date): boolean;
+export function gte_<A>(x: Date): (y: Date) => boolean;
+export function gte_<A>(x: number, y: number): boolean;
+export function gte_<A>(x: number): (y: number) => boolean;
+export function gte_<A>(x: string, y: string): boolean;
+export function gte_<A>(x: string): (y: string) => boolean;
+export function gte_<A>(x: Ord<A>, y: Ord<A>): boolean;
+export function gte_<A>(x: Ord<A>): (y: Ord<A>) => boolean;
+
+export function min<A>(x: Date, y: Date): Date;
+export function min<A>(x: Date): (y: Date) => Date;
+export function min<A>(x: number, y: number): number;
+export function min<A>(x: number): (y: number) => number;
+export function min<A>(x: string, y: string): string;
+export function min<A>(x: string): (y: string) => string;
+export function min<A>(x: Ord<A>, y: Ord<A>): Ord<A>;
+export function min<A>(x: Ord<A>): (y: Ord<A>) => Ord<A>;
+
+export function max<A>(x: Date, y: Date): Date;
+export function max<A>(x: Date): (y: Date) => Date;
+export function max<A>(x: number, y: number): number;
+export function max<A>(x: number): (y: number) => number;
+export function max<A>(x: string, y: string): string;
+export function max<A>(x: string): (y: string) => string;
+export function max<A>(x: Ord<A>, y: Ord<A>): Ord<A>;
+export function max<A>(x: Ord<A>): (y: Ord<A>) => Ord<A>;
+
+export function id<A>(typeRep: TypeRep): (A) => A;
+export function id<A>(typeRep: TypeRep): Category<A>;
+
+export function concat<A>(x: string, y: string): string;
+export function concat<A>(x: string): (y: string) => string;
+export function concat<A>(x: Array<A>, y: Array<A>): Array<A>;
+export function concat<A>(x: Array<A>): (y: Array<A>) => Array<A>;
+export function concat<A>(x: Semigroup<A>, y: Semigroup<A>): Semigroup<A>;
+export function concat<A>(x: Semigroup<A>): (y: Semigroup<A>) => Semigroup<A>;
+
+export function empty<A>(typeRep: TypeRep): Monoid<A>;
+
+export function map<A, B>(f: (A) => B, xs: Array<A>): Array<B>;
+export function map<A, B>(f: (A) => B): (xs: Array<A>) => Array<B>;
+export function map<A, B>(f: (A) => B, functor: Functor<A>): Functor<B>;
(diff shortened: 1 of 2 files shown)


davidchambers:
Thank you very much for your feedback, @tycho01. :clap:

> binary `fantasy-land` methods lack currying

None of the methods specified by Fantasy Land is curried, so I believe this to be correct.

> Why are there `A` generics when not used?

I wanted the specialized signatures to resemble the general signatures as closely as possible, and assumed the unused type variable would simply be ignored. Is my assumption correct?

> In current TS I can't even think of a good fix to make this DRY again.

I've resigned myself to the file being at least 1000 lines. I'm not overly bothered by this, provided we having tooling in place to catch errors. Based on the comment from @ikatyang we should be able to achieve this.

> `Array` typing on `map` precludes input from alternative implementations including Immutable.js's `List`.

This is faithful to Sanctuary's behaviour. Any type with a `fantasy-land/map` method is supported, as are Array, Object, and Function for which implementations are provided by sanctuary-type-classes.

> In `parseJson<A, B>(pred: (A) => boolean, s: string): Maybe<B>`, the `A` generic is completely counter-productive.

In accordance with #424 the `A` should be changed to `any`. I don't understand the rest of your concerns about `parseJson`. I should be clear that the day I've spent working on this pull request so far is the only experience I have with TypeScript to date.

> From your `parseJson` usage example though, the predicate appears not to have parameters?

The predicate takes one argument.

index.d.ts
```diff
@@ -0,0 +1,405 @@
...
+}
+interface PlusTypeRep<A> {
+  'fantasy-land/zero': () => Plus<A>
+}
+interface Plus<A> extends Alt<A> {
+  constructor: PlusTypeRep<A>
+}
+interface Alternative<A> extends Applicative<A>, Plus<A> {
+  constructor: ApplicativeTypeRep<A> & PlusTypeRep<A>
+}
```
davidchambers: I'm unhappy with this definition. I had hoped that

```typescript
interface Alternative<A> extends Applicative<A>, Plus<A> {}
```

would suffice, but the `constructor` property requirements are seen to conflict rather than combine.

Is there an elegant solution to this problem?

tycho01: for what it's worth, I'm not currently aware of any.

tycho01:
> I wanted the specialized signatures to resemble the general signatures as closely as possible, and assumed the unused type variable would simply be ignored. Is my assumption correct?

Yeah, you're right, they are.

> Based on the comment from @ikatyang we should be able to achieve this.

Yeah, agreed, their approach looks good there. For Ramda typings it added conciliating with run-time tests over `typings-checker`, which in turn helped protect against `any` inference over just running the TS compiler.

> This is faithful to Sanctuary's behaviour.

That's what one might assume looking at the Ramda docs as well. But in practice they just iterate over arrays using `for (var i = 0; i < arr.length; i++)`, which any `ArrayLike` supports in exposing `length` and the numerical index. Won't hold if you're actually using `Array.prototype` though obviously.

> In accordance with #424 the A should be changed to any. I don't understand the rest of your concerns about parseJson.

`B` is only used in the return type position, meaning the only way for it to get a type is if the user were to explicitly specify it in the generics. At that point, the user may find it an annoyance they'd be made to manually write out an additional type for the predicate as well. But yeah, if you make it `any` and ditch the generic, that'd solve it.

I'll concede that may make for a trade-off between TS UX and compliance with docs. I guess that was a reason I kinda gave up on trying to make them match for the Ramda typings -- I never really tried since the guys before me hadn't either, but I fear it may be tough to always keep the number of generics equal. These two examples aside, in TS I've had the impression it's easy to end up with more. Maybe for UX here having it say e.g. `pred: (a: any) => boolean` could do to invoke the `a` connection with the docs, I dunno.

> I should be clear that the day I've spent working on this pull request so far is the only experience I have with TypeScript to date.

Well, I'd gotten type-classes wrong on my try, so I think you're doing pretty good! 😃

index.d.ts
```diff
@@ -0,0 +1,490 @@
...
+export function equals<A>(x: string): (y: string) => boolean;
+export function equals<A>(x: Array<A>, y: Array<A>): boolean;
+export function equals<A>(x: Array<A>): (y: Array<A>) => boolean;
+export function equals<A>(x: IArguments, y: IArguments): boolean;
+export function equals<A>(x: IArguments): (y: IArguments) => boolean;
+export function equals<A>(x: Error, y: Error): boolean;
+export function equals<A>(x: Error): (y: Error) => boolean;
+//  XXX: This is too general. Can we match "plain" objects only?
+//  export function equals<A>(x: Object, y: Object): boolean;
+//  export function equals<A>(x: Object): (y: Object) => boolean;
```
davidchambers: Do you have a suggestion, @tycho01?

davidchambers:
> > This is faithful to Sanctuary's behaviour.
>
> That's what one might assume looking at the Ramda docs as well.

Sanctuary has no notion of "array-like" objects:

```javascript
S.map(S.toUpper, {'0': 'foo', '1': 'bar', '2': 'baz', length: 3});
//
//   map :: Functor f => (a -> b) -> f a -> f b
//                                   ^^^
//                                    1
//
//   1)  {"0": "foo", "1": "bar", "2": "baz", "length": 3} :: Object, StrMap ???
//
```

If one evaluates this expression with Sanctuary's run-time type checking disabled, `S.map` will map over what it assumes to be a value of type `StrMap String`, and a run-time exception will be thrown when attempting to invoke `toUpperCase` on `3`.

index.d.ts
tycho01: There's `object` and `{}`. I haven't even bothered to look into their nuances. `Object` appears not used as it'd match almost anything.

On DRY'ing... `equals<A>(x: any, y: any): boolean`? 😅
Similar to the `lt` / `gt`; wanted to say there's `equals<A extends null | undefined | boolean | ...>(x: A, y: A): boolean` (or even drop the constraint), but... generics don't just fail with literals here, it just does nothing useful. At that point it'll just be like, `equals(123, "foo")` totally works for `A`: `any` (or for `number | string`).
So technically without run-time type checks `equals(123, "foo")` may still run and produce false I guess?
I'd wanna go for a type-level type check, but those aren't possible yet. Failing that, it's your solution or any. If you'd consider not doing run-time errors for different types, with your approach you might even add a `equals<A>(x: any, y: any): false` fallback at the bottom.

davidchambers: > On DRY'ing... `equals<A>(x: any, y: any): boolean`?

This is feasible for `S.equals` but not for `S.map` and other polymorphic functions which can operate on values of type `StrMap a` (plain objects with consistently typed values).

> At that point it'll just be like, `equals(123, "foo")` totally works for `A`: `any` (or for `number | string`).

That's unfortunate. Although sanctuary-def provides [`$.Any`][1] this is not considered to be a "unifying" type. It's a pity that TypeScript can't express types such as `a -> a -> a`.

> If you'd consider not doing run-time errors for different types, with your approach you might even add a `equals<A>(x: any, y: any): false` fallback at the bottom.

Is it possible to include implementations in a `.d.ts` file? Keep in mind that the implementation is still written in a `.js` file.


[1]: https://github.com/sanctuary-js/sanctuary-def#Any

index.d.ts
tycho01: Actually, you could probably abstract it into a pattern I guess. But yeah only really pays off given multiple occurrences of this.

index.d.ts
tycho01: What you could probably do is store the type using the generic as a generic type, then type the function using an interface of overloads populated by instances of this generic type, which would... maybe look a bit cleaner. I never tried anything like it though.

`type Equals<A> = (x: any, y: any): boolean;`  then like `interface equals {
  Equals<string>;
  ...
}`

I'm on phone atm so can't test well, pretty sure it'd fail but it's the best improvement I can come up with. So probably just gonna be the way you got it now then.

tycho01: For map consider checking the definition used in the Ramda typings.
What do you mean on the .d.ts vs. js?
Putting JS in a .d.ts file, I forgot if I tried, so don't take my word for it, but it probably fails.
Having js in the .d.ts seems not to make much sense though, so I'm probably misinterpreting what you mean by implementation.

index.d.ts
```diff
@@ -0,0 +1,490 @@
+interface AwaitingTwo<A, B, R> {
+  (a: A, b: B): R
+  (a: A): (b: B) => R
+}
+
+interface AwaitingThree<A, B, C, R> {
+  (a: A, b: B, c: C): R
+  (a: A, b: B): (c: C) => R
+  (a: A): AwaitingTwo<B, C, R>
+}
```
Avaq: I've seen these before somewhere. :thinking:

index.d.ts
```diff
@@ -0,0 +1,490 @@
...
+export function empty<A>(typeRep: TypeRep): Monoid<A>;
+
+export function map<A, B>(f: (A) => B, xs: Array<A>): Array<B>;
+export function map<A, B>(f: (A) => B): (xs: Array<A>) => Array<B>;
+export function map<A, B>(f: (A) => B, functor: Functor<A>): Functor<B>;
+export function map<A, B>(f: (A) => B): (functor: Functor<A>) => Functor<B>;
+
+export function bimap<A, B, C, D>(f: (A) => B, g: (C) => D, bifunctor: Bifunctor<A, C>): Bifunctor<B, D>;
+export function bimap<A, B, C, D>(f: (A) => B, g: (C) => D): (bifunctor: Bifunctor<A, C>) => Bifunctor<B, D>;
+export function bimap<A, B, C, D>(f: (A) => B): AwaitingTwo<(C) => D, Bifunctor<A, C>, Bifunctor<B, D>>;
```
ikatyang: Notice that TS won't infer types across the function call, this kind of function should be something like:

```diff
-export function bimap<A, B, C, D>(f: (a: A) => B): AwaitingTwo<(c: C) => D, Bifunctor<A, C>, Bifunctor<B, D>>;
+export function bimap<A, B>(f: (A) => B): {
+  <C, D>(x: (c: C) => D, y: Bifunctor<A, C>): Bifunctor<B, D>;
+  <C, D>(x: (c: C) => D): (y: Bifunctor<A, C>) => Bifunctor<B, D>;
+};
```

Otherwise the `C` and `D` will be inferred as `{}`, since they are not exised in the first function call.

davidchambers: Thank you, @ikatyang! I will make this change and similar changes then push another commit. I'd love you to review it. :)

index.d.ts
Avaq: ...and they may be problematic: https://github.com/sanctuary-js/sanctuary/pull/431#discussion_r130345580 :(

index.d.ts
ikatyang: I have some tools to restructure this kind of definition in my [types-ramda](https://github.com/ikatyang/types-ramda), but it does not expose yet, I'll expose them ASAP to make restructure FP definitions easily.

- simple curry
  - [template](https://github.com/ikatyang/types-ramda/blob/master/templates/append.d.ts) -> [generated](https://github.com/ikatyang/types-ramda/blob/dist-simple/src/append.d.ts)
- various curry
  - [template](https://github.com/ikatyang/types-ramda/blob/master/templates/map.d.ts) -> [generated](https://github.com/ikatyang/types-ramda/blob/dist-simple/src/map.d.ts)

ikatyang: And it's recommended to setup a test first, since our eyes can't catch all the bugs, but tests can.

davidchambers: I've been playing with the fully applied form of `S.bimap` to understand your feedback, but TypeScript seems oblivious to type errors:

```typescript
S.bimap(S.toUpper, Math.sqrt, S.Left(42))
```

Am I doing something wrong, or is TypeScript simply unifying everything to `any`?

ikatyang: Should it be `S.Left('str')`?

See [TS Playground](http://www.typescriptlang.org/play/index.html#src=interface%20Functor%3CA%3E%20%7B%0A%20%20'fantasy-land%2Fmap'%3A%20%3CB%3E(f%3A%20(a%3A%20A)%20%3D%3E%20B)%20%3D%3E%20Functor%3CB%3E%0A%7D%0A%0Ainterface%20Bifunctor%3CA%2C%20C%3E%20extends%20Functor%3CC%3E%20%7B%0A%20%20'fantasy-land%2Fbimap'%3A%20%3CB%2C%20D%3E(f%3A%20(a%3A%20A)%20%3D%3E%20B%2C%20g%3A%20(c%3A%20C)%20%3D%3E%20D)%20%3D%3E%20Bifunctor%3CB%2C%20D%3E%0A%7D%0A%0Adeclare%20function%20toUpper(s%3A%20string)%3A%20string%3B%0A%0Adeclare%20function%20Left%3CA%2C%20B%3E(value%3A%20A)%3A%20Either%3CA%2C%20B%3E%3B%0A%0Ainterface%20Either%3CA%2C%20B%3E%20%7B%0A%20%20'fantasy-land%2Fequals'%3A%20(other%3A%20Either%3CA%2C%20B%3E)%20%3D%3E%20boolean%0A%20%20'fantasy-land%2Fmap'%3A%20%3CC%3E(f%3A%20(b%3A%20B)%20%3D%3E%20C)%20%3D%3E%20Either%3CA%2C%20C%3E%0A%20%20'fantasy-land%2Fbimap'%3A%20%3CC%2C%20D%3E(f%3A%20(a%3A%20A)%20%3D%3E%20C%2C%20g%3A%20(b%3A%20B)%20%3D%3E%20D)%20%3D%3E%20Either%3CC%2C%20D%3E%0A%20%20'fantasy-land%2Fap'%3A%20%3CC%3E(other%3A%20Either%3CA%2C%20(b%3A%20B)%20%3D%3E%20C%3E)%20%3D%3E%20Either%3CA%2C%20C%3E%0A%20%20'fantasy-land%2Fchain'%3A%20%3CC%3E(f%3A%20(b%3A%20B)%20%3D%3E%20Either%3CA%2C%20C%3E)%20%3D%3E%20Either%3CA%2C%20C%3E%0A%20%20'fantasy-land%2Falt'%3A%20(other%3A%20Either%3CA%2C%20B%3E)%20%3D%3E%20Either%3CA%2C%20B%3E%0A%20%20'fantasy-land%2Freduce'%3A%20%3CC%3E(f%3A%20(c%3A%20C%2C%20b%3A%20B)%20%3D%3E%20C%2C%20x%3A%20C)%20%3D%3E%20C%0A%20%20'fantasy-land%2Ftraverse'%3A%20%3CC%3E(typeRep%3A%20TypeRep%2C%20f%3A%20(b%3A%20B)%20%3D%3E%20Applicative%3CC%3E)%20%3D%3E%20Applicative%3CEither%3CA%2C%20C%3E%3E%0A%7D%0A%0Ainterface%20Apply%3CA%3E%20extends%20Functor%3CA%3E%20%7B%0A%20%20'fantasy-land%2Fap'%3A%20%3CB%3E(other%3A%20Apply%3C(a%3A%20A)%20%3D%3E%20B%3E)%20%3D%3E%20Apply%3CB%3E%0A%7D%0Ainterface%20ApplicativeTypeRep%3CA%3E%20%7B%0A%20%20'fantasy-land%2Fof'%3A%20(value%3A%20A)%20%3D%3E%20Applicative%3CA%3E%0A%7D%0Ainterface%20Applicative%3CA%3E%20extends%20Apply%3CA%3E%20%7B%0A%20%20constructor%3A%20ApplicativeTypeRep%3CA%3E%0A%7D%0A%0Ainterface%20TypeRep%20%7B%7D%0A%0Adeclare%20function%20bimap%3CA%2C%20B%2C%20C%2C%20D%3E(f%3A%20(a%3A%20A)%20%3D%3E%20B%2C%20g%3A%20(c%3A%20C)%20%3D%3E%20D%2C%20bifunctor%3A%20Bifunctor%3CA%2C%20C%3E)%3A%20Bifunctor%3CB%2C%20D%3E%3B%0Adeclare%20function%20bimap%3CA%2C%20B%2C%20C%2C%20D%3E(f%3A%20(a%3A%20A)%20%3D%3E%20B%2C%20g%3A%20(c%3A%20C)%20%3D%3E%20D)%3A%20(bifunctor%3A%20Bifunctor%3CA%2C%20C%3E)%20%3D%3E%20Bifunctor%3CB%2C%20D%3E%3B%0Adeclare%20function%20bimap%3CA%2C%20B%3E(f%3A%20(a%3A%20A)%20%3D%3E%20B)%3A%20%7B%0A%20%20%3CC%2C%20D%3E(x%3A%20(c%3A%20C)%20%3D%3E%20D%2C%20y%3A%20Bifunctor%3CA%2C%20C%3E)%3A%20Bifunctor%3CB%2C%20D%3E%3B%0A%20%20%3CC%2C%20D%3E(x%3A%20(c%3A%20C)%20%3D%3E%20D)%3A%20(y%3A%20Bifunctor%3CA%2C%20C%3E)%20%3D%3E%20Bifunctor%3CB%2C%20D%3E%3B%0A%7D%3B%0A%0Aconst%20result%20%3D%20bimap(toUpper%2C%20Math.sqrt%2C%20Left(42))%0A).

davidchambers: > Shouldn't it be `S.Left('str')`?

Exactly. `S.bimap(S.toUpper, Math.sqrt, S.Left(42))` should fail to compile, but in fact it compiles successfully only to throw an exception at run time.

This requires further investigation. :)

ikatyang: TS compiler always emits codes unless you set `noEmitOnError` to true. And it indeed fail to compile to me, what do you mean compiles successfully?

image

davidchambers: Perhaps you could pull this branch and run these commands:

```console
$ mkdir experiment
$ cd experiment
$ echo $'import * as S from "..";\n\nS.bimap(S.toUpper, Math.sqrt, S.Left(42));' >index.ts
$ tsc index.ts
$ cat index.js
"use strict";
exports.__esModule = true;
var S = require("..");
S.bimap(S.toUpper, Math.sqrt, S.Left(42));
```

Do you observe the same behaviour?

ikatyang: OK, I know what you mean. The current branch haven't updated the `(A) => B` to `(a: A) => B`, so that `A` is considered a variable name and its type is `any`.

ikatyang: You can copy code from the playground link I mentioned above, and just put it in `test.ts` then `tsc test.ts`, you'll see the error message.

davidchambers: > The current branch haven't updated the `(A) => B` to `(a: A) => B`, so that `A` is considered a variable name and its type is `any`.

Aha! I overlooked that in your first comment. Thank you for being patient with me. :)

index.d.ts
```diff
@@ -128,6 +133,14 @@ export const Either: TypeRep;
 export function Left<A, B>(value: A): Either<A, B>;
 export function Right<A, B>(value: B): Either<A, B>;
 
+//  Placeholder
+
+interface Placeholder {
+  '@@functional/placeholder': boolean
```
Avaq: I think you could simply use `true` instead of `boolean`. Wouldn't want Sanctuary to treat my value as a placeholder even though I *explicitly* stated that `@@function/placeholder` is `false`!

davidchambers: `true` is not a type, though, is it?

ikatyang: `true` is a literal type.

literal type: `boolean` (`true`, `false`), `number`(`1`, `2`, etc.) and `string` (`'str1'`, `'str2'`, etc.)

davidchambers: If I change `boolean` to `true`, this code example does not type check:

```typescript
import {concat} from 'sanctuary';

const _ = {'@@functional/placeholder': true};

concat(_, 'def')('abc');
```

Do you know what's wrong, @ikatyang? I'd love to use `true` if possible. :)

ikatyang: For objects, TS always widen their type to non-literal, so in this case you have to assert it to `true` manually, e.g.

```ts
const _ = {'@@functional/placeholder': true as true}; //=> {'@@functional/placeholder': true}
```

```ts
const _ = {'@@functional/placeholder': true}; //=> {'@@functional/placeholder': boolean}
```

davidchambers: That works! Thank you for the explanation. I'm enjoying learning from you. :)

index.d.ts
ikatyang: @davidchambers  I've done my [dts-element-fp](https://github.com/ikatyang/dts-element-fp#readme) for restructuring curried function definitions (placeholder available), might be useful to you.

miangraham:
Interested party here. Any ETA on a merge? Any specific help needed?

I'm looking at giving Sanctuary a spin from a TS project and adding it to an "How to Do Immutable Typescript" starter package I'm working on. Can work from @davidchambers' fork for a bit if I need to but it's behind master now that this has been quiet for about a month. Thoughts appreciated.

davidchambers:
> Interested party here. Any ETA on a merge? Any specific help needed?

Thank you for taking an interest, @miangraham, and for submitting a helpful pull request. I don't know when this pull request will be merged, as I'm focused on sanctuary-js/sanctuary-scripts#1 at present.

This pull request is not blocked, as far as I recall, but there's still a lot to do:

  - add type definitions for remaining functions;
  - programmatically assert that the type definitions don't introduce any test failures; and
  - add documentation—or documentation links—alongside type definitions.

@miangraham, if you're interested in contributing to this effort I will give you write access so you can push commits directly to `davidchambers/typescript`. :)

miangraham:
> * programmatically assert that the type definitions don't introduce any test failures

@davidchambers This seems like good bang for buck and a spitball should be super easy. Something like https://github.com/sanctuary-js/sanctuary/compare/davidchambers/typescript...miangraham:davidchambers/typescript could hit index.d.ts with strict type checking, tslint, and some new type-level tests ("this line should fail to type check in this way") from your existing hooks.

Can def chip in some type defs too. Feel free to add me and I can try pushing a few things your way over the weekend.

davidchambers:
> Feel free to add me and I can try pushing a few things your way over the weekend.

Excellent! Feel free to push to this branch. Please don't `--force` push as it's now a collaborative branch.

miangraham:
The last CircleCI pass included these:
```
make lint
(type check index.d.ts but not tests b/c tests have type errors by design)
node_modules/.bin/tslint \
	  --type-check \
	  --project \
	  tsconfig.json
node_modules/.bin/tslint \
	  --project \
	  test/typescript/tsconfig.json

make test
node_modules/.bin/typings-checker --allow-expect-error --project test/typescript/tsconfig.json test/typescript/*.ts
test/typescript/map.ts: 2 / 2 checks passed.
```

miangraham:
For reference, current obvious index.d.ts TODOs:

- [ ] fantasyland/extend
- [ ] fantasyland/extract
- [ ] fantasyland/contramap
- [ ] fantasyland/filter
- [ ] fantasyland/filterM
- [ ] fantasyland/takeWhile
- [ ] fantasyland/dropWhile
- [ ] Function
- [ ] Composition
- [ ] Maybe (has interface, needs functions)
- [ ] Either
- [ ] Logic
- [ ] List
- [ ] Array
- [ ] Object

I got filter and tests working for Array<A> without issue then bounced off the interface for non-arrays (I'm new to Fantasy Land). Will come back to it. In the meantime, Function and Logic look straightforward so I'll try grinding through those for a win.

tycho01:
Monads aside there might well be overlap with ramda, so feel to take from there what you need.

miangraham:
Many-param `curryN` plus placeholders is a bit nuts. Automation may be the only way to preserve sanity. `curry2` and `curry3` are in for now, without placeholders in their direct parameters yet. Those two are at least straightforward to write manually, especially with AwaitingTwo/Three interfaces already in place.

Logic is pretty much done. Shout if any of those look off.

I'll push `curryN` a bit further, then I'm thinking Composition -> Object -> List -> Array.

tycho01:
@miangraham: if I can finish my current [PRs](https://github.com/Microsoft/TypeScript/pulls/tycho01) overloads may no longer be needed (curry PoC at https://github.com/Microsoft/TypeScript/issues/5453#issuecomment-309035811).
I presume you already saw, but Ramda typings have a `curryN` with template [here](https://github.com/types/npm-ramda/blob/master/templates/curryN.c.ts), and code-gen'd results for branches [`dist-simple`](https://github.com/types/npm-ramda/blob/dist-simple/src/curryN.d.ts), [`dist-placeholder`](https://github.com/types/npm-ramda/blob/dist-placeholder/src/curryN.d.ts), [`dist`](https://github.com/types/npm-ramda/blob/dist/src/curryN.d.ts).

miangraham:
@tycho01 Ooo, very nice. I'd seen Ramda's generation approach but didn't know about any of the TS work on variadic kinds. That looks like an awesome shortcut. Existing interfaces with multiple fixed arities would still need overloads but they could just be typed in terms of the variadic version, saving a ton of boilerplate. I think. `arity(5)(curry)`, etc. Super cool.

tycho01:
@miangraham: Yeah. Ideally type programming should be no worse than the expression level. Heck, in the ideal scenario the standard library would be well-typed enough for it to just infer these return types rather than us having to do them manually. Much less with clunky overload codegen.

davidchambers:
Thanks for the contributions, @miangraham. Keep them coming. :)

You may have an opinion on #438. If we decide to abandon Ramda-style currying our TypeScript type definitions will no longer be so unwieldy.

miangraham:
I'm currently erring towards wide and shallow treatment to try and get an MVP/first pass done for the whole library. Means skimping on currying where it's not easy while leaving as many explicit TODOs as possible, but could result in getting something out fairly soon. I'm thinking this month if we're lucky?

Going breadth-first also enables fundamental changes in approach. If the TS defs really want to bite off whole-hog fancy currying, that probably means code generation, a separate repo to cordon off the infrastructure to generate the final .d.ts, more exhaustive testing, etc etc etc. Plenty of great Ramda examples to lean on there when it's needed.

In the meantime it'd be nice to get users just statically typing the basics to see how it goes!

davidchambers:
> In the meantime it'd be nice to get users just statically typing the basics to see how it goes!

I agree. Let's focus on the curried signatures initially. :)

PR #432, opened 2017-08-06
number: flip S.div argument order

### LINKED ISSUE #430: S.div argument order should match that of S.sub
Currently we have:

```haskell
sub :: FiniteNumber -> (FiniteNumber -> FiniteNumber)
sub_ :: FiniteNumber -> FiniteNumber -> FiniteNumber
```

I suggest we update [`S.div`][1] to match, giving:

```haskell
div :: FiniteNumber -> (FiniteNumber -> FiniteNumber)
div_ :: FiniteNumber -> FiniteNumber -> FiniteNumber
```

The _divide by two_ function would then be `S.div(2)` rather than `S.div(S.__, 2)`. :)

[1]: https://github.com/sanctuary-js/sanctuary/tree/v0.13.2#div

davidchambers:
Closes #430

### DIFF
--- a/index.js
+++ b/index.js
@@ -3827,24 +3827,49 @@
       [f($.FiniteNumber), $.FiniteNumber],
       product);
 
-  //# div :: FiniteNumber -> NonZeroFiniteNumber -> FiniteNumber
+  //# div :: NonZeroFiniteNumber -> (FiniteNumber -> FiniteNumber)
+  //.
+  //. Takes a non-zero finite number `n` and returns the _divide by `n`_
+  //. function.
+  //.
+  //. See also [`div_`](#div_).
+  //.
+  //. ```javascript
+  //. > S.map(S.div(2), [0, 1, 2, 3])
+  //. [0, 0.5, 1, 1.5]
+  //. ```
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -3827,24 +3827,49 @@
...
+  S.div_ =
+  def('div_',
+      {},
+      [$.FiniteNumber, $.NonZeroFiniteNumber, $.FiniteNumber],
+      div_);
 
   //# pow :: FiniteNumber -> (FiniteNumber -> FiniteNumber)
   //.
-  //. Takes a finite number `n` and returns the _`n` squared_ function.
+  //. Takes a finite number `n` and returns the _power of `n`_ function.
```
Avaq: Nit: I would commit this separately, allowing the "flip `div`" to be treated (eg. reverted) separately from this unrelated fix.

davidchambers: I agree that atomic commits are desirable. In this case I consulted the `pow` description when writing the `div` description, which led me to notice this error. I'd argue that this fix is _indirectly_ related rather than unrelated. ;)

Issue #434, opened 2017-08-21
Add insert for Objects


ucodescott:
Would be great to have an `insert` function similar to [Ramda's `assoc`](http://ramdajs.com/docs/#assoc). I like the name `insert` better as that seems more clear when working with other dictionary APIs ([Elm](http://package.elm-lang.org/packages/elm-lang/core/latest/Dict#insert) and [Haskell](http://hackage.haskell.org/package/containers-0.5.10.2/docs/Data-Map-Lazy.html#v:insert)).

The type signature might be something like:
```haskell
insert :: String -> Any -> Object -> Object
```

An implementation might look like the following:
```javascript
function insert(key, value, object) {
	var newEntry = {};
	newEntry[key] = value;
	return Object.assign(newEntry, object);
}
S.insert = def(‘insert’, {}, [$.String, $.Any, $.Object, $.Object], insert);
```

If an insert function makes sense for Sanctuary, I'm happy to create a pull request.

Avaq:
Should `insert` operate on `StrMap` rather than `Object`?

```sh
insert :: String -> a -> StrMap a -> StrMap a
```

davidchambers:
I'm very much in favour of this addition. I agree with @Avaq that we should stick to a strict type.

samueller:
I like `StrMap`, surprised I missed that from the existing Object functions. Seems like the implementation I wrote above would still work.

davidchambers:
> Seems like the implementation I wrote above would still work.

It would, although we try to stick to ES5 language features and APIs to support lots of environments without polyfills.

I'd love to see a pull request, Scott. :)

davidchambers:
We could utilize `Z.concat` to avoid making decisions about inherited properties:

```javascript
function insert(key, val, obj) {
  var result = Z.concat(obj, {});
  result[key] = val;
  return result;
}
```

samueller:
BTW, @ucodescott is me. I accidentally created this issue under the GitHub account I use to teach others with. But GitHub wouldn't let me edit the account or delete the issue... Didn't want anyone to think I hijacked this issue from someone else :).

masaeedu:
Would this be identical to `const insert = k => v => S.concat({ [k]: v })`? Maybe having a separate `k => v => { [k]: v }` primitive as well would be useful. This can then be composed with `concat`.

davidchambers:
[...]

PR #435, opened 2017-08-22
strmap: add S.insert and S.remove

samueller:
As discussed in issue #434.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3757,6 +3757,51 @@
   }
   S.singleton = def('singleton', {}, [$.String, a, $.StrMap(a)], singleton);
 
+  //# insert :: String -> a -> StrMap a -> StrMap a
+  //.
+  //. Takes a string, a value of any type, and a string map, and returns a
+  //. string map comprising all the entries of the given string map plus the
+  //. entry specified by the first two arguments (which takes precedence).
+  //.
+  //. Equivalent to Haskell's `insert` function. Similar to Clojure's `assoc`
+  //. function.
+  //.
+  //. ```javascript
+  //. > S.insert('c', 3, {a: 1, b: 2})
+  //. {a: 1, b: 2, c: 3}
+  //.
+  //. > S.insert('a', 4, {a: 1, b: 2})
+  //. {a: 4, b: 2}
+  //. ```
+  function insert(key, val, strMap) {
+    return Z.concat(strMap, singleton(key, val));
+  }
+  S.insert =
+  def('insert', {}, [$.String, a, $.StrMap(a), $.StrMap(a)], insert);
+
+  //# remove :: String -> StrMap a -> StrMap a
+  //.
+  //. Takes a string and a string map, and returns a string map comprising all
+  //. the entries of the given string map except the one whose key matches the
+  //. given string (if such a key exists).
+  //.
+  //. Equivalent to Haskell's `delete` function. Similar to Clojure's `dissoc`
+  //. function.
+  //.
+  //. ```javascript
+  //. > S.remove('c', {a: 1, b: 2, c: 3})
+  //. {a: 1, b: 2}
+  //.
+  //. > S.remove('c', {})
+  //. {}
+  //. ```
+  function remove(key, strMap) {
+    var result = Z.concat(strMap, {});
+    delete result[key];
+    return result;
+  }
+  S.remove = def('remove', {}, [$.String, $.StrMap(a), $.StrMap(a)], remove);
+
   //# keys :: StrMap a -> Array String
   //.
   //. Returns the keys of the given string map, in arbitrary order.
--- a/test/insert.js
+++ b/test/insert.js
@@ -1,0 +1,25 @@
+'use strict';
+
+var jsc = require('jsverify');
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+
+
+test('insert', function() {
+
+  eq(typeof S.insert, 'function');
+  eq(S.insert.length, 3);
+  eq(S.insert.toString(), 'insert :: String -> a -> StrMap a -> StrMap a');
+
+  eq(S.insert('a', 1, {}), {a: 1});
+  eq(S.insert('b', 2, {a: 1}), {a: 1, b: 2});
+  eq(S.insert('c', 3, {a: 1, b: 2, c: 4}), {a: 1, b: 2, c: 3});
+
+  jsc.assert(jsc.forall(jsc.string, jsc.number, jsc.dict(jsc.number), function(key, val, map) {
+    var insert = S.insert(key, val);
+    return S.equals(insert(insert(map)), insert(map));
+  }), {tests: 1000});
+
+});
--- a/test/remove.js
+++ b/test/remove.js
@@ -1,0 +1,25 @@
+'use strict';
+
+var jsc = require('jsverify');
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+
+
+test('remove', function() {
+
+  eq(typeof S.remove, 'function');
+  eq(S.remove.length, 2);
+  eq(S.remove.toString(), 'remove :: String -> StrMap a -> StrMap a');
+
+  eq(S.remove('a', {}), {});
+  eq(S.remove('b', {a: 1}), {a: 1});
+  eq(S.remove('c', {a: 1, b: 2, c: 3}), {a: 1, b: 2});
+
+  jsc.assert(jsc.forall(jsc.string, jsc.dict(jsc.number), function(key, map) {
+    var remove = S.remove(key);
+    return S.equals(remove(remove(map)), remove(map));
+  }), {tests: 1000});
+
+});


gabejohnson:
We may want to consider adding `delete` and `update` as well while we're at it.

davidchambers:
Lovely pull request, @samueller! I see that you took care to follow the project's conventions. :tada:

> We may want to consider adding `delete` and `update` as well while we're at it.

I agree. I'd been thinking [`delete`][1] would be a natural inclusion in this pull request. [`update`][2] is new to me, but it's closely related to the other two functions.

Scott, please add `delete` to this pull request. You're welcome to add `update` too, but I don't mind leaving that for another pull request.


[1]: https://hackage.haskell.org/package/containers-0.5.10.2/docs/Data-Map-Internal.html#v:delete
[2]: https://hackage.haskell.org/package/containers-0.5.10.2/docs/Data-Map-Internal.html#v:update

index.js
```diff
@@ -3700,6 +3700,26 @@
   S.pairs =
   def('pairs', {}, [$.StrMap(a), $.Array($.Pair($.String, a))], pairs);
 
+  //# insert :: String -> a -> StrMap a -> StrMap a
+  //.
+  //. Inserts a given property key and value into a shallow clone of a given
+  //. string map. Overrides the property if it already exists.
```
davidchambers: Let's avoid using the term _property_ in relation to string maps. The fact that string maps are unadorned objects is an implementation detail in most contexts.

Adding the following would help those coming from Ramda who are likely to expect the function to be named `assoc`:

_Equivalent to Haskell's `insert` function. Similar to Clojure's `assoc` function._

gabejohnson:
@davidchambers I'm rethinking `update`. It should probably go in a separate PR with `alter`.

They both rely on `Maybe` whereas `insert` and `delete` do not.

`alter` can be implemented in terms of `insert` and `delete` and `update` in terms of `alter`.

davidchambers:
> It should probably go in a separate PR with `alter`.

Okay. `insert` and `delete` should still be inserted into the project atomically. :)

index.js
samueller: Makes sense. I'll remove "property" and add your 2 sentences.

samueller:
@davidchambers, @gabejohnson. So I'll add `delete` to this PR and nothing else, correct?

davidchambers:
> So I'll add `delete` to this PR and nothing else, correct?

:thumbsup:

samueller:
I created the `S.delete` function, but ESLint is reporting an error when I run `make test lint`:
```
/Users/scott/Documents/code/sanctuary/index.js
  3754:5  error  .delete is a syntax error  dot-notation

✖ 1 problem (1 error, 0 warnings)

make: *** [lint] Error 1
```
I'm guessing this is a problem because `delete` is a keyword in JavaScript. Do we modify the ESLint configuration or change the name of this function?

gabejohnson:
We have to change it. Calling it `delete` will mean that no one will be able to use it by that name unqualified.

We could call it `remove`, `deleteFrom`, `deleteEntry`, or what have you.

I apologize. This should have occurred to me before I was suggesting it.

samueller:
I like `remove`.

index.js
```diff
@@ -3700,6 +3700,60 @@
...
+  S.insert =
+  def('insert', {}, [$.String, a, $.StrMap(a), $.StrMap(a)], insert);
+
+  //# remove :: String -> StrMap a -> StrMap a
+  //.
+  //. Removes a given key and its value from a shallow clone of a given
+  //. string map. The original string map is returned if the key doesn't exist.
+  //.
+  //. Equivalent to Haskell's `delete` function. Similar to Clojure's `dissoc`
+  //. functiom.
```
davidchambers: `s/functiom/function/`

index.js
```diff
@@ -3700,6 +3700,60 @@
...
+  //. {a: 1, b: 2}
+  //.
+  //. > S.remove('c', {a: 1, b: 2})
+  //. {a: 1, b: 2}
+  //.
+  //. > S.remove('c', {})
+  //. {}
+  //. ```
+  function remove(key, strMap) {
+    if (key in strMap) {
```
davidchambers: This is unsafe due to inherited properties. `'toString' in {}`, for example, evaluates to `true`.

Is this implementation suitable?

```javascript
var result = Z.concat(strMap, {});
delete result[key];
return result;
```

index.js
samueller: good catch!

index.js
samueller: Wasn't aware of the inherited properties issue. The initial thought was to use `concat` in the way you wrote. But it bothered me to add a property to an object and then delete it. In addition, I liked the efficiency of Haskell's `delete` function where it returns the original `Map` if the key didn't exist.

What about `if (strMap.hasOwnProperty(key))`?

test/remove.js
```diff
@@ -0,0 +1,28 @@
...
+
+  eq(typeof S.remove, 'function');
+  eq(S.remove.length, 2);
+  eq(S.remove.toString(), 'remove :: String -> StrMap a -> StrMap a');
+
+  function testOriginalReturned(key, strMap) {
+    var strMapAfterDeletion = S.remove(key, strMap);
+    eq(strMap, strMapAfterDeletion);
+    assert.strictEqual(strMap, strMapAfterDeletion);
+  }
```
davidchambers: Even if we decide to return the input in some cases, I don't think this should be enforced by the test suite. If, on the other hand, we decide it's important to return a new string map in all cases we could consider asserting that the output is *not* identical to the input.

index.js
```diff
@@ -3700,6 +3700,60 @@
...
+  //.
+  //. Equivalent to Haskell's `delete` function. Similar to Clojure's `dissoc`
+  //. functiom.
+  //.
+  //. ```javascript
+  //. > S.remove('c', {a: 1, b: 2, c: 3})
+  //. {a: 1, b: 2}
+  //.
+  //. > S.remove('c', {a: 1, b: 2})
+  //. {a: 1, b: 2}
```
davidchambers: Let's drop this example. The following example is sufficient to demonstrate that "removing" a non-existent key is a valid operation.

test/remove.js
samueller: If we're specifying that `remove` returns the original object, shouldn't we test for it? We could just not specify whether removing a non-existing key will return the original object. Or we could _always_ return a new string map.

index.js
samueller: True

index.js
davidchambers: > What about `if (strMap.hasOwnProperty(key))`?

This would get us a bit closer, but we must allow for the possibility of `'hasOwnProperty'` being one of the keys. We'd need `Object.prototype.hasOwnProperty.call(strMap, key)`.

I think we should match the semantics of `Z.concat`, though, which is interested in *enumerable* properties rather than *own* properties. For example:

```javascript
function Point() {}
Point.prototype.x = 0;
Point.prototype.y = 0;

Z.concat(new Point(), {});  // => {x: 0, y: 0}
```

samueller: Good catch on the `'hasOwnProperty'` as a key issue.

I understand `Z.concat` uses for..in to iterate through keys of enumerable and inherited enumerable properties. We could use`Object.prototype.propertyIsEnumerable.call(strMap, key)`, but that will only tell us if a property is enumerable. It will return `false` if a property is _inherited_ enumerable as in your `Point` example above. So, if we want to stick with returning the original string map when it doesn't contain the key to remove, we need to iterate through the keys before determining whether we can simply return the original string map:

```javascript
  function remove(key, strMap) {
    var found = false;
    var result = {};
    for (var k in strMap) {
      if (k === key) {
        found = true;
      } else {
        result[k] = strMap[k];
      }
    }
    if (found) {
      return result;
    } else {
      return strMap;
    }
  }
```
If optimizing on removing non-existing keys more often, we could search for the key first before copying.

davidchambers: > ```javascript
> if (found) {
>   return result;
> } else {
>   return strMap;
> }
> ```

If we're to do the work to generate `result`, why not return it in all cases? Sanctuary users, ideally, will not care about object identity, so it shouldn't be necessary to have `S.remove(key, strMap) === strMap` evaluate to `true` in certain cases. It may, though, be useful to guarantee that this expression *never* evaluates to `true` (the `delete`-based implementation above takes advantage of the fact that `Z.concat` always returns a new string map).

samueller: The idea was it was more efficient to return the original string map back if the key didn't exist. If we're going to build the `result` object anyway, the performance/efficiency benefit is diminished. I can imagine there are still benefits as the temporary `result` will get garbage collected. But I don't know how much that matters vs the benefit of guaranteeing `remove` never returns the original string map.

I can use your `Z.concat` implementation or I can use the above implementation without the code to return the original string map if the key wasn't found.

davidchambers: I appreciate @Avaq's countervailing input on performance matters, as I'm inclined to place more weight on consistency (always returning a new string map, in this case) than on performance. ;)

I'd like to know what others think. @Avaq? @gabejohnson? @safareli?

safareli: If we know that user will not mutate input of the function, than it is fine to return same value back if key is not there. But problem is that you don't know it and i'm not sure if you can assume it, some people might use sanctuary just for type checking not for controlling effects.

but if input `isFrozen` then we know that it will not be mutated and it might be correct to return same value.

gabejohnson: I don't know that we necessarily _should_ match the semantics of `S.concat` as it works on objects in general. I think we may inadvertantly cause confusion by not explicitly defining `StrMap`. In my opinion inherited properties should not be copied for a `StrMap`.

gabejohnson: I don't know that we necessarily _should_ match the semantics of `S.concat` as it works on objects in general. I think we may inadvertantly cause confusion by not explicitly defining `StrMap`. In my opinion inherited properties should not be copied for a `StrMap`.

Avaq: @gabejohnson You might want to join the discussion here: https://github.com/sanctuary-js/sanctuary-def/issues/164

On this specific case, it looks like the whole `result` object must be computed either way, and the choice is between returning a clone or the original. My preference would be returning a clone. I think the argument of memory usage does not outweigh the argument of consistency. In fact, if this function is used in a pure pipeline, the original object will be garbage collected anyway. If this function is used in a way where references are kept to *both* the input and output, I think users would be safer with the output being a clone. I'm in favor of `return result;`.

davidchambers: Thanks for raising your concern, @gabejohnson. Let's discuss the matter in sanctuary-js/sanctuary-type-classes#65.

> I think we should match the semantics of `Z.concat`, though, which is interested in *enumerable* properties rather than *own* properties.

We may to decide to change the behaviour of `Z.concat`, but I stand by the first clause of the sentence above. The only way to be sure that we're matching the semantics of `Z.concat` is to implement `S.insert` and `S.remove` in terms of `Z.concat`. Let's use the implementation I proposed at the beginning of this thread. If the behaviour of `Z.concat` changes in the future the behaviour of `S.insert` and `S.remove` will adjust accordingly.

If you're concerned, @samueller, that a future release of sanctuary-type-classes could introduce a subtle regression, you're welcome to include tests for the handling of inherited properties. If and when we were to upgrade to a version of sanctuary-type-classes with different semantics we would then update the assertions. These tests would serve to remind us to mention breaking changes to `S.insert` and `S.remove` in future release notes rather than to bind us to the current semantics.

ucodescott: So, which implementation?
```javascript
var result = Z.concat(strMap, {});
delete result[key];
return result;
```
or
```javascript
var result = {};
for (var k in strMap) {
  if (k !== key) result[k] = strMap[k];
}
return result;
```
First one is automatically updated when/if `Z.concat` is updated. Second one is more efficient assuming `delete` is more expensive than the `n` key comparisons.

samueller: So, which implementation?
```javascript
var result = Z.concat(strMap, {});
delete result[key];
return result;
```
or
```javascript
var result = {};
for (var k in strMap) {
  if (k !== key) result[k] = strMap[k];
}
return result;
```
First one is automatically updated when/if `Z.concat` is updated. Second one is more efficient assuming `delete` is more expensive than the `n` key comparisons.

Tests for inherited properties make sense.

davidchambers: Let's use this implementation:

```javascript
var result = Z.concat(strMap, {});
delete result[key];
return result;
```

> [The implementation above] is automatically updated when/if `Z.concat` is updated.

I consider this a requirement. I'm keen to avoid inconsistencies such as the one documented in ramda/ramda#1941.

kedashoe: We also have `S.keys`. Whichever way we go with properties to include, I think we should have (pseudo code, I think I get to the idea):

```js
function insert_test(key, val, strMap) {
  new Set(S.keys(strMap).concat([key])) === new Set(S.keys(S.insert(key, val, strMap)))
}
```
?

davidchambers: Great point, Kevin! It would be good to test this property. Here are a few more:

  - `S.insert (key) (val) (S.insert (key) (val) (map)) = S.insert (key) (val) (map)`
  - `S.remove (key) (S.remove (key) (map)) = S.remove (key) (map)`
  - `S.sort (S.keys (map)) = S.sort (S.keys (S.map (S.I) (map)))`

samueller: I like the tests and have added them, except the last assertion you mentioned @davidchambers. I assume that test was meant for the `keys` or the `map` function? I can add that to this pull request if you want.

davidchambers: > I assume that test was meant for the `keys` or the `map` function?

Yes, you're right. It would assert that `S.map` does not "promote" any properties which `S.keys` does not see. Let's leave that for a separate pull request.

I look forward to seeing your changes. Feel free to push the changes you have made so far even if there are more to come. You can tidy the commit history before we merge the pull request. :)

samueller:
Don't know if there was a notification, but I pushed changes yesterday

davidchambers:
> Don't know if there was a notification, but I pushed changes yesterday

There are notifications for regular pushes but not for `--force` pushes.

I'll have a look at the changes now. :)

index.js
```diff
@@ -3700,6 +3700,52 @@
...
+    result[key] = val;
+    return result;
+  }
+  S.insert =
+  def('insert', {}, [$.String, a, $.StrMap(a), $.StrMap(a)], insert);
+
+  //# remove :: String -> StrMap a -> StrMap a
+  //.
+  //. Removes a given key and its value from a shallow clone of a given
+  //. string map. The original string map is returned if the key doesn't exist.
```
davidchambers: The second sentence is no longer accurate.

samueller: You are correct, I apologize for the oversight! Pushing an update now.

index.js
```diff
@@ -3700,6 +3700,52 @@
   S.pairs =
   def('pairs', {}, [$.StrMap(a), $.Array($.Pair($.String, a))], pairs);
 
+  //# insert :: String -> a -> StrMap a -> StrMap a
+  //.
+  //. Inserts a given key and value into a shallow clone of a given
+  //. string map. Overrides the value if a key already exists.
```
davidchambers: I think we should replace <del>a</del> with <ins>the</ins> in several places. In the second sentence it's important that we write *the* key rather than *a* (any) key.

test/insert.js
```diff
@@ -0,0 +1,57 @@
...
+      unique(S.sort(S.concat(S.keys(strMap), [key]))),
+      S.sort(S.keys(S.insert(key, val, strMap)))
+    );
+  }
+
+  function testDoubleInsert(key, val, strMap) {
+    eq(
+      S.insert(key, val, S.insert(key, val, strMap)),
+      S.insert(key, val, strMap)
+    );
```
davidchambers: This might make the double application even clearer:

```javascript
var f = S.insert(key, val);
eq(f(f(strMap)), f(strMap));
```

test/insert.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  function test(key, val, strMap, strMapResult) {
```
davidchambers: We should avoid shadowing the global `test` function provided by Mocha.

test/insert.js
```diff
@@ -0,0 +1,57 @@
...
+
+  function unique(sortedArray) {
+    var uniqueArray = [];
+    sortedArray.forEach(function(val, i) {
+      if (i === 0 || val !== sortedArray[i - 1]) {
+        uniqueArray.push(val);
+      }
+    });
+    return uniqueArray;
+  }
```
davidchambers: Here's a less imperative implementation:

```javascript
function unique(xs) {
  if (xs.length === 0) return [];

  var sorted = S.sort(xs);
  var x = sorted[0];
  return sorted.slice(1).reduce(function(acc, x) {
    return S.equals(x, acc.prev) ? acc : {xs: S.append(x, acc.xs), prev: x};
  }, {xs: [x], prev: x}).xs;
}
```

I think we should avoid requiring a sorted array as input, even if this means we end up sorting twice in some cases.

samueller: What about:
```javascript
function unique(xs) {
  return S.sort(xs).reduce(function(acc, x, i, xs) {
    return i > 0 && S.equals(x, xs[i-1]) ? acc : S.append(x, acc)
  }, []);
}
```
or
```javascript
function appendIfUnique(acc, x, i, xs) {
  return i > 0 && S.equals(x, xs[i-1])
    ? acc
    : S.append(x, acc);
}

function unique(xs) {
  return S.sort(xs).reduce(appendIfUnique, []);
}
```

test/insert.js
samueller: Yes. I wasn't aware of the global `test` function.

test/insert.js
samueller: I agree.

index.js
samueller: Makes sense.

test/insert.js
samueller: Would this be clearer?
```javascript
var insert = S.insert(key, val);
eq(insert(insert(strMap)), insert(strMap));
```
I can do the same for `remove`:
```javascript
var remove = S.remove(key);
eq(remove(remove(strMap)), remove(strMap));
```

test/insert.js
davidchambers: The use of indexes and the `i > 0` guard in particular are the things which bother me. If we're to deal with indexes I'd prefer to use a `for` loop:

```javascript
function unique(xs) {
  if (xs.length === 0) return [];

  var sorted = S.sort(xs);
  var prev = sorted[0];
  var result = [prev];
  for (var idx = 1; idx < sorted.length; idx += 1) {
    if (!S.equals(sorted[idx], prev)) result.push(prev = sorted[idx]);
  }
  return result;
}
```

test/insert.js
davidchambers: Works for me. :thumbsup:

The name doesn't much matter to me, so long as the pattern is clear. :)

test/insert.js
samueller: I can replace with your `reduce` implementation. Just for my own edification, curious why you don't like indexes and `i > 0`? Is it the potential for getting the wrong array element?

davidchambers: I'd prefer to handle the special case (`[]`) in one place rather than allow it to complicate the common code path. I don't mind using array indexes when implementing Sanctuary functions, but in contexts in which performance is less of a concern (such as a test suite) I'd prefer to avoid "cheating" (keep in mind that `S.map` and friends do not expose indexes). In this case we're using an index to access the previous element, but it seems more natural to me to keep track of this.

This is all subjective, and I don't mean to assert that my preferences are better than alternatives in any meaningful way.

samueller: Updated pull request.

test/remove.js
```diff
@@ -0,0 +1,53 @@
...
+  // test inherited properties
+  testRemove('a', inherit({a: 1}, {}), {});
+  testRemove('b', inherit({a: 1}, {}), {a: 1});
+  testRemove('a', inherit({a: 1}, {a: 2}), {});
+  testRemove('b', inherit({a: 1}, {a: 2}), {a: 2});
+  testRemove('a', inherit({b: 2}, {a: 1}), {b: 2});
+  testRemove('a', inherit({a: 1}, {b: 2}), {b: 2});
+  testRemove('c', inherit({a: 1}, {b: 2}), {a: 1, b: 2});
+  testRemove('a', inherit({b: 2, c: 3}, {a: 1}), {b: 2, c: 3});
+  testRemove('b', inherit({b: 2, c: 3}, {a: 1}), {a: 1, c: 3});
```
davidchambers: I don't see value in testing so many combinations. Would something like this suffice?

```javascript
function Point() {}
Point.prototype.x = 0;
Point.prototype.y = 0;

eq(S.remove('x', new Point()), {y: 0});
eq(S.remove('y', new Point()), {x: 0});
eq(S.remove('z', new Point()), {x: 0, y: 0});
```

test/remove.js
```diff
@@ -0,0 +1,53 @@
...
+  eq(S.remove.toString(), 'remove :: String -> StrMap a -> StrMap a');
+
+  function testBasic(key, strMap, strMapResult) {
+    eq(S.remove(key, strMap), strMapResult);
+  }
+
+  function testDoubleRemove(key, strMap) {
+    var remove = S.remove(key);
+    eq(remove(remove(strMap)), remove(strMap));
+  }
```
davidchambers: Let's replace this with a property-based test:

```javascript
jsc.assert(jsc.forall(jsc.string, jsc.dict(jsc.number), function(key, map) {
  var remove = S.remove(key);
  return S.equals(remove(remove(map)), remove(map));
}), {tests: 1000});
```

You should be able to do something similar for `testDoubleInsert`. :)

samueller: Was unaware of jsverify! Very nice. I've made the changes as well as the above `Point` inherited properties test. Note that it does take well over triple the time to run the tests now on my MacBook due to insert.js taking 4.5secs and remove.js taking 3.5secs with these jsverify tests.

test/insert.js
```diff
@@ -0,0 +1,58 @@
...
+    jsc.assert(jsc.forall(jsc.string, jsc.number, jsc.dict(jsc.number), function(key, val, map) {
+      var insert = S.insert(key, val);
+      return S.equals(insert(insert(map)), insert(map));
+    }), {tests: 1000});
+  }
+
+  function testInsert(key, val, strMap, strMapResult) {
+    testBasic(key, val, strMap, strMapResult);
+    testKeysOfInsert(key, val, strMap);
+    testDoubleInsert();
```
davidchambers: This means we test the property 1000 times each time we call `testInsert`, which seems excessive. ;)

Testing the property 1000 times rather than 8000 times may prevent the tests from timing out on CircleCI.

test/remove.js
```diff
@@ -0,0 +1,47 @@
...
+    testBasic(key, strMap, strMapResult);
+    testDoubleRemove(key, strMap);
+  }
+
+  testRemove('a', {}, {});
+  testRemove('a', {a: 1}, {});
+  testRemove('a', {b: 2}, {b: 2});
+  testRemove('b', {a: 1, b: 2}, {a: 1});
+  testRemove('a', {b: 2, c: 3}, {b: 2, c: 3});
+  testRemove('b', {a: 1, b: 2, c: 3}, {a: 1, c: 3});
```
davidchambers: As mentioned above, let's test the property just 1000 times. This will mean we can remove `testRemove` and make assertions here directly:

```javascript
eq(S.remove('a', {}), {});
...
eq(S.remove('b', {a: 1, b: 2, c: 3}), {a: 1, c: 3});
```

I think we could achieve 100% coverage with fewer test cases. These should suffice:

```javascript
eq(S.remove('x', {x: 1, y: 2}), {y: 2});
eq(S.remove('y', {x: 1, y: 2}), {x: 1});
eq(S.remove('z', {x: 1, y: 2}), {x: 1, y: 2});
```

test/insert.js
samueller: Silly mistake on my part to call `testDoubleInsert` on every call to `testInsert`

davidchambers: No worries. :)

test/remove.js
```diff
@@ -0,0 +1,47 @@
...
+
+  // test inherited properties
+  function Point() {}
+  Point.prototype.x = 0;
+  Point.prototype.y = 0;
+
+  eq(S.remove('x', new Point()), {y: 0});
+  eq(S.remove('y', new Point()), {x: 0});
+  eq(S.remove('z', new Point()), {x: 0, y: 0});
```
davidchambers: Shall we include similar test cases for `insert`?

samueller: Makes sense

test/remove.js
samueller: I've made those changes locally. Are the `insert` tests excessive as well?

davidchambers: I would say so. I like to be able to look at each test case and see what *it alone* tests. I see only two distinct scenarios: either the key is already present or it is not. We really only need two unit tests (in addition to the property tests), but I don't mind including a few more if it makes a nice `a`, `b`, `c` or `1`, `2`, `3` pattern.

samueller: Updated pull request with 3 `insert` tests, making somewhat of an `a`, `b` / `1`, `2`, `3` pattern.

test/insert.js
```diff
@@ -0,0 +1,62 @@
...
+
+  function testInsert(key, val, strMap, strMapResult) {
+    testBasicInsert(key, val, strMap, strMapResult);
+    testKeysOfInsert(key, val, strMap);
+  }
+
+  testInsert('a', 1, {}, {a: 1});
+  testInsert('b', 2, {a: 1}, {a: 1, b: 2});
+  testInsert('b', 3, {a: 1, b: 2}, {a: 1, b: 3});
+  testDoubleInsert();
```
davidchambers: Let's just include the body of `testDoubleInsert` here. The code is self-descriptive.

test/insert.js
```diff
@@ -0,0 +1,62 @@
...
+      return S.equals(x, acc.prev) ? acc : {xs: S.append(x, acc.xs), prev: x};
+    }, {xs: [x], prev: x}).xs;
+  }
+
+  function testKeysOfInsert(key, val, strMap) {
+    eq(
+      unique(S.sort(S.concat(S.keys(strMap), [key]))),
+      S.sort(S.keys(S.insert(key, val, strMap)))
+    );
+  }
```
davidchambers: Correct me if I'm wrong, but I don't think this tests anything not covered by other tests is this file. If we were to test `new Point()` the equivalence would not hold because `S.keys(new Point())` is `[]` rather than `['x', 'y']`. This suggests that `S.keys` cannot be implemented in terms of `Object.keys`. This changes the way I feel about sanctuary-js/sanctuary-type-classes#65. I'll comment on that issue shortly.

samueller: I added the `S.keys` test above because of the conversation between you and @kedashoe earlier. I didn't realize `S.keys(new Point())` would yield `[]`. I understand you're going to decide whether to make `S.keys` get keys of enumerable inherited properties or change `S.concat` to not get enumerable inherited properties. I can wait until you decide or remove `testKeysOfInsert` entirely.

davidchambers: We'll update the behaviour of `Z.concat` and `S.concat`. The behaviour of `S.keys` will remain as is.

samueller: So `Z.concat` and `S.concat` no longer include enumerable inherited properties? And I assume then this PR's `S.insert` and `S.remove` should also no longer include enumerable inherited properties. Then I'll no longer test for enumerable inherited properties or use `new Point()`. Or I could explicitly test that enumerable inherited properties are _not_ included in the results of `S.insert` and `S.remove`?

Do you want me to keep or delete the `testKeysOfInsert` code?

samueller: @davidchambers just following up on this.

davidchambers: Thanks for bringing this pull request to my attention again, Scott. The past few weeks have been busy for me.

I suggest we do the following:

1.  I'll merge the outstanding pull request for sanctuary-type-classes and release v6.1.0.
2.  I'll open a pull request for the change to `Z.concat`.
3.  I'll merge the pull request opened in the previous step.
4.  I'll release v7.0.0.
5.  We'll update to `sanctuary-type-classes@7.0.0`.
6.  You'll rebase this pull request and update the assertions accordingly.
7.  I'll merge this pull request.

We needn't do things in this order, of course, but it would be nice to avoid adding new functions to Sanctuary only to change their behaviour in the following release.

samueller: Makes sense. I'll be ready for step 6 :). Am I correct in understanding that `Z.concat` and `S.concat` no longer include enumerable inherited properties?

davidchambers: > Am I correct in understanding that `Z.concat` and `S.concat` no longer include enumerable inherited properties?

Just about. `Z.concat` and `S.concat` *will* no longer include enumerable inherited properties. ;)

davidchambers: sanctuary-js/sanctuary-type-classes#70

davidchambers: @samueller, I've completed steps 1–5 in the list above so you should no longer be blocked. :)

davidchambers: #439 introduced a StrMap section, so when you rebase please also update the commit message prefix from `object:` to `strmap:`.

davidchambers:
Would you mind having another look at this pull request, @gabejohnson? :eyes:

index.js
```diff
@@ -3757,6 +3757,51 @@
...
+  //.
+  //. ```javascript
+  //. > S.insert('c', 3, {a: 1, b: 2})
+  //. {a: 1, b: 2, c: 3}
+  //.
+  //. > S.insert('a', 4, {a: 1, b: 2})
+  //. {a: 4, b: 2}
+  //. ```
+  function insert(key, val, strMap) {
+    return Z.concat(strMap, singleton(key, val));
```
gabejohnson: I like this implementation in particular 😄

masaeedu: Is there a policy of not "cannibalizing" Sanctuary functions to implement other things? You could do something like `const insert = compose2(concat, singleton)`. I guess without a type system to guide you getting too abstract gets confusing.

gabejohnson: I think `insert`\`assoc` is used frequently enough to rate it's own function. Unless I'm misunderstanding you.

masaeedu: @gabejohnson I agree with having `insert` as a separate function, I didn't mean to say this shouldn't be part of the library. I'm just saying that when we implement these functions, it seems like we prefer to use `Z` in the implementation even when it makes the implementation a little more awkward than if we used the equivalent `S` function.

gabejohnson: This is because of the overhead introduced by `def` (currying, redundant type checking machinery). _Some_ care is taken to write performant implementations.

davidchambers: Haha! Your use of italics just made me chuckle, Gabe. :)

We avoid using the curried functions, Asad, for the reason Gabe gave, but the *implementation* functions can freely be used in other implementations. The `head` implementation, for example, is defined in terms of the `at` implementation:

> ```javascript
> function head(xs) {
>   return at(0, xs);
> }
> ```

davidchambers:
Thanks for your work on this pull request, @samueller. :)

PR #439, opened 2017-09-08
strmap: add StrMap heading and description

davidchambers:
This is the result of discussion in sanctuary-js/sanctuary-type-classes#65.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3663,6 +3663,14 @@
   S.gets =
   def('gets', {}, [$.Predicate($.Any), $.Array($.String), a, $Maybe(b)], gets);
 
+  //. ### StrMap
+  //.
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -3663,6 +3663,14 @@
   S.gets =
   def('gets', {}, [$.Predicate($.Any), $.Array($.String), a, $Maybe(b)], gets);
 
+  //. ### StrMap
+  //.
+  //. StrMap is an abbreviation of _string map_. A string map is an object,
+  //. such as `{foo: 1, bar: 2, baz: 3}`, whose values are all members of
+  //. some type. Formally, a value is a member of type `StrMap a` if its
```
kedashoe: some type -> the same type?

davidchambers: I like it! I'll make this change.

PR #445, opened 2017-09-29
sanctuary-type-classes@6.1.0

davidchambers:
<https://github.com/sanctuary-js/sanctuary-type-classes/releases/tag/v6.1.0>

### DIFF
--- a/index.js
+++ b/index.js
@@ -480,9 +480,9 @@
 
   //# is :: TypeRep a -> Any -> Boolean
   //.
-  //. Takes a [type representative](#type-representatives) and a value of
-  //. any type and returns `true` if the given value is of the specified
-  //. type; `false` otherwise. Subtyping is not respected.
+  //. Takes a [type representative](#type-representatives) and a value of any
+  //. type and returns `true` [iff][] the given value is of the specified type.
+  //. Subtyping is not respected.
   //.
   //. ```javascript
   //. > S.is(Number, 42)
@@ -556,7 +556,8 @@
 
   //# lt :: Ord a => a -> (a -> Boolean)
   //.
-  //. Flipped version of [`Z.lt`][] intended for partial application.
+  //. Returns `true` [iff][] the *second* argument is less than the first
+  //. according to [`Z.lt`][]. The arguments must be provided one at a time.
   //.
   //. See also [`lt_`](#lt_).
   //.
@@ -568,7 +569,8 @@
 
   //# lt_ :: Ord a => a -> a -> Boolean
   //.
-  //. Curried version of [`Z.lt`][].
+  //. Returns `true` [iff][] the first argument is less than the second
+  //. according to [`Z.lt`][].
   //.
   //. See also [`lt`](#lt).
   //.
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -690,10 +690,7 @@
   //. > S.min('10', '2')
   //. '10'
   //. ```
-  function min(x, y) {
-    return Z.lte(x, y) ? x : y;
-  }
-  S.min = def('min', {a: [Z.Ord]}, [a, a, a], min);
+  S.min = def('min', {a: [Z.Ord]}, [a, a, a], Z.min);
```
Avaq: Most Z-derived functions have a doc like:

> Curried version of `Z.fn`

Should we update all these docs?

davidchambers: Which approach do you prefer, Aldwin? Linking to the `Z` functions or duplicating their descriptions? I see value in both, but I agree that we should be consistent.

gabejohnson: Duplication runs the risk of stale documentation in this project. That being said, I would prefer duplication. Every level of indirection is another annoyance.

Avaq: Can we build the documentation from Z's source?

davidchambers: > Can we build the documentation from Z's source?

Potentially, but there's not a one-to-one correspondence between `Z` functions and their corresponding `S` functions. Consider [`Z.chainRec`][1] and [`S.chainRec`][2], for example.


[1]: https://github.com/sanctuary-js/sanctuary-type-classes#chainRec
[2]: https://sanctuary.js.org/#chainRec

davidchambers: I've added a commit which adds descriptions for some of the Fantasy Land functions (taken from sanctuary-type-classes). Please have a look. :eyes:

Some functions still lack “proper” descriptions, but for a function such as [`S.sequence`][1] which is difficult to explain in words (without restating the type signature), I'm happy to link to [`Z.sequence`][2] which in turn links to [`fantasy-land/traverse`][3] (via [`Z.traverse`][4]).


[1]: https://sanctuary.js.org/#sequence
[2]: https://github.com/sanctuary-js/sanctuary-type-classes#sequence
[3]: https://github.com/fantasyland/fantasy-land#traverse-method
[4]: https://github.com/sanctuary-js/sanctuary-type-classes#traverse

davidchambers:
> I think every description adds value, but also maintenance cost.

I'd like to provide a “proper” description for every function, but some functions are not easy to describe in words. I welcome pull requests. :)

index.js
```diff
@@ -3023,10 +3001,9 @@
 
   //# anyPass :: Array (a -> Boolean) -> a -> Boolean
   //.
-  //. Takes an array of unary predicates and a value of any type
-  //. and returns `true` if any of the predicates pass; `false` otherwise.
-  //. None of the subsequent predicates will be evaluated after the
-  //. first passed predicate.
+  //. Takes an array of predicates and a value of any type and returns `true`
+  //. [iff][] the value satisfies any of the predicates. None of the subsequent
+  //. predicates will be applied after the first predicate satisfied.
```
davidchambers: I made use of [iff][1] here then decided to make some other changes to the `S.allPass` and `S.anyPass` descriptions while I was at it.


[1]: https://en.wikipedia.org/wiki/If_and_only_if

PR #454, opened 2017-10-31
Add `S.fromPairs`

futpib:


davidchambers:
Looks very good! I'll leave a few comments, but there's not much to fault here. :)

index.js
```diff
@@ -3778,6 +3778,24 @@
...
   def('pairs', {}, [$.StrMap(a), $.Array($.Pair($.String, a))], pairs);
 
+  //# fromPairs :: Array (Pair String a) -> StrMap a
+  //.
+  //. Returns a new string map formed by a list of key-value pairs.
+  //. If a key appears in multiple pairs, the rightmost pair takes precedence.
+  //.
+  //. ```javascript
+  //. > S.fromPairs([['a', 1], ['b', 2], ['c', 3]])
+  //. {b: 2, a: 1, c: 3}
```
davidchambers: Is there a reason to write `{b: 2, a: 1, c: 3}` rather than `{a: 1, b: 2, c: 3}`?

index.js
```diff
@@ -3778,6 +3778,24 @@
...
+  //. ```
+  function fromPairs(pairs) {
+    return Z.reduce(function(m, pair) {
```
davidchambers: I'd prefer to name the first parameter `strMap`.

index.js
```diff
@@ -3778,6 +3778,24 @@
   S.pairs =
   def('pairs', {}, [$.StrMap(a), $.Array($.Pair($.String, a))], pairs);
 
+  //# fromPairs :: Array (Pair String a) -> StrMap a
+  //.
+  //. Returns a new string map formed by a list of key-value pairs.
```
davidchambers: I think this is a slight improvement:

_Returns a string map containing the key–value pairs specified by the given array._

Note: Please use an [en dash][1] in _key–value pairs_. :)


[1]: http://www.thepunctuationguide.com/en-dash.html

test/fromPairs.js
```diff
@@ -0,0 +1,18 @@
...
+
+  eq(typeof S.fromPairs, 'function');
+  eq(S.fromPairs.length, 1);
+  eq(S.fromPairs.toString(), 'fromPairs :: Array (Pair String a) -> StrMap a');
+
+  eq(S.fromPairs([]), {});
+  eq(S.fromPairs([['a', 1], ['b', 2], ['c', 3]]), {a: 1, b: 2, c: 3});
+
+  eq(S.fromPairs([['x', 1], ['x', 2]]), {x: 2});
+});
```
davidchambers: Nit: Please delete line 16 and add an empty line between lines 17 and 18.

index.js
futpib: Copy-pasted from `S.pairs`. I guess it's a hint on the arbitrary order of things? Should we: leave it as is, sort only `fromPairs`, or sort both `pairs` and `fromPairs` examples?

futpib: Oh, wait, I see that `S.pairs` example is sorted explicitly with `.sort()`. I'll just sort the keys in `fromPairs` example then.

davidchambers: :thumbsup:

davidchambers:
This looks great to me! Please squash the commits. I'll leave this pull request open for a couple of days to give other Sanctuary contributors an opportunity to comment.

test/fromPairs.js
```diff
@@ -0,0 +1,18 @@
...
+
+test('fromPairs', function() {
+
+  eq(typeof S.fromPairs, 'function');
+  eq(S.fromPairs.length, 1);
+  eq(S.fromPairs.toString(), 'fromPairs :: Array (Pair String a) -> StrMap a');
+
+  eq(S.fromPairs([]), {});
+  eq(S.fromPairs([['a', 1], ['b', 2], ['c', 3]]), {a: 1, b: 2, c: 3});
+  eq(S.fromPairs([['x', 1], ['x', 2]]), {x: 2});
```
davidchambers: I just realized that this would also make an excellent doctest. It neatly demonstrates the second sentence of the description. Would you mind adding this to the `` ```javascript `` block as well?

index.js
```diff
@@ -3778,6 +3778,28 @@
...
+  //. {x: 2}
+  //. ```
+  function fromPairs(pairs) {
+    return Z.reduce(function(strMap, pair) {
+      strMap[pair[0]] = pair[1];
+      return strMap;
+    }, {}, pairs);
+  }
+  S.fromPairs =
+  def('fromPairs', {}, [$.Array($.Pair($.String, a)), $.StrMap(a)], fromPairs);
```
Avaq: Since the function uses `Z.reduce`, maybe this signature could be expanded to:

```hs
Foldable f => f (Pair String a) -> StrMap a
```

davidchambers: Good idea! Let's replace `array` with `[Foldable][]` in the description while we're at it. :)

index.js
```diff
@@ -3778,6 +3778,31 @@
...
+    return Z.reduce(function(strMap, pair) {
+      strMap[pair[0]] = pair[1];
+      return strMap;
+    }, {}, pairs);
+  }
+  S.fromPairs =
+  def('fromPairs',
+    {f: [Z.Foldable]},
+    [f($.Pair($.String, a)), $.StrMap(a)],
+    fromPairs);
```
davidchambers: Nit: Please indent the three lines above so that all four arguments to `def` are aligned. :)

index.js
```diff
@@ -3778,6 +3778,31 @@
   S.pairs =
   def('pairs', {}, [$.StrMap(a), $.Array($.Pair($.String, a))], pairs);
 
+  //# fromPairs :: Foldable f => f (Pair String a) -> StrMap a
+  //.
+  //. Returns a string map containing the key–value pairs specified by
+  //. the given [Foldable][]. If a key appears in multiple pairs, the rightmost
```
davidchambers: Nit: Let's make the lengths of the two lines above more even by moving the first word of the second line to the end of the first line. :)

davidchambers:
The changes look good, @futpib! I'll merge this pull request as soon as you've made the two minor formatting changes.

futpib:
I really suspect you are kidding me at this point.

davidchambers:
Thank you for your patience, @futpib. :smile:

index.js
futpib: Consider this #457

PR #456, opened 2017-10-31
test: avoid implementation-dependent parsing of date string

### LINKED ISSUE #455: Tests are flaky with respect to timezone
```
TZ=UTC make test
// exit code 0
```

```
TZ=Europe/Moscow make test
// exit code 2
```

```
  256 passing (3s)
  1 failing

  1)  toString:

      AssertionError [ERR_ASSERTION]: '"new Date(\\"2001-02-03T01:05:06.000Z\\")"' === '"new Date(\\"2001-02-03T04:05:06.000Z\\")"'
      + expected - actual

      -"new Date(\"2001-02-03T01:05:06.000Z\")"
      +"new Date(\"2001-02-03T04:05:06.000Z\")"
      
      at eq (test/internal/eq.js:11:10)
      at Context.<anonymous> (test/toString.js:43:3)

```

davidchambers:
Fixes #455

[MDN][1]:

> __Note:__ parsing of date strings with the `Date` constructor (and `Date.parse`, they are equivalent) is strongly discouraged due to browser differences and inconsistencies. Support for RFC 2822 format strings is by convention only. Support for ISO 8601 formats differs in that date-only strings (e.g. "1970-01-01") are treated as UTC, not local.

@futpib, please confirm that this addresses the issue for you.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date

### DIFF
--- a/test/toString.js
+++ b/test/toString.js
@@ -40,7 +40,7 @@ test('toString', function() {
   eq(S.toString(new Date(0)), 'new Date("1970-01-01T00:00:00.000Z")');
   eq(S.toString(new Date(42)), 'new Date("1970-01-01T00:00:00.042Z")');
   eq(S.toString(new Date(NaN)), 'new Date(NaN)');
-  eq(S.toString(new Date('2001-02-03T04:05:06')), 'new Date("2001-02-03T04:05:06.000Z")');
+  eq(S.toString(new Date('2001-02-03')), 'new Date("2001-02-03T00:00:00.000Z")');
   eq(S.toString([]), '[]');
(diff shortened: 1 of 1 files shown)


futpib:
This does fix the issue for me.

davidchambers:
> You could probably also set up your CI to run tests with different `TZ` environment variable values, although I don't know if that would be worth the hassle.

I don't consider this necessary. The test failure was the result of a poorly written test rather than a faulty implementation. :)

For the record I was unable to reproduce the issue in Node, but I was able to reproduce it in Firefox.

PR #457, opened 2017-11-01
A more clever way to ignore only `createSanctuary` body indentation

futpib:
Instead of disabling `indent` rule for a range of lines, disable it
for direct children of `createSanctuary` function, thus enabling
`indent` rule checks for it's deeper children.

For example, the following code if put inside `createSanctuary` would
have been considered correct by eslint before this commit:

```js
  def('fromPairs',
    {f: [Z.Foldable]},
    [f($.Pair($.String, a)), $.StrMap(a)],
    fromPairs);
```

After this commit, code like this whould have been reported as
`indent` rule violation.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,4 +1,7 @@
 {
   "root": true,
-  "extends": ["./node_modules/sanctuary-style/eslint-es3.json"]
+  "extends": ["./node_modules/sanctuary-style/eslint-es3.json"],
+  "rules": {
+    "indent": ["error", 2, {"SwitchCase": 1, "FunctionDeclaration": {"parameters": "first"}, "FunctionExpression": {"parameters": "first"}, "CallExpression": {"arguments": "first"}, "ArrayExpression": "first", "ObjectExpression": "first", "ignoredNodes": ["ConditionalExpression", "MemberExpression", "FunctionDeclaration[id.name=createSanctuary] > BlockStatement.body"]}]
+  }
 }
--- a/index.js
+++ b/index.js
@@ -336,8 +336,6 @@
   //  createSanctuary :: Options -> Module
   function createSanctuary(opts) {
 
-  /* eslint-disable indent */
-
   //  checkTypes :: Boolean
   var checkTypes = opts.checkTypes;
 
@@ -4557,8 +4555,6 @@
       splitOnRegex);
 
   return S;
-
-  /* eslint-enable indent */
 
   }
 


davidchambers:
Very cool!

My one concern is that we're duplicating all the existing indentation settings in order to add a special case for `createSanctuary`. Is it possible to specify `"ignoredNodes"` but inherit the other settings?

futpib:
I had this concern too.

Here is a rather cumbersome implementation https://github.com/futpib/sanctuary/pull/1/files . It's either that I can't find a proper way to inherit eslint config the way we would like, or that eslint indeed does not provide an easy way to do that.

An easier solution would be to directly require `eslint-common.json`, but that would bring in the assumption that `indent` rule is defined exactly there. I could require `eslint-es3.json` and join in whatever it `extends`, but that's like reimplementing eslint internal config resolution.

All these are clumsy (but the first solution at least seems correct), that's why I just duplicated the `indent` rule options.

davidchambers:
futpib/sanctuary#1 would introduce more of a maintenance burden than the small amount of duplication introduced in this pull request.

> An easier solution would be to directly require `eslint-common.json`, but that would bring in the assumption that `indent` rule is defined exactly there.

I'm willing to make that assumption. The result is quite reasonable:

```patch
diff --git a/.eslintrc.js b/.eslintrc.js
new file mode 100644
index 0000000..3db1526
--- /dev/null
+++ b/.eslintrc.js
@@ -0,0 +1,22 @@
+/* eslint-env node */
+/* eslint-disable quote-props */
+
+'use strict';
+
+var common = require('sanctuary-style/eslint-common.json');
+
+
+var indent = common.rules.indent.slice();
+indent[2] = Object.assign({}, indent[2], {
+  ignoredNodes: indent[2].ignoredNodes.concat([
+    'FunctionDeclaration[id.name=createSanctuary] > BlockStatement.body'
+  ])
+});
+
+module.exports = {
+  root: true,
+  extends: ['./node_modules/sanctuary-style/eslint-es3.json'],
+  rules: {
+    indent: indent
+  }
+};
diff --git a/.eslintrc.json b/.eslintrc.json
deleted file mode 100644
index 741f63f..0000000
--- a/.eslintrc.json
+++ /dev/null
@@ -1,7 +0,0 @@
-{
-  "root": true,
-  "extends": ["./node_modules/sanctuary-style/eslint-es3.json"],
-  "rules": {
-    "indent": ["error", 2, {"SwitchCase": 1, "FunctionDeclaration": {"parameters": "first"}, "FunctionExpression": {"parameters": "first"}, "CallExpression": {"arguments": "first"}, "ArrayExpression": "first", "ObjectExpression": "first", "ignoredNodes": ["ConditionalExpression", "MemberExpression", "FunctionDeclaration[id.name=createSanctuary] > BlockStatement.body"]}]
-  }
-}
diff --git a/Makefile b/Makefile
index 62c6e8f..fa4304c 100644
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ README.md.tmp: index.js
 
 .PHONY: lint
 lint:
-	$(ESLINT) -- index.js karma.conf.js scripts/version-urls test
+	$(ESLINT) --ignore-pattern !.eslintrc.js -- .eslintrc.js index.js karma.conf.js scripts/version-urls test
 	$(REMEMBER_BOWER) $(shell pwd)
 	rm -f README.md
 	VERSION=0.0.0 make README.md
```

I'm not sure whether the patch above is an improvement on this pull request. What do you think?

Perhaps we should stick with the current approach (which is by far the simplest) and have `make lint` assert that these commands produce the same output:

```console
$ node --print 'var indent = require("./.eslintrc.json").rules.indent; indent[2].ignoredNodes.pop(); JSON.stringify(indent)'
["error",2,{"SwitchCase":1,"FunctionDeclaration":{"parameters":"first"},"FunctionExpression":{"parameters":"first"},"CallExpression":{"arguments":"first"},"ArrayExpression":"first","ObjectExpression":"first","ignoredNodes":["ConditionalExpression","MemberExpression"]}]

$ node --print 'JSON.stringify(require("sanctuary-style/eslint-common.json").rules.indent)'
["error",2,{"SwitchCase":1,"FunctionDeclaration":{"parameters":"first"},"FunctionExpression":{"parameters":"first"},"CallExpression":{"arguments":"first"},"ArrayExpression":"first","ObjectExpression":"first","ignoredNodes":["ConditionalExpression","MemberExpression"]}]
```

futpib:
Why do you have this exceptional indentation in the first place? Why not indent the `createSanctuary` function the way `indent` expects it to be indented?

davidchambers:
> Why do you have this exceptional indentation in the first place?

I believe that when I added `createSanctuary` I decided not to indent its body for three reasons:

  - indenting would have necessitated rewrapping descriptions (to adhere to the 79-character limit);
  - indenting would have made `git blame` less convenient; and
  - I have an aesthetic preference for flat code.

davidchambers:
I'll close this pull request for the reason given in <https://github.com/sanctuary-js/sanctuary/pull/467#issuecomment-352295669>. I appreciate the suggestion, @futpib.

PR #464, opened 2017-12-07
env: remove refinement types from default environment

davidchambers:
Commit message:

> This dramatically lowers the performance cost of type checking collections in some cases.

The difference is *staggering* in some cases.

```console
~/github.com/sanctuary-js/sanctuary (master)
$ time node -e 'const S = require("./"), xs = []; for (let id = 1; id <= 8; id += 1) xs.push({id}); S.I(xs)'

real	1m9.221s
user	1m8.379s
sys	0m0.790s
```

```console
~/github.com/sanctuary-js/sanctuary (davidchambers/env)
$ time node -e 'const S = require("./"), xs = []; for (let id = 1; id <= 1000; id += 1) xs.push({id}); S.I(xs)'

real	0m0.450s
user	0m0.456s
sys	0m0.021s
```

Increasing the length of `xs` from 8 to 9 on `master` results in the heap consuming all available memory on my computer. On `davidchambers/env` increasing the length of `xs` from 8 to 1000 only adds about 200 milliseconds to the run time.

Removing refinement types from the default environment will result in less informative error messages in some cases (but much clearer error messages in certain other cases). @Avaq has proposed that sanctuary-def should *explicitly* support refinement types (sanctuary-js/sanctuary-def#162). Were it to do so, refinement types could be ignored when resolving type variables, but could be considered when enumerating a value's types in an error message. The best of both worlds!

Many thanks to @mfidemraizer for providing a minimal test case which made it easy to diagnose this long-standing issue. :clap:

### DIFF
--- a/index.js
+++ b/index.js
@@ -315,19 +315,9 @@
 
   //  defaultEnv :: Array Type
   var defaultEnv = Z.concat($.env, [
-    $.FiniteNumber,
-    $.NonZeroFiniteNumber,
     $Either($.Unknown, $.Unknown),
-    Fn($.Unknown, $.Unknown),
-    $.GlobalRegExp,
-    $.NonGlobalRegExp,
-    $.Integer,
-    $.NonNegativeInteger,
     $Maybe($.Unknown),
-    $.Pair($.Unknown, $.Unknown),
-    $.RegexFlags,
-    $.ValidDate,
-    $.ValidNumber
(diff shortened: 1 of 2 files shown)


mfidemraizer:
@davidchambers No problem :D

davidchambers:
Closing in favour of sanctuary-js/sanctuary-def#177

PR #468, opened 2017-12-18
documentation: use placeholder for version-specific type URL

davidchambers:
Remembering to update such things manually is clearly not a good solution.

### DIFF
--- a/scripts/version-urls
+++ b/scripts/version-urls
@@ -12,7 +12,7 @@ var pkg = require('../package.json');
 var deps = Object.assign({}, pkg.dependencies, pkg.devDependencies);
 
 process.stdout.write(fs.readFileSync(process.argv[2], 'utf8').replace(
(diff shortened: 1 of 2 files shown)


scripts/version-urls
```diff
@@ -12,7 +12,7 @@ var pkg = require('../package.json');
 var deps = Object.assign({}, pkg.dependencies, pkg.devDependencies);
 
 process.stdout.write(fs.readFileSync(process.argv[2], 'utf8').replace(
-  new RegExp(' v:(.+)/(.+)', 'g'),
+  / v:(\S+)[/](\S+)/g,
```
davidchambers: I had been using a string literal to avoid escaping the `/`, but `\\S` would be even uglier. The only significant change here is replacing `.+` with `\S+` in the second capturing group to remove the assumption that the placeholder appears at the end of a line. I updated the first capturing group to match, though this wasn't necessary.

PR #470, opened 2017-12-19
hide doctest-specific imports from build tools

### LINKED ISSUE #469: Dynamic require causes build problems
Is it possible to move the _test_ `require()` calls outside of the core index.js?

Specifically, [this block](https://github.com/sanctuary-js/sanctuary/blob/master/index.js#L412) is causing me some build errors since the `test` directory does not exist:

```
 if (typeof __doctest !== 'undefined') {
    /* eslint-disable no-unused-vars */
    var _List = require('./test/internal/List');
    var Cons = _List.Cons;
    var Nil = _List.Nil;
    var Sum = require('./test/internal/Sum');
    /* eslint-enable no-unused-vars */
    env = Z.concat(env, [_List.Type($.Unknown), Sum.Type]);
 }
```

It may be a problem with Webpack+Rollup+Typescript, or my config, but I think my project setup is probably normal enough that this bug may be encountered by others. Additionally, it uses other libraries and only Sanctuary is causing this error at the moment (commenting out that block fixes it)

davidchambers:
Fixes #469

@dakom, does updating your `sanctuary` dependency to `sanctuary-js/sanctuary#4f6826d` fix your issue?

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   },
   "devDependencies": {
     "browserify": "13.1.x",
-    "doctest": "0.12.x",
+    "doctest": "0.13.x",
     "eslint": "4.9.x",
     "fantasy-land": "3.4.0",
     "istanbul": "0.4.x",
--- a/index.js
+++ b/index.js
@@ -410,11 +410,12 @@
 
   /* istanbul ignore if */
   if (typeof __doctest !== 'undefined') {
+    /* global __doctest:false */
     /* eslint-disable no-unused-vars */
-    var _List = require('./test/internal/List');
+    var _List = __doctest.require('./test/internal/List');
     var Cons = _List.Cons;
     var Nil = _List.Nil;
-    var Sum = require('./test/internal/Sum');
+    var Sum = __doctest.require('./test/internal/Sum');
     /* eslint-enable no-unused-vars */
     env = Z.concat(env, [_List.Type($.Unknown), Sum.Type]);
   }


dakom:
Yes it does :)

davidchambers:
Is there a reason not to use `eval` in this way? I'm concerned that any use of `eval` could violate security policies and prevent Sanctuary from being used in certain contexts, but this is speculation.

dakom:
I don't know of any policies that would prevent it - typescript+rollup+webpack inherently allows it... I'll share this link in a couple places to try and get more eyes on it

davidchambers:
I realized that we can use `arguments[1]` rather than `eval('require')`. This fools webpack. The only problem is that doctest wraps the source code in an additional IIFE, preventing `arguments` from referring to the arguments to the CommonJS IIFE.

Since our doctests are the root cause of these problems, I think the right fix is for doctest to expose `require` in some way that webpack does not detect.

gabejohnson:
There are no security concerns with this use of `eval` as the string is statically known. However, there are performance concerns. I think functions containing `eval` will always be compiled using the generic compiler vs. the optimizing, at least in V8.

If we _must_ use `eval`, it would be better to isolate it in separate function.

davidchambers:
@dakom, would you mind testing `sanctuary-js/sanctuary#fdf4386`? It's `eval`-free. :)

dakom:
@davidchambers - works perfectly :)

package.json
```diff
@@ -17,7 +17,7 @@
   },
   "devDependencies": {
     "browserify": "13.1.x",
-    "doctest": "0.12.x",
+    "doctest": "0.13.x",
```
davidchambers: <https://github.com/davidchambers/doctest/releases/tag/0.13.0>

PR #472, opened 2018-01-01
composition: add pipeK

### LINKED ISSUE #380: Add pipeK for monads
Ramda has Kleisli composition via [composeK](http://ramdajs.com/docs/#composeK) and [pipeK](http://ramdajs.com/docs/#pipeK). We probably only need one of those.

It is worth considering instead a specification for Categories as suggested in https://github.com/fantasyland/fantasy-land/issues/208 and implementing a Kleisli type that exposes its own `compose`.

This would mean `pipe` would be sufficient if it were implemented in terms of `compose`.

gabejohnson:
Closes #380

### DIFF
--- a/index.js
+++ b/index.js
@@ -1489,9 +1489,9 @@
 
   //# pipe :: Foldable f => f (Any -> Any) -> a -> b
   //.
-  //. Takes an array of functions assumed to be unary and a value of any type,
-  //. and returns the result of applying the sequence of transformations to
-  //. the initial value.
+  //. Takes a sequence of functions assumed to be unary and a value of any
+  //. type, and returns the result of applying the sequence of transformations
+  //. to the initial value.
   //.
   //. In general terms, `pipe` performs left-to-right composition of a sequence
   //. of functions. `pipe([f, g, h], x)` is equivalent to `h(g(f(x)))`.
@@ -1504,6 +1504,29 @@
     return Z.reduce(function(x, f) { return f(x); }, x, fs);
   }
   S.pipe = def('pipe', {f: [Z.Foldable]}, [f(Fn($.Any, $.Any)), a, b], pipe);
+
+  //# pipeK :: (Foldable f, Chain m) => f (Any -> m Any) -> m a -> m b
+  //.
+  //. Takes a sequence of functions assumed to be unary which return values
+  //. with a [Chain][], and a value of that Chain, and returns the result
+  //. of applying the sequence of transformations to the initial value.
+  //.
+  //. In general terms, `pipeK` performs left-to-right [Kleisli][] composition
+  //. of an sequence of functions. `pipeK([f, g, h], x)` is equivalent to
+  //. `chain(h, chain(g, chain(f, x)))`.
+  //.
+  //. ```javascript
+  //. > S.pipeK([S.tail, S.tail, S.head], S.Just([1, 2, 3, 4]))
+  //. Just(3)
+  //. ```
+  function pipeK(fs, x) {
+    return Z.reduce(function(x, f) { return Z.chain(f, x); }, x, fs);
+  }
+  S.pipeK =
+  def('pipeK',
+      {f: [Z.Foldable], m: [Z.Chain]},
+      [f(Fn($.Any, m($.Any))), m(a), m(b)],
+      pipeK);
 
   //# on :: (b -> b -> c) -> (a -> b) -> a -> a -> c
   //.
@@ -4636,11 +4659,13 @@
 //. [Apply]:            v:fantasyland/fantasy-land#apply
 //. [Bifunctor]:        v:fantasyland/fantasy-land#bifunctor
 //. [BinaryType]:       v:sanctuary-js/sanctuary-def#BinaryType
+//. [Chain]:            v:fantasyland/fantasy-land#chain
 //. [Either]:           #either-type
 //. [Extend]:           v:fantasyland/fantasy-land#extend
 //. [Fantasy Land]:     v:fantasyland/fantasy-land
 //. [Foldable]:         v:fantasyland/fantasy-land#foldable
 //. [Haskell]:          https://www.haskell.org/
+//. [Kleisli]:          https://en.wikipedia.org/wiki/Kleisli_category
 //. [Maybe]:            #maybe-type
 //. [Monad]:            v:fantasyland/fantasy-land#monad
 //. [Monoid]:           v:fantasyland/fantasy-land#monoid
--- a/test/pipeK.js
+++ b/test/pipeK.js
@@ -1,0 +1,29 @@
+'use strict';
+
+var S = require('./internal/sanctuary');
+
+var List = require('./internal/List');
+var eq = require('./internal/eq');
+
+
+var Cons = List.Cons;
+var Nil = List.Nil;
+
+
+test('pipeK', function() {
+
+  eq(typeof S.pipeK, 'function');
+  eq(S.pipeK.length, 2);
+  eq(S.pipeK.toString(), 'pipeK :: (Foldable f, Chain m) => f (Any -> m Any) -> m a -> m b');
+
+  eq(S.pipeK([], S.Just([1, 2, 3])), S.Just([1, 2, 3]));
+  eq(S.pipeK([S.tail], S.Just([1, 2, 3])), S.Just([2, 3]));
+  eq(S.pipeK([S.tail, S.tail], S.Just([1, 2, 3])), S.Just([3]));
+  eq(S.pipeK([S.tail, S.tail, S.head], S.Just([1, 2, 3])), S.Just(3));
+
+  eq(S.pipeK(Nil, S.Just([1, 2, 3])), S.Just([1, 2, 3]));
+  eq(S.pipeK(Cons(S.tail, Nil), S.Just([1, 2, 3])), S.Just([2, 3]));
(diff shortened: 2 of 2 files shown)


gabejohnson:
I'm seeing some linter warnings that I'm not sure how to fix.

index.js
```diff
@@ -1505,6 +1505,29 @@
   }
   S.pipe = def('pipe', {}, [$.Array($.AnyFunction), a, b], pipe);
 
+  //# pipeK :: Chain m => [(a -> m b), ..., (y -> m z)] -> m a -> m z
```
Avaq: Since it's built on `Z.reduce`, maybe the type can be generalised.

```hs
pipeK :: (Foldable f, Chain m) => f (Any -> m Any) -> m a -> m b
```

Not sure what to do about the `Any` stuff, but it's probably better than using `AnyFunction`.

index.js
```diff
@@ -4641,6 +4664,7 @@
 //. [Fantasy Land]:     v:fantasyland/fantasy-land
 //. [Foldable]:         v:fantasyland/fantasy-land#foldable
 //. [Haskell]:          https://www.haskell.org/
+//. [Kleisli]           https://en.wikipedia.org/wiki/Kleisli_category
```
davidchambers: This line is missing a colon (`:`). ;)

index.js
```diff
@@ -1505,6 +1505,29 @@
   }
   S.pipe = def('pipe', {}, [$.Array($.AnyFunction), a, b], pipe);
 
+  //# pipeK :: Chain m => [(a -> m b), ..., (y -> m z)] -> m a -> m z
+  //.
+  //. Takes an array of functions assumed to be unary which return
+  //. values with a [Chain][], and a value of that Chain, and returns the
+  //. result of applying the sequence of transformations to the initial value.
```
davidchambers: You'll need to add a link definition for `[Chain]` at the bottom of the file.

index.js
```diff
@@ -1505,6 +1505,29 @@
...
+  //.
+  //. In general terms, `pipeK` performs left-to-right [Kleisli][] composition
+  //. of an array of functions. `pipeK([f, g, h], x)` is equivalent to
+  //. `Z.chain(h, Z.chain(g, Z.chain(f, x)))`.
+  //.
+  //. ```javascript
+  //. > S.pipeK([S.tail, S.tail, S.head], S.Just([1,2,3,4]))
```
davidchambers: `[1, 2, 3, 4]`, please. :)

index.js
```diff
@@ -1505,6 +1505,29 @@
...
 
+  //# pipeK :: Chain m => [(a -> m b), ..., (y -> m z)] -> m a -> m z
+  //.
+  //. Takes an array of functions assumed to be unary which return
+  //. values with a [Chain][], and a value of that Chain, and returns the
+  //. result of applying the sequence of transformations to the initial value.
+  //.
+  //. In general terms, `pipeK` performs left-to-right [Kleisli][] composition
+  //. of an array of functions. `pipeK([f, g, h], x)` is equivalent to
+  //. `Z.chain(h, Z.chain(g, Z.chain(f, x)))`.
```
davidchambers: Let's use `chain` here rather than `Z.chain`.

test/pipeK.js
```diff
@@ -0,0 +1,20 @@
...
+var eq = require('./internal/eq');
+
+
+test('pipeK', function() {
+
+  eq(typeof S.pipeK, 'function');
+  eq(S.pipeK.length, 2);
+  eq(S.pipeK.toString(), 'pipeK :: Chain m => Array Function -> m a -> m b');
+
+  var xs = S.Just([1, 2, 3]);
```
davidchambers: I would prefer to avoid this assignment. I like assertions to require as little context as possible.

index.js
davidchambers: Good idea, Aldwin. Perhaps one of us should open a pull request to generalize `S.pipe`.

gabejohnson: I had considered this, but had two reservations:
1. I thought generalizing `pipe` and `pipeK` belonged in a separate PR
2. I wasn't certain generalizing would be useful. If `sanctuary-list` existed it would be a different story.

davidchambers: > I thought generalizing `pipe` and `pipeK` belonged in a separate PR

I agree. If we are to generalize, the neatest order of operations would be to generalize `pipe` in a separate pull request, merge that pull request, then rebase this pull request and generalize `pipeK`.

davidchambers: #473

davidchambers: The `pipe` pull request has been merged, so this pull request can now be updated. :)

davidchambers:
Gabe, would you mind changing the commit message prefix from `chain:` to `composition:` and updating the pull request title accordingly?

gabejohnson:
Done and done!

index.js
```diff
@@ -1505,6 +1505,29 @@
...
+  //. > S.pipeK([S.tail, S.tail, S.head], S.Just([1, 2, 3, 4]))
+  //. Just(3)
+  //. ```
+  function pipeK(fs, x) {
+    return Z.reduce(function(x, f) { return Z.chain(f, x); }, x, fs);
+  }
+  S.pipeK =
+  def('pipeK',
+      {m: [Z.Chain]},
+      [$.Array($.AnyFunction), m(a), m(b)],
```
davidchambers: Let's update the type of the first argument to `f($.Any, m($.Any))` as Aldwin suggested. :)

gabejohnson: 👍

index.js
```diff
@@ -1505,6 +1505,29 @@
...
+  //. ```javascript
+  //. > S.pipeK([S.tail, S.tail, S.head], S.Just([1, 2, 3, 4]))
+  //. Just(3)
+  //. ```
+  function pipeK(fs, x) {
+    return Z.reduce(function(x, f) { return Z.chain(f, x); }, x, fs);
+  }
+  S.pipeK =
+  def('pipeK',
+      {m: [Z.Chain], f: [Z.Foldable]},
```
davidchambers: Let's switch `m` and `f` to match the order in which they appear in the type signature.

index.js
```diff
@@ -1505,6 +1505,29 @@
   }
   S.pipe = def('pipe', {f: [Z.Foldable]}, [f(Fn($.Any, $.Any)), a, b], pipe);
 
+  //# pipeK :: (Foldable f, Chain m) => f (Any -> m Any) -> m a -> m b
+  //.
+  //. Takes an array of functions assumed to be unary which return
+  //. values with a [Chain][], and a value of that Chain, and returns the
+  //. result of applying the sequence of transformations to the initial value.
```
davidchambers: We could tweak the line wrapping to make the lengths of these lines more balanced:

```javascript
  //. Takes an array of functions assumed to be unary which return values
  //. with a [Chain][], and a value of that Chain, and returns the result
  //. of applying the sequence of transformations to the initial value.
```

index.js
```diff
@@ -1505,6 +1505,29 @@
...
+  //.
+  //. In general terms, `pipeK` performs left-to-right [Kleisli][] composition
+  //. of an array of functions. `pipeK([f, g, h], x)` is equivalent to
+  //. `chain(h, chain(g, chain(f, x)))`.
```
davidchambers: Let's replace “an array of functions” with “a sequence of functions” as we did in #473.

index.js
gabejohnson: Fixed

index.js
gabejohnson: Fixed

test/pipeK.js
```diff
@@ -0,0 +1,19 @@
...
+  eq(typeof S.pipeK, 'function');
+  eq(S.pipeK.length, 2);
+  eq(S.pipeK.toString(), 'pipeK :: (Foldable f, Chain m) => f (Any -> m Any) -> m a -> m b');
+
+  eq(S.pipeK([], S.Just([1, 2, 3])), S.Just([1, 2, 3]));
+  eq(S.pipeK([S.tail], S.Just([1, 2, 3])), S.Just([2, 3]));
+  eq(S.pipeK([S.tail, S.tail], S.Just([1, 2, 3])), S.Just([3]));
+  eq(S.pipeK([S.tail, S.tail, S.head], S.Just([1, 2, 3])), S.Just(3));
+
+});
```
davidchambers: Let's add a block of tests for `List (Any -> Maybe Any)` much as we did for `pipe`.

gabejohnson: ⚡️

index.js
```diff
@@ -1505,6 +1505,29 @@
   }
   S.pipe = def('pipe', {f: [Z.Foldable]}, [f(Fn($.Any, $.Any)), a, b], pipe);
 
+  //# pipeK :: (Foldable f, Chain m) => f (Any -> m Any) -> m a -> m b
+  //.
+  //. Takes an array of functions assumed to be unary which return values
```
gabejohnson: Should this line reference "sequence" instead of "array" as well?

davidchambers: I mentioned this in <https://github.com/sanctuary-js/sanctuary/pull/473#discussion_r159290807>. I don't like using “structure” here, but perhaps “sequence” is better than “array” (which is, admittedly, inaccurate). I don't have a strong opinion on this. I'm happy to stick with “array” or to change “array” to “sequence” in both descriptions. I leave this to you, Gabe.

gabejohnson: Sequences can be limited to a single element 😄

davidchambers: I don't understand your previous comment, Gabe. :thinking:

PR #473, opened 2018-01-02
composition: generalize S.pipe

davidchambers:
This change has been discussed in #472.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1487,13 +1487,13 @@
       [s(b, c), s(a, b), s(a, c)],
       Z.compose);
 
-  //# pipe :: [(a -> b), (b -> c), ..., (m -> n)] -> a -> n
+  //# pipe :: Foldable f => f (Any -> Any) -> a -> b
   //.
   //. Takes an array of functions assumed to be unary and a value of any type,
   //. and returns the result of applying the sequence of transformations to
   //. the initial value.
   //.
-  //. In general terms, `pipe` performs left-to-right composition of an array
+  //. In general terms, `pipe` performs left-to-right composition of a sequence
   //. of functions. `pipe([f, g, h], x)` is equivalent to `h(g(f(x)))`.
   //.
   //. ```javascript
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1487,13 +1487,13 @@
       [s(b, c), s(a, b), s(a, c)],
       Z.compose);
 
-  //# pipe :: [(a -> b), (b -> c), ..., (m -> n)] -> a -> n
+  //# pipe :: Foldable f => f (Any -> Any) -> a -> b
   //.
   //. Takes an array of functions assumed to be unary and a value of any type,
   //. and returns the result of applying the sequence of transformations to
   //. the initial value.
```
davidchambers: I *think* I'm happy to continue to refer to _an array of functions_ because replacing “array” with “structure” is awkward.

CrossEye:
I can't really test right now, but Sanctuary likes to fail early with good messages.  What would happen with this?:

```js
S.pipe(Cons(S.add(2), Nil), true)
```

Because as far as I can see, there is no violation of the definitions.  Does it fail later than other functions would?

davidchambers:
In general we can't really know a function's type (I'm keen to avoid treating functions created via `def` specially). All we can do is inspect inputs and outputs at run time in order to eliminate candidate types.

For `pipe` there's an additional complication: the function is not well typed. We want to be able to use `pipe` to compose `a -> b`, `b -> c`, and `c -> d`, but sanctuary-def does not permit heterogeneous lists. `Foldable f => f (Any -> Any)` is a necessary fudge.

Avaq:
@CrossEye `S.pipe` relies on type checks performed by the functions in the pipeline:

```js
S.pipe([S.add(1)], null)
```
```text
! Invalid value

add :: FiniteNumber -> FiniteNumber -> FiniteNumber
                       ^^^^^^^^^^^^
                            1

1)  null :: Null

The value at position 1 is not a member of ‘FiniteNumber’.

See https://github.com/sanctuary-js/sanctuary-def/tree/v0.14.0#FiniteNumber for information about the sanctuary-def/FiniteNumber type.
```

We do get a decent error message, but only after applying `pipe` to its second argument. I don't think this PR changes any of that.

CrossEye:
@davidchambers, @Avaq:  That makes sense.  Thanks!!

PR #475, opened 2018-01-09
sanctuary-type-classes@8.0.1

### LINKED ISSUE #338: Add reject and rejectM
Add inverse versions of [filter](https://github.com/sanctuary-js/sanctuary#filter--applicativeffoldablefmonoidfa--a-boolean---fa---fa) and [filterM](https://github.com/sanctuary-js/sanctuary#filterm--monadmmonoidma--a-boolean---ma---ma).

davidchambers:
Closes #338

<https://github.com/sanctuary-js/sanctuary-type-classes/releases/tag/v8.0.0>


index.js
```diff
@@ -786,6 +786,58 @@
   //. ```
   S.invert = def('invert', {g: [Z.Group]}, [g, g], Z.invert);
 
+  //# filter :: Filterable f => (a -> Boolean) -> f a -> f a
+  //.
+  //. Curried version of [`Z.filter`][]. Discards every element of the given
+  //. structure which does not satisfy the predicate.
```
Avaq: Anywhere in the docs you're currently using "structure" you could be more specific by replacing it by its type class(es).

> Discards every element of the given Filterable which does not satisfy the predicate.

davidchambers: That's true, although the approach only works when there is exactly one constraint on the structure. Admittedly, this is quite often the case.

On the website, constraints in type signatures are hyperlinked, reducing the importance of links to type classes in descriptions.

I consider the above “defences” quite weak. Do you find either of them compelling, Aldwin, or do you think we should replace “structure” in contexts in which it is possible to do so?

Avaq: > the approach only works when there is exactly one constraint on the structure

In theory you could use terms like *Filterable Monoid* and *Monadic Semigroup* when having multiple constraints on one type variable, and terms like *Functor of Foldable* when talking about constraints on layered structures. In practice, I'm not sure if this will make things more clear.

> Takes a Setoidal value and a Foldable of Setoids and returns true iff the value
is an element of the Foldable.

That starts to look like we're communicating with another species.

davidchambers:
I just pushed some significant changes to this pull request. I'd appreciate another review from you, @gabejohnson, or from another member of the community. :)

index.js
gabejohnson: The context for the description is already provided by the function signature itself. Perhaps we could just omit "of the given structure".

index.js
```diff
@@ -786,6 +786,98 @@
...
+  //. > S.reject(S.odd, S.Just(1))
+  //. Nothing
+  //. ```
+  S.reject =
+  def('reject', {f: [Z.Filterable]}, [$.Predicate(a), f(a), f(a)], Z.reject);
+
+  //# takeWhile :: Filterable f => (a -> Boolean) -> f a -> f a
+  //.
+  //. Curried version of [`Z.takeWhile`][]. Discards the first inner value
+  //. which does not satisfy the predicate, and all subsequent inner values.
```
gabejohnson: This description isn't consistent with the previous two. Perhaps replace "inner value" with "element"?

index.js
davidchambers: > Perhaps we could just omit "of the given structure".

:zap:

index.js
davidchambers: Excellent suggestion! I love these small improvements. :)

:zap:

PR #484, opened 2018-01-30
array: remove S.pluck

davidchambers:
I don't see a compelling reason to provide a shorthand for `map` + `prop` but not for `map` + `get`. I also don't find the shorthand clearer:

```javascript
S.map (S.prop ('foo'))
```

```javascript
S.pluck ('foo')
```

The problem with the shorthand is that I forget which of the two possible combinations it is short for. ;)


Avaq:
> I don't see a compelling reason to provide a shorthand for `map` + `prop` but not for `map` + `get`.

Wait.. what? There are either one too many negatives in this sentence, or a "but" where there should be an "and".

davidchambers:
I *think* what I wrote makes sense, Aldwin. I was trying to state that I don't see a reason to do A but not B. In other words, if we do A we should also do B. The implication is that in this case doing both A and B is silly, so we should not do either.

Too much negation, though, I agree. :)

Avaq:
Ahh. I was reading:

> I don't see a compelling reason to (provide a shorthand for `map` + `prop`) (but not for `map` + `get`).

But you meant:

> I don't see a compelling reason to provide a shorthand for (`map` + `prop`) (but not for `map` + `get`).

I read it so many times over and missed that. Okay. All clear now! ;)

davidchambers:
This (is (why)) (I (am)) (so) (fond) (of (parentheses)). ;)

miwillhite:
I used `map(prop('foo'))` the other day thinking that Sanctuary wouldn't have `pluck` (it seems a bit contrary to the simplistic nature of Sanctuary)…then was surprised when I scrolled past it later. I didn't update my code…so I think 👍  from me too ;)

stefano-b2c2:
Ha. You've had it in for this function from pretty much day one, David.

PR #486, opened 2018-02-10
sanctuary-type-classes@8.1.1

### LINKED ISSUE #476: add Either.mapLeft
equivalent to:

```js
S.bimap( S.__, S.I )
```

trivial but handy

davidchambers:
Closes #476

<https://github.com/sanctuary-js/sanctuary-type-classes/releases/tag/v8.1.0>

### DIFF
--- a/index.js
+++ b/index.js
@@ -934,6 +934,21 @@
       [Fn(a, b), Fn(c, d), p(a, c), p(b, d)],
       Z.bimap);
 
+  //# mapLeft :: Bifunctor f => (a -> b) -> f a c -> f b c
+  //.
+  //. Curried version of [`Z.mapLeft`][]. Maps the given function over the left
+  //. side of a Bifunctor.
+  //.
+  //. ```javascript
+  //. > S.mapLeft(S.toUpper, S.Left('foo'))
+  //. Left('FOO')
+  //.
+  //. > S.mapLeft(S.toUpper, S.Right(64))
+  //. Right(64)
+  //. ```
+  S.mapLeft =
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -929,6 +929,21 @@
...
+  //.
+  //. ```javascript
+  //. > S.mapLeft(S.toUpper, S.Left('foo'))
+  //. Left('FOO')
+  //.
+  //. > S.mapLeft(S.toUpper, S.Right(64))
+  //. Right(64)
+  //. ```
+  S.mapLeft =
+  def('mapLeft', {p: [Z.Bifunctor]}, [Fn(a, b), p(a, c), p(b, c)], Z.mapLeft);
```
davidchambers: We're using the “wrong” letter for the binary type variable, because `f` is defined as a unary type variable. We *could* do this:

```javascript
S.mapLeft = (function() {
  var f = $.BinaryTypeVariable('f');
  return def('mapLeft', ...);
}());
```

This strikes me as quite ugly, so I'm inclined to stick with the status quo. If one day we decide to define each function in a separate file we could switch to the “correct” letters at that point. I'll continue to fight to keep the Sanctuary source code a single ES5 file, of course. :)

Avaq: I think it's fine. Give `p` a chance to shine.

gabejohnson: I also think `p` is fine. Assuming `s` and `p` stand for Semigroupoid and Profunctor, respectively, you could always define `bi` for Bifunctor.

Another option would be to get rid of `s` and just use `p` for Semigroupoid, Profunctor, and Bifunctor

Issue #489, opened 2018-02-20
mapMaybe on something else than an array ?


gautaz:
Hello,

Currently this is available:
```js
S.map(S.get(S.is(Number), 'c'))({a: {c: 1}, b: {}}); // {a: Just(1), b: Nothing}
```

But this is not:
```js
S.mapMaybe(S.get(S.is(Number), 'c'))({a: {c: 1}, b: {}}); // would return {a: 1}
```

I was wondering if `mapMaybe` could be extended to operate the same way as `map`?

davidchambers:
Great idea, @gautaz! I'll open a pull request for this improvement. :)

Avaq:
`mapMaybe` is currently implemented as `justs(map(f, xs))`, and `justs` is specialised in Array. Now that we have Filterable, maybe we can generalise `justs` and any function implemented in terms of it.

```diff
-justs :: Array (Maybe a) -> Array a
+justs :: Filterable f => f (Maybe a) -> f a
```

PR #490, opened 2018-02-20
filterable: generalize S.justs, S.lefts, and S.rights

davidchambers:
<https://github.com/sanctuary-js/sanctuary/issues/489#issuecomment-366982183>

### DIFF
--- a/index.js
+++ b/index.js
@@ -244,6 +244,11 @@
     };
   }
 
+  //  value :: { value :: a } -> a
+  function value(r) {
+    return r.value;
+  }
+
   //  readmeUrl :: String -> String
   function readmeUrl(id) {
     var version = '0.14.1';  // updated programmatically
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -2173,10 +2178,10 @@
   }
   S.maybe_ = def('maybe_', {}, [$.Thunk(b), Fn(a, b), $Maybe(a), b], maybe_);
 
-  //# justs :: Array (Maybe a) -> Array a
+  //# justs :: (Filterable f, Functor f) => f (Maybe a) -> f a
   //.
-  //. Takes an array of Maybes and returns an array containing each Just's
-  //. value. Equivalent to Haskell's `catMaybes` function.
+  //. Discards each of the structure's elements which is Nothing, and unwraps
```
Avaq: It was recently discussed that we could omit mentions of *the structure* from these descriptions, eg: *Discards each element which is Nothing...*; Will we do that in a separate PR?

davidchambers: I've updated the descriptions. Please have another look. :eyes:

davidchambers:
> Will you make the changes to the signatures of derived functions in a separate PR?

I believe the only function derived from one of these functions is [`S.mapMaybe`][1]. I will open a pull request to generalize that function once this pull request has been merged. :)


[1]: https://sanctuary.js.org/#mapMaybe

PR #492, opened 2018-02-25
add S.get_, S.gets_, and S.parseJson_

### LINKED ISSUE #481: variants of S.get, S.gets, and S.parseJson which use $.test
Using [`$.test`][1] in conjunction with [`S.get`][2], [`S.gets`][3], or [`S.parseJson`][4] is incredibly powerful. Consider these examples:

```javascript
> S.parseJson ($.test ([]) ($.Array ($.Integer))) ('[1,2')
Nothing

> S.parseJson ($.test ([]) ($.Array ($.Integer))) ('[1,2,3]')
Just ([1, 2, 3])

> S.parseJson ($.test ([]) ($.Array ($.Integer))) ('[1,2,3,0.5]')
Nothing
```

We could provide `_`-suffixed variants for convenience:

```haskell
get_ :: Type -> String -> Any -> Maybe a
gets_ :: Type -> Array String -> Any -> Maybe a
parseJson_ :: Type -> String -> Maybe a
```

One could then write:

```javascript
> S.parseJson_ ($.Array ($.Integer)) ('[1,2')
Nothing

> S.parseJson_ ($.Array ($.Integer)) ('[1,2,3]')
Just ([1, 2, 3])

> S.parseJson_ ($.Array ($.Integer)) ('[1,2,3,0.5]')
Nothing
```

:thumbsup: / :thumbsdown: ?

[1]: https://github.com/sanctuary-js/sanctuary-def#test
[2]: https://sanctuary.js.org/#get
[3]: https://sanctuary.js.org/#gets
[4]: https://sanctuary.js.org/#parseJson

davidchambers:
Closes #481

/cc @miwillhite, @najlachamseddine

### DIFF
--- a/index.js
+++ b/index.js
@@ -3708,7 +3708,7 @@
   //. value of the specified object property if it exists and the value
   //. satisfies the given predicate; Nothing otherwise.
   //.
-  //. See also [`gets`](#gets) and [`prop`](#prop).
+  //. See also [`get_`](#get_), [`gets`](#gets), and [`prop`](#prop).
   //.
   //. ```javascript
   //. > S.get(S.is(Number), 'x', {x: 1, y: 2})
@@ -3718,12 +3718,6 @@
   //. Nothing
   //.
   //. > S.get(S.is(Number), 'x', {})
-  //. Nothing
-  //.
-  //. > S.get($.test([], $.Array($.Number)), 'x', {x: [1, 2, 3]})
-  //. Just([1, 2, 3])
-  //.
-  //. > S.get($.test([], $.Array($.Number)), 'x', {x: [1, 2, 3, null]})
   //. Nothing
   //. ```
   function get(pred, key, x) {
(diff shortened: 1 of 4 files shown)


Avaq:
I'm sorry I missed the discussion around this a little, but I'm not sure I would use these. I'm not too keen on the functions with underscores that save a few characters of typing through a simple composition - if I find myself reusing `$.test` a lot I would make this composition myself. I find that easier than remembering that there is a variant of the function that already does this.

```js
> const get = S.compose (S.get) ($.test (S.env))
> get ($.Array ($.Integer)) ('x') ({x: ['nope']})
Nothing
```

That said, this won't get in my way, and it looks good to me, so I'm okay with including it.

davidchambers:
> I'm not too keen on the functions with underscores that save a few characters of typing through a simple composition

I agree with this sentiment. With these specific functions, though, we would gently push users in the right direction (it's telling that the [`S.parseJson`][1] examples use `$.test`). I don't want to *require* a `Type` value, but much of the time the best way to define a predicate is to first define a `Type` value.

It's not just a matter of convenience: [aeson][2] forever changed the way I think about interacting with JSON strings; I'd like `S.parseJson_` to affect people's thinking in a similar way.


[1]: https://sanctuary.js.org/#parseJson
[2]: https://hackage.haskell.org/package/aeson

davidchambers:
#513 greatly improved the usability of [`S.get`][1], [`S.gets`][2], and [`S.parseJson`][3], so variants of these functions are no longer necessary.


[1]: https://sanctuary.js.org/#get
[2]: https://sanctuary.js.org/#gets
[3]: https://sanctuary.js.org/#parseJson

Issue #438, opened 2017-09-01
readable code without Ramda-style currying


davidchambers:
Sanctuary is defined in part by what it does *not* support. We have done a good job of managing complexity and entropy, and we must continue to do so if Sanctuary is to live a long, healthy life.

Ramda-style currying—the ability to write both `f(x)(y)` and `f(x, y)`—is a source of complexity. I've seen this complexity as necessary to prevent code written with Sanctuary from looking strange to newcomers, which would limit the library's initial appeal and thus limit the library's adoption.

Last night it occurred to me that we could possibly solve (or at least mitigate) the "`)(`" problem by tweaking the way in which we format function applications.

### The "`)(`" problem

In JavaScript, this reads very naturally:

```javascript
f(x, y, z)
```

This, on the other hand, seems unnatural:

```javascript
f(x)(y)(z)
```

A day ago my impression was that the only aesthetic problem was having opening parens follow closing parens. I now see a second aesthetic problem, as I hope this example demonstrates:

```javascript
f(g(x)(y))(h(z))
```

There's no *space*. There's a significant difference visually between `x)(y` and `x, y`. The nesting of subexpressions above is not immediately clear to a human reader. When we include space between arguments—as is common practice in JavaScript—the nesting is clear:

```javascript
f(g(x, y), h(z))
```

This clarity is the primary benefit of Ramda-style currying. I consider `S.concat(x)(y)` bad style not because of the `)(` but because if used consistently this style results in expressions which are less clear than their more spacious equivalents.

It's worth noting that multiline function applications are also natural with the comma style:

```javascript
f(x,
  y,
  z)
```

`x`, `y`, and `z` are obviously placeholders for longer expressions in this case.

Here's the `)(`-style equivalent:

```javascript
f(x)
 (y)
 (z)
```

My concern is that visually the `x` is more tightly bound to `f` than it is to `y` and `z`, making the first argument feel privileged in some way.

### Learning from Haskell

Sanctuary brings many good ideas from Haskell to JavaScript. Perhaps most important is the combination of curried functions and partial application. We might be able to learn from Haskell's approach to function application.

In Haskell, function application is considered so important that a space is all it requires syntactically: `f x` in Haskell is equivalent to `f(x)` in JavaScript. The associativity of function application is such that `f x y` is equivalent to `(f x) y`, which is to say that what we write as `f(x)(y)` in JavaScript could simply be written `f x y` in Haskell.

Let's consider how the previous examples would look in Haskell:

[...]

PR #501, opened 2018-03-12
remove flipped variants of binary operator functions

davidchambers:
<https://github.com/sanctuary-js/sanctuary/issues/438#issuecomment-371809619>


gabejohnson:
I'm slightly concerned that we're going to end up in the state we were in prior to on #388.  Perhaps if there's an entry in the wiki it won't be an problem.

davidchambers:
> Based on the comment you linked it seems you will fix the signatures in a later commit, but I'll mention it in case you wanted to do it in this commit and forgot.

Thanks, Aldwin. It's best not to assume everything I do is intentional. :)

In this case, I think we should stick with the `_ -> (_ -> _)` types until we update sanctuary-def in an upcoming pull request (at which point *all* functions will adopt this stricter behaviour). `S.lt (0, 1)` is currently a type error, and it will be a type error once simple currying lands. I'd like to avoid a window in which it is valid (evaluating to `false`).

davidchambers:
> I'm slightly concerned that we're going to end up in the state we were in prior to on #388.

Could you elaborate, Gabe? Your memory of the discussion preceding that pull request is likely better than mine. What are you concerned that users will incorrectly assume if we merge this pull request?

gabejohnson:
I'm referring to the issues/comments linked to in https://github.com/sanctuary-js/sanctuary/pull/388#pullrequestreview-35532373. In the past, there has been a lot of confusion around argument order for binary operators.

Although their behavior is documented, I suspect the first time a new user reaches for `S.lt` they will assume the wrong argument order. Perhaps calling out the non-commutative "operators" individually or putting them in their own subsection would help.

davidchambers:
> I suspect the first time a new user reaches for `S.lt` they will assume the wrong argument order.

Even though `S.lt (x, y)` will not be supported, Gabe? Feedback from Ramda and Sanctuary users gives me the impression is that the majority assume `S.lt (0)` to define the _less than zero_ function.

PR #511, opened 2018-03-21
array: add S.zip and S.zipWith

### LINKED ISSUE #337: Add S.zip and S.zipWith
Create a type-safe version of [zip](http://ramdajs.com/0.22.1/docs/#zip) and [zipWith](http://ramdajs.com/0.22.1/docs/#zipWith).

sadasant:
Work in progress for zipWith!
I started with Ramda’s code (a single while) but switched to Z.reduce to take in count it’s capabilities over different types.

<del>Related to #337</del> <ins>Closes #337</ins>

Please give me feedback! I’m ready to improve this 🤗

### DIFF
--- a/index.js
+++ b/index.js
@@ -216,6 +216,13 @@
   //  Fn :: (Type, Type) -> Type
   function Fn(x, y) { return $.Function([x, y]); }
 
+  //  pair :: a -> b -> Pair a b
+  function pair(x) {
+    return function(y) {
+      return [x, y];
+    };
+  }
+
   //  toObject :: a -> Object
   function toObject(x) {
     return x == null ? Object.create(null) : Object(x);
@@ -3585,6 +3592,57 @@
       {b: [Z.Ord], m: [Z.Applicative, Z.Foldable, Z.Monoid]},
       [Fn(a, b), m(a), m(a)],
       Z.sortBy);
+
+  //# zip :: Array a -> Array b -> Array (Pair a b)
+  //.
+  //. Returns an array of pairs of corresponding elements from the given
+  //. arrays. The length of the resulting array is equal to the length of
+  //. the shorter input array.
+  //.
+  //. See also [`zipWith`](#zipWith).
+  //.
+  //. ```javascript
+  //. > S.zip(['a', 'b'], ['x', 'y', 'z'])
+  //. [['a', 'x'], ['b', 'y']]
+  //.
+  //. > S.zip([1, 3, 5], [2, 4])
+  //. [[1, 2], [3, 4]]
+  //. ```
+  function zip(xs, ys) {
+    return zipWith(pair, xs, ys);
+  }
+  S.zip =
+  def('zip',
+      {},
+      [$.Array(a), $.Array(b), $.Array($.Pair(a, b))],
+      zip);
+
+  //# zipWith :: (a -> b -> c) -> Array a -> Array b -> Array c
+  //.
+  //. Returns the result of combining, pairwise, the given arrays using the
+  //. given binary function. The length of the resulting array is equal to the
+  //. length of the shorter input array.
+  //.
+  //. See also [`zip`](#zip).
+  //.
+  //. ```javascript
+  //. > S.zipWith(a => b => a + b, ['a', 'b'], ['x', 'y', 'z'])
+  //. ['ax', 'by']
+  //.
+  //. > S.zipWith(a => b => [a, b], [1, 3, 5], [2, 4])
+  //. [[1, 2], [3, 4]]
+  //. ```
+  function zipWith(f, xs, ys) {
+    var result = [];
+    var len = Math.min(xs.length, ys.length);
+    for (var idx = 0; idx < len; idx += 1) result.push(f(xs[idx])(ys[idx]));
+    return result;
+  }
+  S.zipWith =
+  def('zipWith',
+      {},
+      [Fn(a, Fn(b, c)), $.Array(a), $.Array(b), $.Array(c)],
+      zipWith);
 
   //. ### Object
 
--- a/test/internal/pair.js
+++ b/test/internal/pair.js
@@ -1,0 +1,8 @@
+'use strict';
+
+//  pair :: a -> b -> Pair a b
+module.exports = function pair(x) {
+  return function(y) {
+    return [x, y];
+  };
+};
--- a/test/zipWith.js
+++ b/test/zipWith.js
@@ -1,0 +1,19 @@
+'use strict';
+
+var S = require('..');
+
+var concat = require('./internal/concat');
+var eq = require('./internal/eq');
+var pair = require('./internal/pair');
+
+
+test('zipWith', function() {
+
+  eq(typeof S.zipWith, 'function');
+  eq(S.zipWith.length, 3);
+  eq(S.zipWith.toString(), 'zipWith :: (a -> b -> c) -> Array a -> Array b -> Array c');
+
+  eq(S.zipWith(concat, ['a', 'b'], ['x', 'y', 'z']), ['ax', 'by']);
+  eq(S.zipWith(pair, [1, 3, 5], [2, 4]), [[1, 2], [3, 4]]);
+
+});
--- a/test/zip.js
+++ b/test/zip.js
@@ -1,0 +1,17 @@
+'use strict';
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+
+
+test('zip', function() {
+
+  eq(typeof S.zip, 'function');
+  eq(S.zip.length, 2);
+  eq(S.zip.toString(), 'zip :: Array a -> Array b -> Array (Pair a b)');
+
+  eq(S.zip(['a', 'b'], ['x', 'y', 'z']), [['a', 'x'], ['b', 'y']]);
+  eq(S.zip([1, 3, 5], [2, 4]), [[1, 2], [3, 4]]);
+
+});


davidchambers:
Thanks for submitting this pull request, @sadasant. :)

index.js
```diff
@@ -3656,6 +3656,42 @@
       [Fn(a, b), m(a), m(a)],
       Z.sortBy);
 
+  //# zipWith :: (Applicative m, Foldable m, Monoid m) => (a -> b -> c) -> m a -> m b -> Arrray c
```
davidchambers: There are two types which seem sensible to me:

  - `(a -> b -> c) -> Array a -> Array b -> Array c`
  - `(a -> b -> c) -> m a -> m b -> m c` (with appropriate constraints on `m`)

Supporting various input types but fixing the output type does not feel right to me.

index.js
```diff
@@ -3656,6 +3656,42 @@
...
+  //. > S.zipWith(S.add, [1, 2, 3], [3, 2, 1])
+  //. [4, 4, 4]
+  //.
+  //. > S.zipWith(a => b => [a, b], [1, 3, 5], [2, 4, 6])
+  //. [[1, 2], [3, 4], [5, 6]]
+  //.
+  //. > S.zipWith(a => b => Cons(a, Cons(b, Nil)), Cons('foo', Nil), Cons('bar', Nil))
+  //. [Cons('foo', Cons('bar', Nil))]
+  //. ```
+  function zipWith(pred, a, b) {
```
davidchambers: A value of type `a -> b -> c` is not a predicate, so `pred` is a confusing identifier. Let's use `f` instead. :)

index.js
sadasant: Yes! How can we do the later though? I tried to use S.type wth S.of but the type class in S.type is a string. Any tip?

index.js
sadasant: thanks!

index.js
```diff
@@ -3656,6 +3656,42 @@
...
+  //.
+  //. > S.zipWith(a => b => [a, b], [1, 3, 5], [2, 4, 6])
+  //. [[1, 2], [3, 4], [5, 6]]
+  //.
+  //. > S.zipWith(a => b => Cons(a, Cons(b, Nil)), Cons('foo', Nil), Cons('bar', Nil))
+  //. [Cons('foo', Cons('bar', Nil))]
+  //. ```
+  function zipWith(f, a, b) {
+    var idy = 0;
+    var aIsShorter = S.size(a) < S.size(b);
```
davidchambers: This is inefficient because `S.size` traverses the whole structure to determine its size (setting aside the special case for arrays). Additionally, we avoid using `S` functions internally to avoid additional type checking.

A good starting point would be to convert both structures to arrays:

```javascript
//  toArray :: Foldable f => f a -> Array a
function toArray(foldable) {
  return Z.reduce(function(xs, x) { xs.push(x); return xs; }, [], foldable);
}
```

```javascript
function zipWith(f, xs, ys) {
  var $xs = toArray(xs);
  var $ys = toArray(ys);
  ...
}
```

To avoid using `aIsShorter ? ... : ...` in four places (including twice in a reducing function), we could truncate the longer array. To avoid allocating two new arrays, we could update the `length` properties like so:

```javascript
function zipWith(f, xs, ys) {
  var $xs = toArray(xs);
  var $ys = toArray(ys);
  $xs.length = $ys.length = Math.min($xs.length, $ys.length);
  ...
}
```

From this point the code for combining `$xs` and `$ys` should be relatively simple. :)

sadasant: Gotcha! I didn’t know about to array!!! That’s really nice!

sadasant: Wait, toArray doesn’t exist! 😱 I’ll define it as you put it 🤔

sadasant: Should we expose toArray?

davidchambers: One day, perhaps, but let's stay focused on the task at hand. Define `toArray` near the top of the file with the other private functions. ;)

test/zipWith.js
```diff
@@ -0,0 +1,23 @@
...
+  eq(typeof S.zipWith, 'function');
+  eq(S.zipWith.length, 3);
+  eq(S.zipWith.toString(), 'zipWith :: (Applicative m, Foldable m, Monoid m) => (a -> b -> c) -> m a -> m b -> Array c');
+
+  eq(S.zipWith(S.add, [1, 2, 3], [3, 2, 1]), [4, 4, 4]);
+  eq(S.zipWith(S.add, [1, 2, 3, 4], [3, 2, 1]), [4, 4, 4]);
+  eq(S.zipWith(S.add, [1, 2, 3], [3, 2, 1, 0]), [4, 4, 4]);
+  eq(S.zipWith(a => b => [a, b], [1, 3, 5], [2, 4, 6]), [[1, 2], [3, 4], [5, 6]]);
+  eq(S.zipWith(a => b => Cons(a, Cons(b, Nil)), Cons('foo', Nil), Cons('bar', Nil)), [Cons('foo', Cons('bar', Nil))]);
+
```
syaiful6: should also test with different array length..

index.js
```diff
@@ -3656,6 +3661,42 @@
...
+  function zipWith(f, a, b) {
+    var rv = [];
+    var idx = 0;
+    var as = toArray(a);
+    var bs = toArray(b);
+    var len = Math.min(as.length, bs.length);
+    while (idx < len) {
+      rv[idx] = f(as[idx])(bs[idx]);
+      idx += 1;
+    }
```
davidchambers: I don't like having `idx` declared so far from its place of use. I would prefer a `for` loop:

```javascript
for (var idx = 0, min = Math.min(as.length, bs.length); idx < min; idx += 1) {
  ...
}
```

To appease the linter we may need to move one of the declarations:

```javascript
var min = Math.min(as.length, bs.length);
for (var idx = 0; idx < min; idx += 1) {
  ...
}
```

index.js
```diff
@@ -338,6 +338,11 @@
   //  Options :: Type
   var Options = $.RecordType({checkTypes: $.Boolean, env: $.Array($.Any)});
 
+  //  toArray :: Foldable f => f a -> Array a
```
davidchambers: Please move this even further up the file. Just before `toObject` would be ideal.

test/zipWith.js
syaiful6: oh, yes, that's good then. Look like i am bit tired.

index.js
sadasant: On my way!

index.js
```diff
@@ -3656,6 +3661,42 @@
...
+  //. [4, 4, 4]
+  //.
+  //. > S.zipWith(a => b => [a, b], [1, 3, 5], [2, 4, 6])
+  //. [[1, 2], [3, 4], [5, 6]]
+  //.
+  //. > S.zipWith(a => b => Cons(a, Cons(b, Nil)), Cons('foo', Nil), Cons('bar', Nil))
+  //. Cons(Cons('foo', Cons('bar', Nil)), Nil)
+  //. ```
+  function zipWith(f, a, b) {
+    var rv = Z.empty(a.constructor);
```
davidchambers: Let's write `var A = a.constructor;` to avoid the property access in the loop.

index.js
```diff
@@ -3656,6 +3661,42 @@
...
+  //.
+  //. ```javascript
+  //. > S.zipWith(S.add, [1, 2, 3], [3, 2, 1])
+  //. [4, 4, 4]
+  //.
+  //. > S.zipWith(a => b => [a, b], [1, 3, 5], [2, 4, 6])
+  //. [[1, 2], [3, 4], [5, 6]]
+  //.
+  //. > S.zipWith(a => b => Cons(a, Cons(b, Nil)), Cons('foo', Nil), Cons('bar', Nil))
+  //. Cons(Cons('foo', Cons('bar', Nil)), Nil)
```
davidchambers: I find this example confusing. I'd prefer to see something like this:

```javascript
> S.zipWith(S.pow, Cons(1, Cons(2, Cons(3, Nil))), Cons(10, Cons(20, Nil)))
Cons(10, Cons(400, Nil))
```

index.js
```diff
@@ -3656,6 +3656,43 @@
...
+  function zipWith(f, a, b) {
+    var rv = [];
+    var idx = 0;
+    var len = Math.min(a.length, b.length);
+    while (idx < len) {
+      rv[idx] = f(a[idx])(b[idx]);
+      idx += 1;
+    }
+    return rv;
+  }
```
davidchambers: Let's use this formatting for consistency with existing code:

```javascript
function zipWith(f, xs, ys) {
  var result = [];
  var len = Math.min(xs.length, ys.length);
  for (var idx = 0; idx < len; idx += 1) result.push(f(xs[idx])(ys[idx]));
  return result;
}
```

index.js
```diff
@@ -3656,6 +3656,43 @@
...
+  //. [4, 4, 4]
+  //.
+  //. > S.zipWith(S.add, [1, 2, 3, 4], [3, 2, 1])
+  //. [4, 4, 4]
+  //.
+  //. > S.zipWith(S.add, [1, 2, 3], [3, 2, 1, 0])
+  //. [4, 4, 4]
+  //.
+  //. > S.zipWith(S.on(S.concat, S.of(Array)), [1, 3, 5], [2, 4, 6])
+  //. [[1, 2], [3, 4], [5, 6]]
```
davidchambers: `S.on(S.concat, S.of(Array))` is far too complex for a doctest. Let's use `x => y => [x, y]` instead.

index.js
```diff
@@ -3656,6 +3656,43 @@
...
+  //.
+  //. Creates a new array out of the two supplied arrays
+  //. by applying the given function to each equally-positioned
+  //. pair of elements from each given array. The returning
+  //. array is truncated to the length of the shorter of the two
+  //. input arrays.
+  //.
+  //. ```javascript
+  //. > S.zipWith(S.add, [1, 2, 3], [3, 2, 1])
+  //. [4, 4, 4]
```
davidchambers: Let's drop this example. It's clear from the following examples what should happen when the arrays are of equal length.

index.js
```diff
@@ -3656,6 +3656,43 @@
       [Fn(a, b), m(a), m(a)],
       Z.sortBy);
 
+  //# zipWith :: (a -> b -> c) -> Array a -> Array b -> Array c
+  //.
+  //. Creates a new array out of the two supplied arrays
+  //. by applying the given function to each equally-positioned
+  //. pair of elements from each given array. The returning
+  //. array is truncated to the length of the shorter of the two
+  //. input arrays.
```
davidchambers: This could be shortened a little. I suggest:

_Returns the result of combining, pairwise, the given arrays using the given binary function. The length of the resulting array is equal to the length of the shorter input array._

test/zipWith.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+var S = require('..');
+
+var eq = require('./internal/eq');
+
+test('zipWith', function() {
```
davidchambers: Please add a second empty line after the imports.

test/zipWith.js
```diff
@@ -0,0 +1,18 @@
...
+
+  eq(typeof S.zipWith, 'function');
+  eq(S.zipWith.length, 3);
+  eq(S.zipWith.toString(), 'zipWith :: (a -> b -> c) -> Array a -> Array b -> Array c');
+
+  eq(S.zipWith(S.add, [1, 2, 3], [3, 2, 1]), [4, 4, 4]);
+  eq(S.zipWith(S.add, [1, 2, 3, 4], [3, 2, 1]), [4, 4, 4]);
+  eq(S.zipWith(S.add, [1, 2, 3], [3, 2, 1, 0]), [4, 4, 4]);
+  eq(S.zipWith(S.on(S.concat, S.of(Array)), [1, 3, 5], [2, 4, 6]), [[1, 2], [3, 4], [5, 6]]);
```
davidchambers: I'd prefer not to rely on so many other Sanctuary functions here.

test/zipWith.js
```diff
@@ -0,0 +1,18 @@
...
+
+test('zipWith', function() {
+
+  eq(typeof S.zipWith, 'function');
+  eq(S.zipWith.length, 3);
+  eq(S.zipWith.toString(), 'zipWith :: (a -> b -> c) -> Array a -> Array b -> Array c');
+
+  eq(S.zipWith(S.add, [1, 2, 3], [3, 2, 1]), [4, 4, 4]);
+  eq(S.zipWith(S.add, [1, 2, 3, 4], [3, 2, 1]), [4, 4, 4]);
+  eq(S.zipWith(S.add, [1, 2, 3], [3, 2, 1, 0]), [4, 4, 4]);
```
davidchambers: The problem with using `S.add` is that addition is commutative. Using a noncommutative combining function would make the tests more valuable, as it would ensure that the function is applied to its arguments in the right order. In fact, it would be ideal to use a combining function which takes arguments of different types.

index.js
sadasant: `x => y => [x,y]` was causing browser issues with the build, seems like we stopped supporting some browsers at some point, but we didn't clean up the karma settings.

sadasant: If it breaks again after I push this change, you'll need to help me ლ(ಠ益ಠლ)

davidchambers: No problem. :)

test/zipWith.js
sadasant: Now I an example with S.gte! But don’t fear to tell me if you’d rather have examples with other specific functions 🤔 my point of reference is definitely not helping my creativity.

sadasant:
@davidchambers The arrow functions seems to have unleashed havoc 🤔

Other than that I’m ready for a next round of feedback \o/ then zip then whatever else is on the table, let’s go! 😃

test/zipWith.js
```diff
@@ -0,0 +1,19 @@
...
+test('zipWith', function() {
+
+  eq(typeof S.zipWith, 'function');
+  eq(S.zipWith.length, 3);
+  eq(S.zipWith.toString(), 'zipWith :: (a -> b -> c) -> Array a -> Array b -> Array c');
+
+  eq(S.zipWith(S.add, [1, 2, 3, 4], [3, 2, 1]), [4, 4, 4]);
+  eq(S.zipWith(S.add, [1, 2, 3], [3, 2, 1, 0]), [4, 4, 4]);
+  eq(S.zipWith(S.gte, [1, 2, 3], [3, 2, 1]), [true, true, false]);
+  eq(S.zipWith(a => b => [a, b], [1, 3, 5], [2, 4, 6]), [[1, 2], [3, 4], [5, 6]]);
```
davidchambers: You may have noticed that the source code and test suite do not use any ES6 syntax. This is intentional. You'll need to use `function(a) { ... }` here. ES6 syntax can, though, be used in doctests because we do not run doctests in browsers.

index.js
```diff
@@ -3656,6 +3656,40 @@
...
+  //.
+  //. ```javascript
+  //. > S.zipWith(S.add, [1, 2, 3, 4], [3, 2, 1])
+  //. [4, 4, 4]
+  //.
+  //. > S.zipWith(S.add, [1, 2, 3], [3, 2, 1, 0])
+  //. [4, 4, 4]
+  //.
+  //. > S.zipWith(S.gte, [1, 2, 3], [3, 2, 1])
+  //. [true, true, false]
```
davidchambers: I find this example confusing because the parameter order of `S.gte` is surprising when used this way. Let's replace all four doctests with the following two:

```javascript
> S.zipWith(s => t => s + t, ['a', 'b'], ['x', 'y', 'z'])
['ax', 'by']

> S.zipWith(s => t => s + t, ['a', 'b', 'c'], ['x', 'y'])
['ax', 'by']
```

Keep in mind that doctests and unit tests serve different roles: doctests show readers how to use a function; unit tests ensure that every code path is correct. As a result, there need not be a one-to-one correlation between a function's doctests and unit tests.

index.js
```diff
@@ -3656,6 +3656,40 @@
...
       Z.sortBy);
 
+  //# zipWith :: (a -> b -> c) -> Array a -> Array b -> Array c
+  //.
+  //. Creates a new array out of the two supplied arrays
+  //. by applying the given function to each equally-positioned
+  //. pair of elements from each given array. Returns the result
+  //. of combining, pairwise, the given arrays using the given
+  //. binary function. The length of the resulting array is equal
+  //. to the length of the shorter input array.
```
davidchambers: The first and second sentences are duplicative. This may be sufficient:

```javascript
  //. Returns the result of combining, pairwise, the given arrays using the
  //. given binary function. The length of the resulting array is equal to the
  //. length of the shorter input array.
```

test/zipWith.js
sadasant: Thanks! I'll do that

index.js
sadasant: Way better! Thank you!

index.js
sadasant: Gotcha!

index.js
sadasant: `s => t => s + t` is just S.concat, right? would you rather have S.concat or the longer one? We also have an example with `a => b > [a, b]`.

davidchambers: The problem with `S.concat` is that it currently accepts its arguments one at a time or both at once. Were we to use `S.concat` here, it wouldn't be clear to readers that the function must be of type `a -> b -> c` rather than `(a, b) -> c`.

sadasant: gotcha

index.js
```diff
@@ -3656,6 +3656,37 @@
...
+  //. Returns the result of combining, pairwise, the given arrays using the
+  //. given binary function. The length of the resulting array is equal to the
+  //. length of the shorter input array.
+  //.
+  //. ```javascript
+  //. > S.zipWith(S.add, [1, 2, 3, 4], [3, 2, 1])
+  //. [4, 4, 4]
+  //.
+  //. > S.zipWith(S.add, [1, 2, 3], [3, 2, 1, 0])
+  //. [4, 4, 4]
```
davidchambers: Let's drop these two examples.

test/zipWith.js
```diff
@@ -0,0 +1,31 @@
...
+  eq(S.zipWith.toString(), 'zipWith :: (a -> b -> c) -> Array a -> Array b -> Array c');
+
+  eq(S.zipWith(S.add, [1, 2, 3, 4], [3, 2, 1]), [4, 4, 4]);
+  eq(S.zipWith(S.add, [1, 2, 3], [3, 2, 1, 0]), [4, 4, 4]);
+
+  function concat(a) {
+    return function(b) {
+      return a + b;
+    };
+  }
```
davidchambers: Let's reuse [`concat`][1] here.


[1]: https://github.com/sanctuary-js/sanctuary/blob/master/test/internal/concat.js

test/zipWith.js
```diff
@@ -0,0 +1,31 @@
...
+  eq(S.zipWith(concat, ['a', 'b'], ['x', 'y', 'z']), ['ax', 'by']);
+
+  function pair(a) {
+    return function(b) {
+      return [a, b];
+    };
+  }
```
davidchambers: Let's move this definition to __test/internal/pair.js__.

sadasant: Neat!!

index.js
```diff
@@ -3656,6 +3656,31 @@
...
+  //. Returns the result of combining, pairwise, the given arrays using the
+  //. given binary function. The length of the resulting array is equal to the
+  //. length of the shorter input array.
+  //.
+  //. ```javascript
+  //. > S.zipWith(a => b => a + b, ['a', 'b'], ['x', 'y', 'z'])
+  //. ['ax', 'by']
+  //.
+  //. > S.zipWith(a => b => [a, b], [1, 3, 5], [2, 4, 6])
+  //. [[1, 2], [3, 4], [5, 6]]
```
davidchambers: Let's make the second array shorter than the first array in this example.

davidchambers:
We could define `S.zip` in this pull request as well. What do you think, @sadasant?

sadasant:
@davidchambers yes! On board :)

index.js
```diff
@@ -3656,6 +3663,50 @@
...
+  //. [['a', 'x'], ['b', 'y']]
+  //.
+  //. > S.zip([1, 3, 5], [2, 4])
+  //. [[1, 2], [3, 4]]
+  //. ```
+  S.zip =
+  def('zip',
+      {},
+      [$.Array(a), $.Array(b), $.Array($.Pair(a, b))],
+      S.zipWith(pair));
```
davidchambers: Let's avoid using `S` functions internally. You'll need to define `zip` in terms of `zipWith`.

index.js
```diff
@@ -225,6 +225,13 @@
     };
   }
 
+  //  pair :: a -> b -> Pair a b
+  function pair(a) {
+    return function(b) {
+      return [a, b];
```
davidchambers: I would name the parameters `x` and `y` so they don't match the names of the type variables, but I don't feel strongly about this.

index.js
```diff
@@ -3586,6 +3593,53 @@
...
+  def('zipWith',
+      {},
+      [Fn(a, Fn(b, c)), $.Array(a), $.Array(b), $.Array(c)],
+      zipWith);
+
+  //# zip :: Array a -> Array b -> Array (Pair a b)
+  //.
+  //. Returns the result of combining, pairwise, the given arrays into a
+  //. resulting array with length equal to the length of the shorter input
+  //. array.
```

_Returns an array of pairs of corresponding elements from the given arrays. The length of the resulting array is equal to the length of the shorter input array._

Please add “see also” links in both directions.

I wonder whether we should list `zip` before `zipWith`. What do you think?

test/internal/pair.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+//  pair :: a -> b -> Pair a b
+module.exports = function pair(a) {
+  return function(b) {
+    return [a, b];
```
davidchambers: Let's use `x` and `y` here too.

index.js
sadasant: doin thy changes!

davidchambers:
This pull request is looking great, Daniel!

Before we merge the pull request I would like you to squash all the commits into a single commit which sits atop the most recent commit on `master`. While you are at it, please update the commit message to `array: add S.zip and S.zipWith`, and update the title of the pull request to match.

Issue #506, opened 2018-03-13
Reasonable defaults for env


blackxored:
Correct me if this is the case already, but `env` configuration should be something only needed if we're indeed customizing. As it stands, disabling type checking in production or any other common uses along those lines should be default. 
Libraries like React do this automatically and I believe Sanctuary should too. 
Again, apologies if it does this already, in which case the documentation is unclear.

davidchambers:
Are you suggesting that `require ('sanctuary')` should produce a Sanctuary module which does not perform type checking?

blackxored:
It should figure out whether is running in production and disable it, unless explicitly opted-in by the user creating its own env. React, Webpack and others has such sane defaults.

gabejohnson:
I'd rather have type checking on by default and have a production build of `sanctuary-def` with the type checking code removed (to save file size).

davidchambers:
> It should figure out whether is running in production and disable it, unless explicitly opted-in by the user creating its own env. React, Webpack and others has such sane defaults.

I don't consider that to be a default, since the value is context-dependent. I consider that to be magic. ;)

I remain happy with our response to sanctuary-js/sanctuary-def#6. I'm open to changing the default value of the `checkTypes` option from `true` to `false`, if there are compelling reasons to do so.

blackxored:
You might not realize these as authors, contributors or people who had been using this for a while, but I ought to say the barrier of entry to this library is really high, and not because of FP but because these kind of decisions and obvious omissions.

In the JS ecosystem (and others, but can't speak for all let's keep it simple) I would expect a library, any library, to give me extra warnings on development, and remove those if I'm running on production *by default*. *NODE_ENV* exists for a reason, what's really the argument against using that? A shared var? It's meant to be, babel uses it, your app uses it, webpack uses it.

Does it couple it to the browser, no it doesn't, most people are bundling and you can safe-check on process. I'm not really getting it. 

If I want type checking in production then I would go and create my own Sanctuary `env`.
It's how Webpack works, React, flow-runtime even, the latter is essentially type checking BTW. 

Am I really tripping here? I don't see the magic. Type check on on development by .default, off on production by default. Or provide an entry with it disabled I can at least alias. 

I've enough stuff to configure and I'm sure you do as well, reason (not ml) please. 😂

davidchambers:
> the barrier of entry to this library is really high

[...]

PR #512, opened 2018-03-21
disable type checking for default module in "production"

davidchambers:
In #506, @blackxored has made a case for changing the default value of `checkTypes` from `true` to an expression involving `process.env.NODE_ENV`. I have long opposed this change, but when Aldwin disagrees with me it's wise to reconsider my position. I have done so, and I have changed my mind.

Requirements:

  - default Sanctuary module should perform type checking *by default*;
  - behaviour must be easy to understand; and
  - the user must be permitted to specify whether type checking should be performed.

Although not a requirement, it would be nice to reduce the proportion of users who find it necessary to use [`S.create`][1]. I believe this is your primary goal, Adrian.

The current rule is as follows:

> Type checking is enabled for the default module.

This satisfies the requirements but forces many users to use `S.create` even though they are happy with the default environment.

The new rule is as follows:

> Type checking is enabled for the default module _unless `process.env.NODE_ENV` is `'production'`_.

This satisfies the requirements and obviates the need for `S.create` in many cases. :)

[1]: https://sanctuary.js.org/#create

### DIFF
--- a/index.js
+++ b/index.js
@@ -172,19 +172,16 @@
 //. // => 3
 //. ```
 //.
-//. There is a performance cost to run-time type checking. One may wish to
-//. disable type checking in certain contexts to avoid paying this cost.
-//. [`create`](#create) facilitates the creation of a Sanctuary module which
-//. does not perform type checking.
-//.
-//. In Node, one could use an environment variable to determine whether to
-//. perform type checking:
+//. There is a performance cost to run-time type checking. Type checking is
+//. disabled by default if `process.env.NODE_ENV` is `'production'`. If this
+//. rule is unsuitable for a given program, one may use [`create`](#create)
+//. to create a Sanctuary module based on a different rule. For example:
 //.
 //. ```javascript
-//. const {create, env} = require ('sanctuary');
-//.
-//. const checkTypes = process.env.NODE_ENV !== 'production';
-//. const S = create ({checkTypes, env});
+//. const S = sanctuary.create ({
+//.   checkTypes: localStorage.getItem ('SANCTUARY_CHECK_TYPES') === 'true',
+//.   env: sanctuary.env,
+//. });
 //. ```
 //.
 //. ## API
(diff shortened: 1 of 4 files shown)


blackxored:
This looks great, thanks a lot for considering this.

index.js
```diff
@@ -174,17 +174,17 @@
...
 //. There is a performance cost to run-time type checking. One may wish to
 //. disable type checking in certain contexts to avoid paying this cost.
-//. [`create`](#create) facilitates the creation of a Sanctuary module which
-//. does not perform type checking.
+//. Type checking is disabled by default if `process.env.NODE_ENV` is
+//. `'production'`.
 //.
-//. In Node, one could use an environment variable to determine whether to
-//. perform type checking:
+//. [`create`](#create) facilitates the creation of a Sanctuary module which
```
blackxored: Maybe it's just me (may be contextual) but I believe it implies you need to use `create` to enable type-checking and it is enabled by default on development and test environments. Correct?

davidchambers: Good point. The sentence can be interpreted in two ways. I'll rework it to remove the ambiguity.

test/NODE_ENV.js
```diff
@@ -0,0 +1,105 @@
...
+var path = require('path');
+var vm = require('vm');
+
+var eq = require('./internal/eq');
+var throws = require('./internal/throws');
+
+
+suite('NODE_ENV', function() {
+
+  var source = fs.readFileSync(path.join(__dirname, '..', 'index.js'), 'utf8');
```
Avaq: It looks like this breaks browser tests.

davidchambers: Indeed. I excluded the file in this context, which seems reasonable.

index.js
davidchambers: I have updated this section of the documentation. Let me know what you think of the changes, @blackxored.

davidchambers:
> Note that this could be a breaking change to some (those who have come to rely on type checking in production, if any).

This is well worth mentioning in the release notes.

PR #467, opened 2017-12-14
lint: detect incorrect indentation within createSanctuary

davidchambers:
This is an alternative to #457. It adds more code, but may prove easier to maintain. I'm interested in your thoughts, @futpib.

PR #514, opened 2018-03-24
share definitions of constraints, types, and functions

davidchambers:
Currently, each Sanctuary function is defined once for each Sanctuary module created. This pull request restructures the code so that constraints, types, and functions are defined just once.

### DIFF
--- a/index.js
+++ b/index.js
@@ -213,6 +213,25 @@
 
   'use strict';
 
+  /* global __doctest:false */
+  /* istanbul ignore if */
+  if (typeof __doctest !== 'undefined') {
+    /* eslint-disable no-unused-vars */
+    var Nil = __doctest.require('./test/internal/List').Nil;
+    var Cons = __doctest.require('./test/internal/List').Cons;
+    var Sum = __doctest.require('./test/internal/Sum');
+    var S = (function(S) {
+      return S.create({
+        checkTypes: true,
+        env: S.env.concat([
+          __doctest.require('./test/internal/List').Type($.Unknown),
+          Sum.Type
+        ])
+      });
+    }(require('.')));
+    /* eslint-enable no-unused-vars */
+  }
+
   //  Fn :: (Type, Type) -> Type
   function Fn(x, y) { return $.Function([x, y]); }
 
@@ -336,18 +355,7 @@
   //  Options :: Type
   var Options = $.RecordType({checkTypes: $.Boolean, env: $.Array($.Any)});
 
-  //  createSanctuary :: Options -> Module
-  function createSanctuary(opts) {
-
-  /* eslint-disable indent */
-
-  //  checkTypes :: Boolean
-  var checkTypes = opts.checkTypes;
-
-  //  env :: Array Type
-  var env = opts.env;
-
-  var S = {};
+  var _ = {};
 
   //# create :: { checkTypes :: Boolean, env :: Array Type } -> Module
   //.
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -213,6 +213,25 @@
...
+    var Sum = __doctest.require('./test/internal/Sum');
+    var S = (function(S) {
+      return S.create({
+        checkTypes: true,
+        env: S.env.concat([
+          __doctest.require('./test/internal/List').Type($.Unknown),
+          Sum.Type
+        ])
+      });
+    }(require('.')));
```
davidchambers: We define `S` here for the benefit of the doctests.

index.js
```diff
@@ -329,18 +348,7 @@
   //  Options :: Type
   var Options = $.RecordType({checkTypes: $.Boolean, env: $.Array($.Any)});
 
-  //  createSanctuary :: Options -> Module
-  function createSanctuary(opts) {
-
-  /* eslint-disable indent */
```
davidchambers: This pull request solves the problem discussed in #457 and #467.

/cc @futpib

index.js
```diff
@@ -392,31 +400,31 @@
...
   //.
   //. See also [`env`](#env).
-  S.create =
-  $.create({checkTypes: checkTypes, env: defaultEnv})('create',
-                                                      {},
-                                                      [Options, $.Object],
-                                                      createSanctuary);
+  function create(opts) {
+    var S = {
+      env: defaultEnv,
```
davidchambers: This should be `opts.env` rather than `defaultEnv`, should it not? If others agree, I'll open a separate pull request for this change.

index.js
```diff
@@ -392,31 +400,31 @@
...
+      Nothing: Nothing,
+      EitherType: $Either,
+      Either: Either
+    };
+    var def = $.create(opts);
+    Object.keys(_).forEach(function(name) {
+      S[name] = def(name, _[name].consts, _[name].types, _[name].impl);
+    });
+    return S;
+  }
```
davidchambers: This function is several thousand lines smaller than `createSanctuary`. ;)

gabejohnson:
To make certain I understand, you pulled the definitions out of `createSanctuary` so they're defined on `_` which itself is defined in the closure and referenced in `create`. Is this correct?

davidchambers:
Before:

```javascript
function createSanctuary(opts) {
  var S = {};
  //  Sanctuary functions (including ‘create’) defined here.
  return S;
}

return createSanctuary({checkTypes: true, env: defaultEnv});
```

`require('sanctuary').create(...)` would result in each exported function being defined *twice*.

After:

```javascript
//  Sanctuary functions (including ‘create’) defined at the top level.

return create({checkTypes: true, env: defaultEnv});
```

`require('sanctuary').create(...)` would now result in each exported function being defined just once, since `create` now reuses the functions defined at the top level (stored in the `_` string map).

index.js
```diff
@@ -1318,21 +1434,20 @@
...
   //# curry4 :: ((a, b, c, d) -> e) -> a -> b -> c -> d -> e
   //.
   //. Curries the given quaternary function.
   //.
   //. ```javascript
-  //. > global.createRect = S.curry4((x, y, width, height) =>
+  //. > const createRect = S.curry4((x, y, width, height) =>
   //. .   ({x, y, width, height})
   //. . )
-  //. createRect
```
davidchambers: :tada:

index.js
davidchambers: #524

PR #518, opened 2018-04-04
reinstate S.unchecked

### LINKED ISSUE #517: Request: Add `sanctuary/unchecked` directory
https://github.com/mwalkerwells/sanctuary-dynamic/commit/b9146870197eb543e4bc39102bfb73657571b7c3#commitcomment-28404523

davidchambers:
Closes #517

`S.unchecked` was removed in #206. Two years later, it's time to bring it back. :)

/cc @mwalkerwells


index.js
```diff
@@ -402,7 +406,7 @@
   function create(opts) {
     var def = $.create(opts);
     var S = {
-      env: defaultEnv,
+      env: opts.env,
```
davidchambers: I'll open a separate pull request for this change.

davidchambers:
Would this work, @mwalkerwells?

```javascript
import { unchecked as S_ } from 'sanctuary';
```

mwalkerwells:
Oh duh, yeah that works. 😬

davidchambers:
That's good to hear, @mwalkerwells! Let's stick with the `unchecked` property in that case because it has the advantage of also working in the browser.

tomkel:
See [here](https://github.com/rollup/rollup/pull/2023#issuecomment-370125474) for a little discussion on the feature/mistake of node.js automatically resolving `index.js`.

index.js
davidchambers: #524

PR #519, opened 2018-04-04
documentation: add doctest for S.env

davidchambers:


davidchambers:
Indeed. The input will become simply `S.env` once `S.equals` can operate on `Type` values. :)

PR #520, opened 2018-04-04
abandon Ramda-style currying in favour of simple currying

### LINKED ISSUE #438: readable code without Ramda-style currying

davidchambers:
Closes #438

<https://github.com/sanctuary-js/sanctuary-def/releases/tag/v0.15.0>

### DIFF
--- a/index.js
+++ b/index.js
@@ -151,7 +151,7 @@
 //. silent failures due to type coercion (at worst). For example:
 //.
 //. ```javascript
-//. S.add(2, true);
+//. S.add (2, true);
 //. // ! TypeError: Invalid value
 //. //
 //. //   add :: FiniteNumber -> FiniteNumber -> FiniteNumber
@@ -168,7 +168,7 @@
 //. Compare this to the behaviour of Ramda's unchecked equivalent:
 //.
 //. ```javascript
-//. R.add(2, true);
+//. R.add (2, true);
 //. // => 3
 //. ```
 //.
@@ -181,10 +181,10 @@
 //. perform type checking:
 //.
 //. ```javascript
-//. const {create, env} = require('sanctuary');
+//. const {create, env} = require ('sanctuary');
 //.
 //. const checkTypes = process.env.NODE_ENV !== 'production';
-//. const S = create({checkTypes, env});
+//. const S = create ({checkTypes, env});
 //. ```
 //.
 //. ## API
@@ -195,21 +195,21 @@
 
   /* istanbul ignore else */
   if (typeof module === 'object' && typeof module.exports === 'object') {
-    module.exports = f(require('sanctuary-def'),
-                       require('sanctuary-type-classes'),
-                       require('sanctuary-type-identifiers'));
+    module.exports = f (require ('sanctuary-def'),
+                        require ('sanctuary-type-classes'),
+                        require ('sanctuary-type-identifiers'));
   } else if (typeof define === 'function' && define.amd != null) {
-    define(['sanctuary-def',
-            'sanctuary-type-classes',
-            'sanctuary-type-identifiers'],
-           f);
+    define (['sanctuary-def',
+             'sanctuary-type-classes',
+             'sanctuary-type-identifiers'],
+            f);
   } else {
-    self.sanctuary = f(self.sanctuaryDef,
-                       self.sanctuaryTypeClasses,
-                       self.sanctuaryTypeIdentifiers);
+    self.sanctuary = f (self.sanctuaryDef,
+                        self.sanctuaryTypeClasses,
+                        self.sanctuaryTypeIdentifiers);
   }
 
-}(function($, Z, type) {
+} (function($, Z, type) {
 
   'use strict';
 
@@ -217,23 +217,42 @@
   /* istanbul ignore if */
   if (typeof __doctest !== 'undefined') {
     /* eslint-disable no-unused-vars */
-    var Nil = __doctest.require('./test/internal/List').Nil;
-    var Cons = __doctest.require('./test/internal/List').Cons;
-    var Sum = __doctest.require('./test/internal/Sum');
+    var Nil = (__doctest.require ('./test/internal/List')).Nil;
+    var Cons = (__doctest.require ('./test/internal/List')).Cons;
+    var Sum = __doctest.require ('./test/internal/Sum');
     var S = (function(S) {
-      return S.create({
+      return S.create ({
         checkTypes: true,
-        env: S.env.concat([
-          __doctest.require('./test/internal/List').Type($.Unknown),
+        env: S.env.concat ([
+          (__doctest.require ('./test/internal/List')).Type ($.Unknown),
           Sum.Type
         ])
       });
-    }(require('.')));
+    } (require ('.')));
     /* eslint-enable no-unused-vars */
   }
 
-  //  Fn :: (Type, Type) -> Type
-  function Fn(x, y) { return $.Function([x, y]); }
+  //  B :: (b -> c) -> (a -> b) -> a -> c
+  function B(f) {
+    return function(g) {
+      return function(x) {
+        return f (g (x));
+      };
+    };
+  }
+
+  //  Fn :: Type -> Type -> Type
+  function Fn(x) {
+    return function(y) {
+      return $.Function ([x, y]);
+    };
+  }
+
+  //  get_ :: String -> a -> Maybe b
+  function get_(key) {
+    return B (function(obj) { return key in obj ? Just (obj[key]) : Nothing; })
+             (toObject);
+  }
 
   //  pair :: a -> b -> Pair a b
   function pair(x) {
@@ -244,20 +263,13 @@
 
   //  toObject :: a -> Object
   function toObject(x) {
-    return x == null ? Object.create(null) : Object(x);
+    return x == null ? Object.create (null) : Object (x);
   }
 
   //  typeEq :: String -> a -> Boolean
   function typeEq(typeIdent) {
     return function(x) {
-      return type(x) === typeIdent;
-    };
-  }
-
-  //  uncurry2 :: (a -> b -> c) -> ((a, b) -> c)
-  function uncurry2(f) {
-    return function(x, y) {
-      return f(x)(y);
+      return type (x) === typeIdent;
     };
   }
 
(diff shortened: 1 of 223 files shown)


Avaq:
This is a pretty daunting diff.

Most seem like stylistic changes of sorts, eg `a, b` -> `a) (b`. Could you point me to the significant changes, if any? Or otherwise if you'd really like me to take a good look at it, we could arrange a call of sorts so you can guide me through this.

scripts/lint
```diff
@@ -0,0 +1,39 @@
...
+  done
+}
+
+trap 'mv README.md.orig README.md && rm -f README.md.temp' EXIT
+cp README.md README.md.orig
+VERSION=0.0.0 PREVIOUS_VERSION=0.0.0 node_modules/.bin/sanctuary-generate-readme
+cp README.md README.md.temp
+rewrite <README.md.temp >README.md
+
+node_modules/.bin/eslint --config eslint/es6.js -- README.md eslint
```
davidchambers: This script enables `npm run lint` to run `eslint` on files that would not otherwise be linted.

Linting the readme involves a number of steps:

1.  Save a copy of the file as it exists.
2.  Regenerate the readme (the `VERSION` and `PREVIOUS_VERSION` values are unimportant).
3.  Use `rewrite` to remove leading `>` and `.` characters, parenthesize object literals which would otherwise be parsed as blocks, and insert a semicolon at the end of each input and output line.
4.  Run `eslint`.
5.  Restore the readme to its original state (via an EXIT trap).

test/internal/eq.js
```diff
@@ -1,13 +1,16 @@
...
-  assert.strictEqual(equals(actual)(expected), true);
+//  eq :: a -> b -> Undefined !
+module.exports = function eq(actual) {
+  assert.strictEqual (arguments.length, eq.length);
+  return function eq$1(expected) {
+    assert.strictEqual (arguments.length, eq$1.length);
+    assert.strictEqual (toString (actual), toString (expected));
+    assert.strictEqual (equals (actual) (expected), true);
+  };
 };
```
davidchambers: The internal `eq` function is now curried.

test/internal/throws.js
```diff
@@ -1,11 +1,14 @@
...
+var equals = require ('./equals');
+
+//  throws :: (() -> Undefined !) -> Error -> Undefined !
+module.exports = function throws(thunk) {
+  assert.strictEqual (arguments.length, throws.length);
+  return function throws$1(expected) {
+    assert.strictEqual (arguments.length, throws$1.length);
+    assert.throws (thunk, equals (expected));
+  };
 };
```
davidchambers: The internal `throws` function is now curried, and now takes a single value representing the expected error rather than a type representative and an error message.

Before:

```javascript
throws(thunk, TypeError, 'Invalid value');
```

After:

```javascript
throws (thunk) (new TypeError ('InvalidValue'));
```

test/internal/List.js
```diff
@@ -25,89 +24,91 @@ function _List(tag, head, tail) {
...
+var Nil = List.Nil = new _List ('Nil');
+
+//  Cons :: a -> List a -> List a
+var Cons = List.Cons = function Cons(head) {
+  eq (arguments.length) (Cons.length);
+  return function Cons$1(tail) {
+    eq (arguments.length) (Cons$1.length);
+    return new _List ('Cons', head, tail);
+  };
 };
```
davidchambers: `Cons` is now curried.

index.js
```diff
@@ -195,45 +195,52 @@
...
   }
 
-  //  Fn :: (Type, Type) -> Type
-  function Fn(x, y) { return $.Function([x, y]); }
+  //  Fn :: Type -> Type -> Type
+  function Fn(x) {
+    return function(y) {
+      return $.Function ([x, y]);
+    };
+  }
```
davidchambers: `Fn` is now curried.

index.js
```diff
@@ -274,78 +274,75 @@
...
-    typeEq(eitherTypeIdent),
-    function(either) { return either.isLeft ? [either.value] : []; },
-    function(either) { return either.isRight ? [either.value] : []; }
-  );
+  var $Either = $.BinaryType
+    (eitherTypeIdent)
+    (readmeUrl ('EitherType'))
+    (typeEq (eitherTypeIdent))
+    (either (of (Array)) (K ([])))
+    (either (K ([])) (of (Array)));
```
davidchambers: Now that functions such as `K` are curried (and hoisted) we're able to use them throughout the file. :)

index.js
```diff
@@ -505,103 +502,101 @@
...
-    impl: Z.equals
+    impl: curry2 (Z.equals)
   };
 
-  //# lt :: Ord a => a -> (a -> Boolean)
+  //# lt :: Ord a => a -> a -> Boolean
   //.
   //. Returns `true` [iff][] the *second* argument is less than the first
-  //. according to [`Z.lt`][]. The arguments must be provided one at a time.
+  //. according to [`Z.lt`][].
```
davidchambers: This is one of the important changes listed in <https://github.com/sanctuary-js/sanctuary/issues/438#issuecomment-371809619>. Now that all functions are “simply” curried there's no reason for this function's type to be `[a, $.Predicate (a)]` rather than `[a, a, $.Boolean]`. :)

index.js
```diff
@@ -734,25 +729,30 @@
...
-  //. > S.filter(S.odd, S.Just(1))
-  //. Just(1)
+  //. > S.filter (S.odd) (S.Just (1))
+  //. Just (1)
   //. ```
+  function filter(pred) {
+    return function(filterable) {
+      return Z.filter (pred, filterable);
+    };
+  }
```
davidchambers: We are defining `filter` rather than setting `impl` to `curry2 (Z.filter)` in order to partially apply the function in other definitions.

index.js
```diff
@@ -1241,21 +1257,28 @@
...
-    }
-    return Z.chainRec(typeRep, step, x);
+  function chainRec(typeRep) {
+    return function(f) {
+      return function(x) {
+        return Z.chainRec (typeRep, step, x);
+      };
+      function step(next, done, x) {
+        return Z.map (either (next) (done), f (x));
+      }
```
davidchambers: I decided to rely on function hoisting here to avoid interrupting the `function > return > > return` staircase.

index.js
```diff
@@ -1461,26 +1497,33 @@
...
   //. ```
-  function curry5(f, v, w, x, y, z) {
-    return f(v, w, x, y, z);
+  function curry5(f) {
+    return function(v) {
+      return function(w) {
+        return function(x) {
+          return function(y) {
+            return function(z) {
+              return f (v, w, x, y, z);
```
davidchambers: For some reason this makes me happy. :D

index.js
```diff
@@ -1711,7 +1773,7 @@
   //. > S.Nothing.isNothing
   //. true
   //.
-  //. > S.Just(42).isNothing
+  //. > (S.Just (42)).isNothing
```
davidchambers: This is a situation in which unnecessary parens are justified: `isNothing` in `S.Just (42).isNothing` is tightly bound, visually, to `42`.

Fortunately, such property access is rare in code written with Sanctuary (one could use `S.isNothing` in this case).

index.js
```diff
@@ -3122,19 +3226,16 @@
...
   //. ```
-  function complement(pred, x) {
-    return !pred(x);
-  }
   _.complement = {
     consts: {},
-    types: [$.Predicate(a), a, $.Boolean],
-    impl: complement
+    types: [$.Predicate (a), a, $.Boolean],
+    impl: compose (not)
```
davidchambers: Curried `compose` is very useful!

index.js
```diff
@@ -3650,26 +3779,30 @@
...
   }
   _.find = {
     consts: {f: [Z.Foldable]},
-    types: [$.Predicate(a), f(a), $Maybe(a)],
+    types: [$.Predicate (a), f (a), $Maybe (a)],
     impl: find
   };
 
-  //# unfoldr :: (b -> Maybe (Pair a b)) -> b -> Array a
+  //# unfoldr :: (b -> Maybe (Array2 a b)) -> b -> Array a
```
davidchambers: We're now using [`$.Array2`][1], which replaced `$.Pair`.


[1]: https://github.com/sanctuary-js/sanctuary-def#Array2

index.js
Avaq: I was typing a similar comment to `curry`:

> So blissfully transparent! :D

...but then I scrolled down and saw `curry2`, which made me happier, so I removed the comment to move it over there. But then I scrolled down further to find `curry3`. Woah, how cool is this going to get? I scrolled all the way to `curry5` and found your comment. :)

index.js
```diff
@@ -2241,19 +2308,16 @@
...
   //. ```
-  function mapMaybe(f, filterable) {
-    return justs(Z.map(f, filterable));
-  }
   _.mapMaybe = {
     consts: {f: [Z.Filterable, Z.Functor]},
-    types: [Fn(a, $Maybe(b)), f(a), f(b)],
-    impl: mapMaybe
+    types: [Fn (a) ($Maybe (b)), f (a), f (b)],
+    impl: blackbird (justs) (map)
```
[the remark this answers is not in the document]
Avaq: I would think so. Especially if used sparsely and considering that the definition of blackbird itself is also somewhat mind-bending.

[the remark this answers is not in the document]
Avaq: Now that I've figured out what's going on, I also see the appeal though ;)

index.js
Avaq: I have been saying the same thing! This is one of the reasons why Ramda's compose is so crippled, but you don't really realise until you try it.

index.js
davidchambers: :rofl:

index.js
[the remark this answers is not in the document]
davidchambers: I've removed `blackbird`. The implementation is now `B (B (justs)) (map)`.

The implementations of `encase2` and `encase3` sell me on the idea:

```javascript
B (B (B (eitherToMaybe))) (encaseEither2 (I))
B (B (B (B (eitherToMaybe)))) (encaseEither3 (I))
```

Beautiful!

On several occasions I've defended [`map (map (map (map (nbsp))))`][1]. I feel a similar way about `B (B (B (B (eitherToMaybe))))`. It looks strange at first, but once one learns how to interpret the nesting it's much clearer than the equivalent nested lambdas.


[1]: https://github.com/plaid/transcribe/blob/v1.0.0/bin/transcribe#L33

scripts/lint
davidchambers: sanctuary-js/sanctuary-scripts#13

dakom:
> I would be interested to find out how these changes affect performance. I would imagine JavaScript function context creation time has increased quite a lot.

That would be fascinating to see benchmarks, ideally of actual projects (not just adding a bunch of numbers)

Issue #479, opened 2018-01-23
Module scoping 


mfidemraizer:
Hi!

As @davidchambers and I have been discussing, there's an interesting feature that would allow us to create Sanctuary module scopes.

What's the purpose of this?

- We would be able to register new types for a given section of code or module instead of being forced to declare them just for the entire project.
- Scopes would allow us to disable type checking for a given section of code or module too.

Pseudo-code:
```javascript
const 
         Sa = S.create(...),
         s = Sa.scope({ checkTypes: <true|false>, env: [...] })
```

That is, *child Sanctuary scope* would inherit parent's `env` while it would be able to override `checkTypes`.

I find this convenient in order to let us define new types as you need them instead of having to load them during program startup.

davidchambers:
Perhaps we don't need a dedicated function for this: IIFEs work well.

In this example we disable type checking for a block of code:

```javascript
(S => S.add ('abc') (123)) (S.create ({checkTypes: false, env: []}));
// => 'abc123'
```

In this example we use a stricter environment which does not permit the use of `NaN`:

```javascript
(S => {
  S.I (Infinity);
  // => Infinity
  S.I (NaN);
  // ! TypeError: Unrecognized value
  //
  //   I :: a -> a
  //        ^
  //        1
  //
  //   1)  NaN :: (no types)
  //
  //   The value at position 1 is not a member of any type in the environment.
  //
  //   The environment contains the following types:
  //
  //     - ValidNumber
}) (S.create ({checkTypes: true, env: [$.ValidNumber]}));
```

Similarly, we could provide an environment which contains some of our own types.

mfidemraizer:
@davidchambers And what about a new param `inherit` which means that omitted settings are inherited if they're already configured? 

Or with `S` you're referring to an `S` instance already gotten from calling `S.create`?

davidchambers:
> And what about a new param `inherit` which means that omitted settings are inherited if they're already configured?

Something like this, do you mean?

```javascript
S.create ({inherit: true, env: [$.ValidNumber]})
```

In the example above `checkTypes` would be set to whichever value was used when `S` was defined. This feels very JavaScripty, and would not type check. ;)

We could do this instead:

```javascript
S.create ({checkTypes: S.Nothing, env: S.Just ([$.ValidNumber])})
```

Simpler, though, would be to expose `S.checkTypes`. One could then write:

```javascript
S.create ({checkTypes: S.checkTypes, env: [$.ValidNumber]})
[...]

PR #524, opened 2018-04-08
update S.env to reference the module's environment

davidchambers:
`S.env` is currently a reference to the default environment for *every* Sanctuary module `S`. It would be more useful for `S.env` to reference the particular environment provided when creating `S`, enabling:

```javascript
const S0 = require ('sanctuary');
const S1 = S0.create ({checkTypes: ..., env: S0.env.concat ([...])});
const S2 = S1.create ({checkTypes: ..., env: S1.env.concat ([...])});
```

One would currently need to keep track of the various environments in some other way:

```javascript
const S0 = require ('sanctuary');
const env1 = S0.env.concat ([...]);
const S1 = S0.create ({checkTypes: ..., env: env1});
const env2 = env1.concat ([...]);
const S2 = S1.create ({checkTypes: ..., env: env2});
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -221,13 +221,15 @@
     var Cons = (__doctest.require ('./test/internal/List')).Cons;
     var Sum = __doctest.require ('./test/internal/Sum');
     var S = (function(S) {
-      return S.create ({
+      var S_ = S.create ({
         checkTypes: true,
         env: S.env.concat ([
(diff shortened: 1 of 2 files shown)


Avaq:
I believe this makes progress towards (or is completely?) solving #479

Avaq:
It might also be related to https://github.com/sanctuary-js/sanctuary-def/issues/137

PR #546, opened 2018-05-17
Add addition comments and stying to documentation 

syves:
The 'i' of identity appears in a font which makes it read as an l.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1496,7 +1496,7 @@
   //.
   //. The T ([thrush][]) combinator. Takes a value and a function, and returns
   //. the result of applying the function to the value. Equivalent to Haskell's
-  //. `(&)` function.
+  //. `(&)` function, which is a reverse application operator.
   //.
   //. ```javascript
   //. > S.T (42) (S.add (1))
--- a/index.js
+++ b/index.js
@@ -1449,7 +1449,7 @@
     impl: curry2 (Z.contramap)
   };
 
-  //. ### Combinator
+  //. ### Combinator: A function or definition with no free variables.
 
   //# I :: a -> a
   //.
--- a/index.js
+++ b/index.js
@@ -1453,7 +1453,7 @@
 
   //# I :: a -> a
   //.
-  //. The I combinator. Returns its argument. Equivalent to Haskell's `id`
+  //. The `I` combinator. Returns its argument. Equivalent to Haskell's `id`
   //. function.
   //.
   //. ```javascript


index.js
```diff
@@ -1496,7 +1496,7 @@
   //.
   //. The T ([thrush][]) combinator. Takes a value and a function, and returns
   //. the result of applying the function to the value. Equivalent to Haskell's
-  //. `(&)` function.
+  //. `(&)` function, which is a reverse application operator.
   //.
   //. ```javascript
   //. > S.T (42) (S.add (1))
```
syves: If we give S.T one argument we get back a huge blob of code. How can we return a representation of : (a -> b) -> b?

index.js
```diff
@@ -1449,11 +1449,11 @@
     impl: curry2 (Z.contramap)
   };
 
-  //. ### Combinator
+  //. ### Combinator: A function or definition with no free variables.
```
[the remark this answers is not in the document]
davidchambers: I don't know anything about Church's thesis. :stuck_out_tongue_winking_eye:

I'd like to make a comment about this change which does not relate to your comment above. The following links are generated from the heading text:

  - <https://github.com/sanctuary-js/sanctuary#combinator>
  - <https://sanctuary.js.org/#combinator>

I'd like to preserve these elegant URLs. We could make your proposed addition the section's opening paragraph rather than part of its heading.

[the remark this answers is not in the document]
davidchambers: This is the current implementation of `T`:

```javascript
function T(x) {
  return function(f) {
    return f (x);
  };
}
```

Arrow functions would make the definition clearer:

```javascript
const T = x => f => f (x);
```

This could be written `Txf = fx` which is equivalent to `Txy = yx`. I believe `S.T` really is the thrush combinator.

index.js
davidchambers: Deciding how to represent functions as strings is [tricky][1]. [`show`][2] simply returns `String (f)` for a given function `f`, which will either produce a useful result or a daunting wall of text depending on the nature of `f`. Perhaps in the REPL it would be better to see a placeholder such as `<Function>` in all cases.


[1]: https://github.com/tc39/Function-prototype-toString-revision/blob/master/proposal.html
[2]: https://github.com/sanctuary-js/sanctuary-show#show

index.js
```diff
@@ -1449,11 +1449,11 @@
...
 
   //# I :: a -> a
   //.
-  //. The I combinator. Returns its argument. Equivalent to Haskell's `id`
+  //. The `I` combinator. Returns its argument. Equivalent to Haskell's `id`
```
davidchambers: If we're to change this we should change the other occurrences of combinator names too. I would prefer, though, to reserve `code` formatting for function names and other code snippets. While we *could* address the `I`/`l` ambiguity by updating the formatting of the line above, it would be more appropriate to address the problem in [sanctuary-site][1]. Switching to a typeface with a seriffed `I` would obviate the need for markup changes, though it's an extreme solution!


[1]: https://github.com/sanctuary-js/sanctuary-site

davidchambers:
@syves, these changes will be easier to merge if you submit individual pull requests. :)

PR #565, opened 2018-06-01
fix(typo): `with` should be `which`

jazen:


davidchambers:
Thank you, @jazen. :tada:

PR #571, opened 2018-06-17
documentation: state primary difference between Folktale and Sanctuary

### LINKED ISSUE #366: Compared to Folktale?
Hi there,

It would be quite interesting to hear what are some of the ways that Folktale and Sanctuary differ and compare in their approaches, as well as how they can interoperate (as hinted [here](https://github.com/sanctuary-js/sanctuary/issues/210#issuecomment-217225168))

Thanks!

Jun

davidchambers:
Closes #366

Is this an accurate characterization of Folktale, @robotlolita? Is there anything you would like to add?

### DIFF
--- a/index.js
+++ b/index.js
@@ -43,6 +43,14 @@
 //.
 //. Sanctuary is designed to work in Node.js and in ES5-compatible browsers.
 //.
+//. ## Folktale
+//.
+//. [Folktale][], like Sanctuary, is a standard library for functional
+//. programming in JavaScript. It is well designed and well documented.
+//. Whereas Sanctuary treats JavaScript as a member of the ML language
+//. family, Folktale embraces JavaScript's object-oriented programming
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -4704,6 +4712,7 @@
 //. [Either]:                   #either-type
 //. [Fantasy Land]:             v:fantasyland/fantasy-land
 //. [Foldable]:                 v:fantasyland/fantasy-land#foldable
+//. [Folktale]:                 https://folktale.origamitower.com/
```
davidchambers: You may like to update the link at <https://github.com/folktale> to use HTTPS, @robotlolita. :)

robotlolita:
I think that's good. Folktale is very inspired by Scala.

davidchambers:
Thank you for reviewing this addition, @robotlolita.

PR #579, opened 2018-10-07
docs(foldMap): added an extra example to foldMap

danielo515:
I think foldMap is a great function that has many usages.
However it lacks more examples and people is not used to the concept of giving a typeRep, this tries to make the function more easy to understand with an extra example that uses array.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3338,6 +3338,9 @@
   //. ```javascript
   //. > S.foldMap (String) (f => f.name) ([Math.sin, Math.cos, Math.tan])
   //. 'sincostan'
+  //.
+  //. > S.foldMap (Array) (x => [x + 1, x + 2]) ([10, 20, 30])
+  //. [11, 12, 21, 22, 31, 32]
   //. ```
   _.foldMap = {
     consts: {b: [Z.Monoid], f: [Z.Foldable]},


index.js
```diff
@@ -3330,6 +3330,8 @@
   //. ```javascript
   //. > S.foldMap (String) (f => f.name) ([Math.sin, Math.cos, Math.tan])
   //. 'sincostan'
+  //. > S.foldMap (Array) (x => [x+0, x+2, x+4] ) ([1,2,3])
+  //. [1, 3, 5, 2, 4, 6, 3, 5, 7]
```
davidchambers: Please tweak the formatting like so:

```javascript
> S.foldMap (Array) (x => [x + 0, x + 2, x + 4]) ([1, 2, 3])
[1, 3, 5, 2, 4, 6, 3, 5, 7]
```

`x + 0` looks a bit odd. We could do something like this instead:

```javascript
> S.foldMap (Array) (x => [x + 1, x + 2]) ([10, 20, 30])
[11, 12, 21, 22, 31, 32]
```

Separating the inputs from the increments by an order of magnitude makes it easier to see how the results are interleaved.

Please also include an “empty” line between doctests as we do elsewhere in the file.

danielo515: Thanks for your comments @davidchambers, I agree with all you said

index.js
```diff
@@ -3330,6 +3330,9 @@
   //. ```javascript
   //. > S.foldMap (String) (f => f.name) ([Math.sin, Math.cos, Math.tan])
   //. 'sincostan'
+  //.
+  //. > S.foldMap (Array) (x => [x + 1, x + 2] ) ([10, 20, 30])
```
davidchambers: Please remove the stray space before the second closing paren. ;)

davidchambers:
Thank you for this contribution, @danielo515!

PR #589, opened 2019-02-08
Fix README typo

qm3ster:


davidchambers:
Good catch, @qm3ster! Would you mind making the change to __index.js__ rather than to __README.md__? The latter is automatically generated from the former each time we release a new version of the library.

qm3ster:
Done in #590, I suggest you just merge both to not have to run generate.
Otherwise, just close this one.

davidchambers:
I have a batch of documentation improvements to publish, so I'll merge #590 then release v0.15.1. :)

PR #593, opened 2019-02-17
array: add S.array

### LINKED ISSUE #581: Add more folds
It'd be nice to have deconstructors for more sorts of things in JS. A couple examples from discussions in the gitter room:

```
array :: x -> (a -> [a] -> x) -> [a] -> x
boolean :: x -> x -> Boolean -> x
```

Feel free to add more in the comments.

davidchambers:
Closes #581 (in conjunction with #592)

### DIFF
--- a/index.js
+++ b/index.js
@@ -2924,6 +2924,36 @@
 
   //. ### Array
 
+  //# array :: b -> (a -> Array a -> b) -> Array a -> b
+  //.
+  //. Case analysis for the `Array a` type.
+  //.
+  //. ```javascript
+  //. > S.array (S.Nothing) (head => tail => S.Just (head)) ([])
+  //. Nothing
+  //.
+  //. > S.array (S.Nothing) (head => tail => S.Just (head)) ([1, 2, 3])
+  //. Just (1)
+  //.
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -2899,6 +2899,36 @@
...
+  //. > S.array (S.Nothing) (head => tail => S.Just (tail)) ([])
+  //. Nothing
+  //.
+  //. > S.array (S.Nothing) (head => tail => S.Just (tail)) ([1, 2, 3])
+  //. Just ([2, 3])
+  //. ```
+  function array(y) {
+    return function(f) {
+      return function(xs) {
+        return xs.length === 0 ? y : f (xs[0]) (xs.slice (1));
```
masaeedu: Is the general policy to use underlying methods wherever possible instead of e.g. using `S.slice`?

davidchambers: I'm happy to define specialized functions in terms of fundamental functions, as evidenced by the use of `array` in the implementations of `S.head` and `S.tail`. `array` is the fundamental array function, whereas [`S.slice`][1] is quite complex, accepting negative indices and returning a value of type `Maybe (Array a)`.


[1]: https://sanctuary.js.org/#slice

PR #594, opened 2019-02-20
sanctuary-type-classes@10.0.0

### LINKED ISSUE #574: Implementation of R.clamp
So, the name speaks for itself. 
I vote we should implement [R.clamp](https://ramdajs.com/docs/#clamp) and I agree with the interface given by @davidchambers 

`S.clamp :: Ord a => a -> a -> a -> a`

This means we would not only be able to clamp Numbers, but also Strings and pretty much anything that has an order.

As for implementation ... I would have _no idea_ on how to implement it ....

davidchambers:
Closes #574

<https://github.com/sanctuary-js/sanctuary-type-classes/releases/tag/v10.0.0>

### DIFF
--- a/index.js
+++ b/index.js
@@ -910,6 +910,29 @@
     impl: curry2 (Z.max)
   };
 
+  //# clamp :: Ord a => a -> a -> a -> a
+  //.
+  //. Takes a lower bound, an upper bound, and a value of the same type.
+  //. Returns the value if it is within the bounds; the nearer bound otherwise.
+  //.
+  //. See also [`min`](#min) and [`max`](#max).
+  //.
+  //. ```javascript
+  //. > S.clamp (0) (100) (42)
+  //. 42
+  //.
+  //. > S.clamp (0) (100) (-1)
+  //. 0
+  //.
+  //. > S.clamp ('A') ('Z') ('~')
+  //. 'Z'
+  //. ```
+  _.clamp = {
+    consts: {a: [Z.Ord]},
+    types: [a, a, a, a],
+    impl: curry3 (Z.clamp)
+  };
+
   //# id :: Category c => TypeRep c -> c
   //.
   //. [Type-safe][sanctuary-def] version of [`Z.id`][].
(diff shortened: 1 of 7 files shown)


index.js
```diff
@@ -3241,6 +3264,83 @@
...
+  //. ```
+  _.any = {
+    consts: {f: [Z.Foldable]},
+    types: [$.Predicate (a), f (a), $.Boolean],
+    impl: curry2 (Z.any)
+  };
+
+  //# none :: Foldable f => (a -> Boolean) -> f a -> Boolean
+  //.
+  //. Returns `true` [iff][] none of the elements of the structure satisfies
```
Avaq:
```suggestion
  //. Returns `true` [iff][] none of the elements of the structure satisfy
```

I believe you've made a copy-paste typo here.

index.js
```diff
@@ -3241,6 +3264,83 @@
...
+  //. the predicate.
+  //.
+  //. Properties:
+  //.
+  //.   - `forall p :: a -> Boolean, xs :: Foldable f => f a.
+  //.      S.none (p) (xs) = S.not (S.any (p) (xs))`
```
Avaq: Is this rendered correctly by markdown? I've never seen use of multiline `<code>` tags.

index.js
davidchambers: A thorough review as always, Aldwin!

You have highlighted a [contentious facet of English grammar][1]. I favour _**none** of the … **satisfies**_ as it corresponds to _**not one** of the … **satisfies**_. I consider both options correct, but I prefer _satisfies_ in this sentence.


[1]: https://english.stackexchange.com/q/1425

index.js
davidchambers: It seems to work. See, for example, the [Markdown][1] for [`S.joinWith`][2].


[1]: https://github.com/sanctuary-js/sanctuary/blob/v0.15.1/index.js#L4639-L4644
[2]: https://sanctuary.js.org/#joinWith

index.js
Avaq: I had no idea. Using _none_ as plural seems so foreign to me that I just assumed it was incorrect. I would prefer _satisfy_, but I guess we're just talking about personal preference at this point.

davidchambers: > Using _none_ as plural seems so foreign to me that I just assumed it was incorrect.

To be clear, the usage you find foreign is the singular:

  - _**one** of these pens **is** red_
  - _**not one** of these pens **is** red_
  - _**none** of these pens **is** red_

The usage you prefer is the plural:

  - _these pens **are** red_
  - _**none** of these pens **are** red_

> I guess we're just talking about personal preference at this point.

The spelling and grammar of the documentation reflect my reproducibly inconsistent preferences. :)

Avaq: > To be clear, the usage you find foreign is the singular

Ah, yes. My bad.

PR #600, opened 2019-02-22
parse: document inconsistencies of S.parseDate

### LINKED ISSUE #488: behaviour of S.parseDate is unspecified for some inputs
[MDN][1]:

> __Note:__ parsing of date strings with the `Date` constructor (and `Date.parse`, they are equivalent) is strongly discouraged due to browser differences and inconsistencies. Support for RFC 2822 format strings is by convention only. Support for ISO 8601 formats differs in that date-only strings (e.g. "1970-01-01") are treated as UTC, not local.

#455 is an example of problems that arise when one relies on implementation-dependent behaviour.

It's possible that `S.parseDate` could reject strings in formats we know to be handled inconsistently. This may not be practical, in which case we should remove the function from the library. @syaiful6 raised the possibility of creating a dedicated library for parsing date strings in various formats. :)

:calendar:

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date

davidchambers:
Closes #488

I used `<blockquote>` rather than the equivalent Markdown syntax, because lines beginning with `>` are treated as doctest input. Ideally, only lines beginning with `>` within `` ```javascript `` code blocks would be evaluated, but [doctest][1] doesn't support this use case.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4275,8 +4275,18 @@
 
   //# parseDate :: String -> Maybe ValidDate
   //.
-  //. Takes a string and returns Just the date represented by the string
-  //. if it does in fact represent a date; Nothing otherwise.
+  //. Takes a string `s` and returns `Just (new Date (s))` if `new Date (s)`
+  //. evaluates to a [`ValidDate`][ValidDate] value; Nothing otherwise.
+  //.
+  //. As noted in [#488][], this function's behaviour is unspecified for some
+  //. inputs! [MDN][date parsing] warns against using the `Date` constructor
+  //. to parse date strings:
+  //.
+  //. > __Note:__ parsing of date strings with the `Date` constructor \[…] is
+  //. > strongly discouraged due to browser differences and inconsistencies.
+  //. > Support for RFC 2822 format strings is by convention only. Support for
+  //. > ISO 8601 formats differs in that date-only strings (e.g. "1970-01-01")
+  //. > are treated as UTC, not local.
   //.
   //. ```javascript
   //. > S.parseDate ('2011-01-19T17:40:00Z')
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -4278,6 +4278,20 @@
   //. Takes a string and returns Just the date represented by the string
   //. if it does in fact represent a date; Nothing otherwise.
   //.
+  //. As noted in [#488][], this function's behaviour is unspecified for some
```
Bradcomp: It might be good to explicitly note that we are using the Date constructor to perform the parsing. 
```
Takes a string and returns Just the date represented by the string if it does in fact represent a date, as determined by the native `Date` constructor; Nothing otherwise.
```

davidchambers: Thanks, Brad. Excellent suggestion. I'll make that change.

Bradcomp:
:bowtie:

davidchambers:
I'll update this pull request once davidchambers/doctest#110 has been merged and released. :)

index.js
```diff
@@ -4275,8 +4275,18 @@
...
+  //.
+  //. As noted in [#488][], this function's behaviour is unspecified for some
+  //. inputs! [MDN][date parsing] warns against using the `Date` constructor
+  //. to parse date strings:
+  //.
+  //. > __Note:__ parsing of date strings with the `Date` constructor \[…] is
+  //. > strongly discouraged due to browser differences and inconsistencies.
+  //. > Support for RFC 2822 format strings is by convention only. Support for
+  //. > ISO 8601 formats differs in that date-only strings (e.g. "1970-01-01")
+  //. > are treated as UTC, not local.
```
davidchambers: Thanks to davidchambers/doctest#110, sanctuary-js/sanctuary-scripts#20, and #607 we are now able to use Markdown syntax for this block quotation. Quite the diversion!

As explained in remarkjs/remark-lint#207, it is necessary to use `\[…]` rather than simply `[…]` to avoid triggering a `no-undefined-references` error.

PR #605, opened 2019-03-09
test: use ES6 syntax

davidchambers:
Now that we're no longer running tests in browsers (#603) we're no longer restricted to ES5 syntax.


svozza:
Nice!

davidchambers:
I thought you would like this, Stefano. :D

svozza:
You were correct!

PR #610, opened 2019-03-17
array: generalize S.head, S.last, S.tail, and S.init

### LINKED ISSUE #570: Update head and tail to work for all Foldable
Currently Head and Tail only work for `Array`, but it is possible to implement them for all `Foldable`.

syves:
Closes #570

Initial attempt to generalize head, tail, last and init functions for Foldables.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3063,10 +3063,10 @@
     impl: at
   };
 
-  //# head :: Array a -> Maybe a
+  //# head :: Foldable f => f a -> Maybe a
   //.
-  //. Returns Just the first element of the given array if the array contains
-  //. at least one element; Nothing otherwise.
+  //. Returns Just the first element of the given structure if the structure
+  //. contains at least one element; Nothing otherwise.
   //.
   //. ```javascript
   //. > S.head ([1, 2, 3])
@@ -3074,17 +3074,32 @@
   //.
   //. > S.head ([])
   //. Nothing
+  //.
+  //. > S.head (Cons (1) (Cons (2) (Cons (3) (Nil))))
+  //. Just (1)
+  //.
+  //. > S.head (Nil)
+  //. Nothing
   //. ```
+  function head(foldable) {
+    //  Fast path for arrays.
+    if (Array.isArray (foldable)) {
+      return foldable.length > 0 ? Just (foldable[0]) : Nothing;
+    }
+    return Z.reduce (function(m, x) { return m.isJust ? m : Just (x); },
+                     Nothing,
+                     foldable);
+  }
   _.head = {
-    consts: {},
-    types: [$.Array (a), $Maybe (a)],
-    impl: array (Nothing) (B (K) (Just))
+    consts: {f: [Z.Foldable]},
+    types: [f (a), $Maybe (a)],
+    impl: head
   };
 
-  //# last :: Array a -> Maybe a
+  //# last :: Foldable f => f a -> Maybe a
   //.
-  //. Returns Just the last element of the given array if the array contains
-  //. at least one element; Nothing otherwise.
+  //. Returns Just the last element of the given structure if the structure
+  //. contains at least one element; Nothing otherwise.
   //.
   //. ```javascript
   //. > S.last ([1, 2, 3])
@@ -3092,20 +3107,31 @@
   //.
   //. > S.last ([])
   //. Nothing
+  //.
+  //. > S.last (Cons (1) (Cons (2) (Cons (3) (Nil))))
+  //. Just (3)
+  //.
+  //. > S.last (Nil)
+  //. Nothing
   //. ```
-  function last(xs) {
-    return xs.length > 0 ? Just (xs[xs.length - 1]) : Nothing;
+  function last(foldable) {
+    //  Fast path for arrays.
+    if (Array.isArray (foldable)) {
+      return foldable.length > 0 ? Just (foldable[foldable.length - 1])
+                                 : Nothing;
+    }
+    return Z.reduce (function(_, x) { return Just (x); }, Nothing, foldable);
   }
   _.last = {
-    consts: {},
-    types: [$.Array (a), $Maybe (a)],
+    consts: {f: [Z.Foldable]},
+    types: [f (a), $Maybe (a)],
     impl: last
   };
 
-  //# tail :: Array a -> Maybe (Array a)
+  //# tail :: (Applicative f, Foldable f, Monoid (f a)) => f a -> Maybe (f a)
   //.
-  //. Returns Just all but the first of the given array's elements if the
-  //. array contains at least one element; Nothing otherwise.
+  //. Returns Just all but the first of the given structure's elements if the
+  //. structure contains at least one element; Nothing otherwise.
   //.
   //. ```javascript
   //. > S.tail ([1, 2, 3])
@@ -3113,17 +3139,33 @@
   //.
   //. > S.tail ([])
   //. Nothing
+  //.
+  //. > S.tail (Cons (1) (Cons (2) (Cons (3) (Nil))))
+  //. Just (Cons (2) (Cons (3) (Nil)))
+  //
+  //. > S.tail (Nil)
+  //. Nothing
   //. ```
+  function tail(foldable) {
+    //  Fast path for arrays.
+    if (Array.isArray (foldable)) {
+      return foldable.length > 0 ? Just (foldable.slice (1)) : Nothing;
+    }
+    var empty = Z.empty (foldable.constructor);
+    return Z.reduce (function(m, x) {
+      return Just (maybe (empty) (append (x)) (m));
+    }, Nothing, foldable);
+  }
   _.tail = {
-    consts: {},
-    types: [$.Array (a), $Maybe ($.Array (a))],
-    impl: array (Nothing) (K (Just))
+    consts: {f: [Z.Applicative, Z.Foldable, Z.Monoid]},
+    types: [f (a), $Maybe (f (a))],
+    impl: tail
   };
 
-  //# init :: Array a -> Maybe (Array a)
+  //# init :: (Applicative f, Foldable f, Monoid (f a)) => f a -> Maybe (f a)
   //.
-  //. Returns Just all but the last of the given array's elements if the
-  //. array contains at least one element; Nothing otherwise.
+  //. Returns Just all but the last of the given structure's elements if the
+  //. structure contains at least one element; Nothing otherwise.
   //.
   //. ```javascript
   //. > S.init ([1, 2, 3])
@@ -3131,13 +3173,26 @@
   //.
   //. > S.init ([])
   //. Nothing
+  //.
+  //. > S.init (Cons (1) (Cons (2) (Cons (3) (Nil))))
+  //. Just (Cons (1) (Cons (2) (Nil)))
+  //.
+  //. > S.init (Nil)
+  //. Nothing
   //. ```
-  function init(xs) {
-    return xs.length > 0 ? Just (xs.slice (0, -1)) : Nothing;
+  function init(foldable) {
+    //  Fast path for arrays.
+    if (Array.isArray (foldable)) {
+      return foldable.length > 0 ? Just (foldable.slice (0, -1)) : Nothing;
+    }
+    var empty = Z.empty (foldable.constructor);
+    return Z.map (Pair.snd, Z.reduce (function(m, x) {
+      return Just (Pair (x) (maybe (empty) (pair (append)) (m)));
+    }, Nothing, foldable));
   }
   _.init = {
-    consts: {},
-    types: [$.Array (a), $Maybe ($.Array (a))],
+    consts: {f: [Z.Applicative, Z.Foldable, Z.Monoid]},
+    types: [f (a), $Maybe (f (a))],
     impl: init
   };
 
@@ -3379,10 +3434,15 @@
   //. > S.append ([3]) (S.Just ([1, 2]))
   //. Just ([1, 2, 3])
   //. ```
+  function append(x) {
+    return function(xs) {
+      return Z.append (x, xs);
+    };
+  }
   _.append = {
     consts: {f: [Z.Applicative, Z.Semigroup]},
     types: [a, f (a), f (a)],
-    impl: curry2 (Z.append)
+    impl: append
   };
 
   //# prepend :: (Applicative f, Semigroup (f a)) => a -> f a -> f a
--- a/test/head.js
+++ b/test/head.js
@@ -1,7 +1,8 @@
 'use strict';
 
-const S = require ('..');
+const S = require ('./internal/sanctuary');
 
+const {Nil, Cons} = require ('./internal/List');
 const eq = require ('./internal/eq');
 
 
@@ -9,11 +10,16 @@ test ('head', () => {
 
   eq (typeof S.head) ('function');
   eq (S.head.length) (1);
-  eq (S.show (S.head)) ('head :: Array a -> Maybe a');
+  eq (S.show (S.head)) ('head :: Foldable f => f a -> Maybe a');
 
   eq (S.head ([])) (S.Nothing);
   eq (S.head (['foo'])) (S.Just ('foo'));
   eq (S.head (['foo', 'bar'])) (S.Just ('foo'));
   eq (S.head (['foo', 'bar', 'baz'])) (S.Just ('foo'));
+
+  eq (S.head (Nil)) (S.Nothing);
+  eq (S.head (Cons ('foo') (Nil))) (S.Just ('foo'));
+  eq (S.head (Cons ('foo') (Cons ('bar') (Nil)))) (S.Just ('foo'));
+  eq (S.head (Cons ('foo') (Cons ('bar') (Cons ('baz') (Nil))))) (S.Just ('foo'));
 
 });
--- a/test/init.js
+++ b/test/init.js
@@ -1,7 +1,8 @@
 'use strict';
 
-const S = require ('..');
+const S = require ('./internal/sanctuary');
 
+const {Nil, Cons} = require ('./internal/List');
 const eq = require ('./internal/eq');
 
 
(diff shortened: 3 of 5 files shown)


index.js
```diff
@@ -1,7 +1,4 @@
 /*    #######
-   ####     ####
- ####   ###   ####
-#####   ###########   sanctuary
```
davidchambers: Please reinstate these three lines. :laughing:

index.js
```diff
@@ -3073,38 +3087,78 @@
   //.
   //. > S.last ([])
   //. Nothing
+  //.
+  //. > S.last (Cons ('baz') (Cons ('foo') (Nil)))
+  //. S.Just ('foo')
```
davidchambers: I'd prefer to see `S.last (Cons ('foo') (Cons ('bar') (Cons ('baz') (Nil))))` here. Deviating from the customary _foo_, _bar_, _baz_ order could lead to confusion.

index.js
```diff
@@ -3073,38 +3087,78 @@
...
+  //. Foldable contains at least one element; Nothing otherwise.
   //.
   //. ```javascript
   //. > S.tail ([1, 2, 3])
   //. Just ([2, 3])
   //.
   //. > S.tail ([])
+  //.
+  //. > S.tail (Cons ('baz') (Cons ('foo') (Nil)))
+  //. Just (Cons ('foo') (Nil))
```
davidchambers: Let's stick with _foo_, _bar_, _baz_ order here as well.

index.js
```diff
@@ -3073,38 +3087,78 @@
...
+  //
+  //. > S.tail (Nil)
   //. Nothing
   //. ```
+
+  function tail(foldable) {
+    var counter = 0
```
davidchambers: Is a counter necessary? Could you infer whether or not you're processing the first element from the value of the accumulator?

test/head.js
```diff
@@ -1,19 +1,22 @@
...
-  eq (S.show (S.head)) ('head :: Array a -> Maybe a');
+  eq (S.show (S.head)) ('head :: Foldable f => f a -> Maybe a');
 
   eq (S.head ([])) (S.Nothing);
   eq (S.head (['foo'])) (S.Just ('foo'));
   eq (S.head (['foo', 'bar'])) (S.Just ('foo'));
   eq (S.head (['foo', 'bar', 'baz'])) (S.Just ('foo'));
 
+  eq (S.head (Cons ('foo') (Cons ('bar') (Cons ('baz') (Nil))))) (S.Just ('foo'));
+  eq (S.head (Nil)) (S.Nothing);
```
davidchambers: Let's include four list tests in the same order as the array tests.

test/init.js
```diff
@@ -9,11 +9,11 @@ test ('init', () => {
...
   eq (typeof S.init) ('function');
   eq (S.init.length) (1);
-  eq (S.show (S.init)) ('init :: Array a -> Maybe (Array a)');
+  eq (S.show (S.init)) ('init :: (Foldable f, Applicative f) => f a -> Maybe (f a)');
 
   eq (S.init ([])) (S.Nothing);
   eq (S.init (['foo'])) (S.Just ([]));
   eq (S.init (['foo', 'bar'])) (S.Just (['foo']));
-  eq (S.init (['foo', 'bar', 'baz'])) (S.Just (['foo', 'bar']));
+  //eq (S.init (['foo', 'bar', 'baz'])) (S.Just (['foo', 'bar']));
```
davidchambers: Please remove this line, and add four list tests.

test/tail.js
```diff
@@ -1,19 +1,22 @@
...
+  eq (S.show (S.tail)) ('tail :: (Foldable f, Applicative f) => f a -> Maybe (f a)');
 
   eq (S.tail ([])) (S.Nothing);
   eq (S.tail (['foo'])) (S.Just ([]));
   eq (S.tail (['foo', 'bar'])) (S.Just (['bar']));
   eq (S.tail (['foo', 'bar', 'baz'])) (S.Just (['bar', 'baz']));
 
+  eq (S.tail (Cons ('baz') (Cons ('foo') (Nil)))) (S.Just (Cons ('foo') (Nil)));
+  eq (S.tail (Nil)) (S.Nothing);
 });
```
davidchambers: Please insert a blank line after the last assertion.

index.js
```diff
@@ -3073,38 +3087,78 @@
...
   _.last = {
-    consts: {},
-    types: [$.Array (a), $Maybe (a)],
+    consts: {f: [Z.Foldable]},
+    types: [f (a), $Maybe (a)],
     impl: last
   };
 
-  //# tail :: Array a -> Maybe (Array a)
+  //# tail :: (Foldable f, Applicative f) => f a -> Maybe (f a)
```
davidchambers: Could you justify the Applicative constraint?

index.js
davidchambers: The expected output should now be `Just ('baz')`, I believe. You could run `npm run doctest` to check.

index.js
```diff
@@ -3073,52 +3090,121 @@
...
-  //. array contains at least one element; Nothing otherwise.
+  //. Returns a Just of all but the first of the given Foldable if the
+  //. Foldable contains at least one element; Nothing otherwise.
   //.
   //. ```javascript
   //. > S.tail ([1, 2, 3])
   //. Just ([2, 3])
   //.
   //. > S.tail ([])
+  //.
```
davidchambers: The doctest above lost its output line.

index.js
```diff
@@ -3073,52 +3090,121 @@
...
+  //. > S.init (Cons ('bar') (Nil))
+  //. Just (Nil)
+  //.
+  //. > S.init (Nil)
+  //. Nothing
   //. ```
-  function init(xs) {
-    return xs.length > 0 ? Just (xs.slice (0, -1)) : Nothing;
+
+//. ```
```
davidchambers: This line can be deleted, I think.

index.js
syves: 🆗 ok

index.js
syves: 🆗

test/head.js
syves: 🆗

test/init.js
syves: 🆗

index.js
syves: In order to build a new foldable with the right order I need to use 'append' which requires that f be an Applicative and Semigroup. I'm now a bit confused about why it is working without both being required.

index.js
syves: 🆗

test/tail.js
syves: This file has been updated.

index.js
syves: 🆗

index.js
davidchambers: > In order to build a new foldable with the right order I need to use 'append' which requires that f be an Applicative and Semigroup.

Good point! It looks as though you're also using [`Z.empty`][1], so there may be other constraints to consider.

> I'm now a bit confused about why it is working without both being required.

Type classes need not be specified in order for a function to work. In fact, the only reason to specify a function's types and type-class constraints is to ensure that *invalid* applications of the function result in errors with descriptive messages. For example:

```javascript
S.concat (123) (456);
//
//   concat :: Semigroup a => a -> a -> a
//             ^^^^^^^^^^^    ^
//                            1
//
//   1)  123 :: Number, FiniteNumber, NonZeroFiniteNumber, Integer, NonNegativeInteger, ValidNumber
//
//   ‘concat’ requires ‘a’ to satisfy the Semigroup type-class constraint; the value at position 1 does not.
//
//   See https://github.com/sanctuary-js/sanctuary-type-classes/tree/v10.0.0#Semigroup for information about the sanctuary-type-classes/Semigroup type class.
```

If we replace `{a: [Z.Semigroup]}` with `{}` and try again, the error is less helpful:

```javascript
S.concat (123) (456);
// ! TypeError: Semigroup.methods.concat(...) is not a function
```


[1]: https://github.com/sanctuary-js/sanctuary-type-classes/tree/v10.0.0#empty

index.js
davidchambers: This output looks incorrect.

test/last.js
```diff
@@ -1,19 +1,26 @@
...
   eq (S.last (['foo'])) (S.Just ('foo'));
   eq (S.last (['foo', 'bar'])) (S.Just ('bar'));
   eq (S.last (['foo', 'bar', 'baz'])) (S.Just ('baz'));
 
+  eq (S.last (Nil)) (S.Nothing);
+  eq (S.last (Cons ('foo') (Nil))) (S.Just ('foo'))
+  eq (S.last (Cons ('foo') (Cons ('bar') (Nil))))
+  (S.Just ('bar'));
+  eq (S.last (Cons ('foo') (Cons ('bar') (Cons ('baz') (Nil)))))
+     (S.Just ('baz'));
```
davidchambers: Don't worry about line length in the test files. These assertions would read better as one-liners.

index.js
syves: 🆗

index.js
syves: Verbose messages are much more helpful.

test/last.js
syves: 🆗

syves:
I've addressed the above changes, but for some reason doctest is not running.

> 20 error code ELIFECYCLE
> 21 error errno 1
> 22 error sanctuary@1.0.0 doctest: `sanctuary-doctest`
> 22 error Exit status 1
> 23 error Failed at the sanctuary@1.0.0 doctest script.

test/last.js
davidchambers: It looks as though you deleted some of the assertions. I was suggesting that you remove the line breaks rather than the whole lines. ;)

index.js
davidchambers: > It looks as though you're also using [`Z.empty`][1], so there may be other constraints to consider.

Have you given this some thought?


[1]: https://github.com/sanctuary-js/sanctuary-type-classes/tree/v10.0.0#empty

syves: Now I have :)

syves: 🆗

index.js
```diff
@@ -3044,81 +3044,151 @@
...
+  //. > S.head (Cons ('foo') (Cons ('bar') (Nil)))
+  //. S.Just ('foo')
+  //.
+  //. > S.head (Nil)
+  //. S.Nothing
   //. ```
+
+  function head(foldable) {
+    return Z.reduce (
+      function(acc, a) {
```
davidchambers: For consistency with other implementations, let's name these parameters `maybe` and `x`.

index.js
```diff
@@ -3044,81 +3044,151 @@
     impl: at
   };
 
-  //# head :: Array a -> Maybe a
+  //# head :: Foldable f => f a -> Maybe a
   //.
-  //. Returns Just the first element of the given array if the array contains
-  //. at least one element; Nothing otherwise.
+  //. Returns Just the first element of the given collection if the foldable
+  //. contains at least one element; Nothing otherwise.
```
davidchambers: We avoid the term _collection_ because people tend not to think of values of types such as `Identity a`, `Maybe a`, and `Either a b` as collections.

I think we should also avoid _foldable_. Referring to a type class works in this case, but this approach would be problematic for functions which place multiple constraints on a single type variable. Let's avoid referring to _applicative semigroups_ and _foldable applicative monoids_. :joy:

_Structure_ is the best term we have considered thus far.

index.js
syves: :ok done

index.js
syves: 🆗

index.js
syves: applied to all 4

index.js
syves: I understand this now.

index.js
```diff
@@ -3044,81 +3044,151 @@
...
   _.last = {
-    consts: {},
-    types: [$.Array (a), $Maybe (a)],
+    consts: {f: [Z.Foldable]},
+    types: [f (a), $Maybe (a)],
     impl: last
   };
 
-  //# tail :: Array a -> Maybe (Array a)
+  //# tail :: (Applicative f, Foldable f, Monoid f) => f a -> Maybe (f a)
```
davidchambers:
```suggestion
  //# tail :: (Applicative f, Foldable f, Monoid (f a)) => f a -> Maybe (f a)
```

It's not yet possible to precisely describe this function's type using `def`, but the textual signature can be precise. The third constraint should be `Monoid (f a)` rather than `Monoid f`. `Maybe a`, for example, satisfies Monoid only if `a` satisfies Semigroup ([source][1]). The question we want to ask is not whether `Maybe` satisfies Monoid, but whether `Maybe String` satisfies Monoid. Does this make sense?


[1]: https://github.com/sanctuary-js/sanctuary-maybe/blob/v1.1.0/index.js#L117

index.js
```diff
@@ -3044,81 +3044,151 @@
...
+      function(maybe, x) {
+        return maybe.isNothing ?
+               Just (Z.empty (foldable.constructor)) :
+               Z.map (append (x), maybe);
+      },
+      Nothing,
+      foldable
+    );
+  }
+
```
davidchambers: The empty lines before and after the implementations can be removed.

index.js
```diff
@@ -3044,81 +3044,151 @@
...
   //. Just (1)
   //.
   //. > S.head ([])
   //. Nothing
+  //.
+  //. > S.head (Cons ('foo') (Cons ('bar') (Nil)))
+  //. S.Just ('foo')
+  //.
+  //. > S.head (Nil)
+  //. S.Nothing
```
davidchambers: For the expected output we like to use `Nothing` and `Just` rather than `S.Nothing` and `S.Just`. This makes the doctests for [`S.match`][1] and several other functions less noisy.


[1]: https://sanctuary.js.org/#match

index.js
syves: yeah that makes sense

index.js
syves: 🆗

index.js
syves: 🆗

index.js
```diff
@@ -3044,81 +3044,150 @@
...
+  //. > S.tail (Nil)
+  //. Nothing
   //. ```
+
+  function tail(foldable) {
+    return Z.reduce (
+      function(maybe, x) {
+        return maybe.isNothing ?
+               Just (Z.empty (foldable.constructor)) :
+               Z.map (append (x), maybe);
```
davidchambers: Something doesn't feel right about using `Z.map` in conjunction with the `isNothing` check. If we're to keep the `isNothing` check we could safely use `maybe.value` in the Just case:

```javascript
return maybe.isNothing ?
       Just (Pair (Z.empty (foldable.constructor)) (x)) :
       Just (Pair (Z.append (maybe.value.snd, maybe.value.fst)) (x));
```

This could be improved by removing the common code from the branches:

```javascript
return Just (maybe.isNothing ?
             Pair (Z.empty (foldable.constructor)) (x) :
             Pair (Z.append (maybe.value.snd, maybe.value.fst)) (x));
```

```javascript
return Just (Pair (maybe.isNothing ?
                   Z.empty (foldable.constructor) :
                   Z.append (maybe.value.snd, maybe.value.fst))
                  (x));
```

We could even remove the explicit conditional logic by using `Z.reduce`:

```javascript
return Just (Pair (Z.reduce (function(_, pair) { return Z.append (pair.snd, pair.fst); },
                             Z.empty (foldable.constructor),
                             maybe))
                  (x));
```

The problem with this approach is that `Z.empty (foldable.constructor)` would be evaluated once for each element of the structure. Can you think of a solution to this problem?

index.js
davidchambers: > Returns Just the first element of the given structure if *it* contains at least one element; Nothing otherwise.

This is ambiguous: _it_ could refer to _the given structure_ or to _the first element of the given structure_. Let's use _the structure_ in place of _it_ to be clear, even though it's repetitive.

index.js
```diff
@@ -3044,81 +3044,150 @@
...
   //. > S.head ([])
   //. Nothing
+  //.
+  //. > S.head (Cons ('foo') (Cons ('bar') (Nil)))
+  //. Just ('foo')
+  //.
+  //. > S.head (Nil)
+  //. Nothing
   //. ```
+
```
davidchambers: Let's remove the empty lines preceding the implementations.

index.js
syves: 🆗

index.js
syves: 🆗

index.js
syves: 🆗

index.js
davidchambers: The comment above applies to `init` rather than to `tail`. I'm sorry for causing confusion.

syves: I stored the empty in a var.

davidchambers:
Let's complete this pull request together. I will push some commits to your branch. :)

index.js
```diff
@@ -3063,81 +3063,123 @@
...
     impl: last
   };
 
-  //# tail :: Array a -> Maybe (Array a)
+  //# tail :: (Applicative f, Foldable f, Monoid (f a)) => f a -> Maybe (f a)
   //.
-  //. Returns Just all but the first of the given array's elements if the
-  //. array contains at least one element; Nothing otherwise.
+  //. Returns a Just of all but the first of the given structure if the
+  //. structure contains at least one element; Nothing otherwise.
```
davidchambers: We lost the word _elements_ from this sentence.

The change from _Just_ to _a Just of_ should be discussed in a separate pull request or issue. If we're to make this change we should do so for all descriptions, not just those we happen to be editing.

index.js
```diff
@@ -3063,81 +3063,123 @@
...
+  //. > S.tail (Nil)
+  //. Nothing
   //. ```
+  function tail(foldable) {
+    return Z.reduce (function(m, x) {
+      return m.isNothing ?
+             Just (Z.empty (foldable.constructor)) :
+             Z.map (append (x), m);
+    }, Nothing, foldable);
+  }
```
davidchambers: What do you think of this change, @syves?

```diff
   function tail(foldable) {
+    var empty = Z.empty (foldable.constructor);
     return Z.reduce (function(m, x) {
-      return m.isNothing ?
-             Just (Z.empty (foldable.constructor)) :
-             Z.map (append (x), m);
+      return Just (maybe (empty) (append (x)) (m));
     }, Nothing, foldable);
   }
```

It's very similar to the change we made to remove the `isNothing` check from `init`.

index.js
syves: I understand. There should be a conversation about why the change is or isn't an improvement.

syves: 🆗

index.js
syves: This is very elegant and easier to read.

index.js
davidchambers: Let's drop _of_ as well.

index.js
syves: 🆗

davidchambers:
I just added fast paths for arrays, so these functions will continue to operate efficiently on arrays. @syves, if you're happy with these changes please squash the commits in preparation for merging. :)

davidchambers:
Thank you, @syves!

PR #612, opened 2019-03-17
strmap: add S.value

### LINKED ISSUE #611: S.get and S.gets without first argument
Hello there, I'm relatively new at FP in JS and i'm finding this library quire useful.

I am a heavy Ramda user and two of the functions I use most are [`R.prop`](https://ramdajs.com/docs/#prop) and [`R.path`](https://ramdajs.com/docs/#path). 

What I really like about both of them is that they allow me to get a value in a safe way.
If the prop or the path is missing in the provided object, `undefined` is returned.

It would be perfect if:
1) the prop/path exists, a `Just(value)` is returned
2) the prop/path does not exists a `Nothing` is returned

Your library implements an `S.prop` and an `S.props` having a different behavior.
Indeed in case 1 the value is returned, while in case 2 an error is thrown.

So in your documentation it is suggested to use the `S.get`/`S.gets` functions to access a property/path of uncertain objects.
But in such a case the functions signature is different because I need - first of all - to provide a predicate.

Eventually I ended up to use a combination of `monet.js` and `ramda` to obtain a function satisfying point 1 and 2 of my previous list.
For example, this is a possible solution (where R is ramda and Maybe is from monet). 

```
const propMaybe = prop => R.pipe(
  R.prop(prop),
  Maybe.fromNull,
)
```

As I said at the beginning, I am quite new in the FP world, so I am sure I am missing the point. But wouldn't be useful a something like the function I described?

I think you have already something very similar that is `S.head`.

Thank you in advance for your answers.

davidchambers:
Closes #611

Please let me know if you prefer `lookup` or some other name.

/cc @daaaaa

### DIFF
--- a/index.js
+++ b/index.js
@@ -487,11 +487,6 @@
     };
   }
 
-  //  value :: { value :: a } -> a
-  function value(r) {
-    return r.value;
-  }
-
   //  :: Type
   var a = $.TypeVariable ('a');
   var b = $.TypeVariable ('b');
@@ -2283,7 +2278,7 @@
   //. ['foo', 'baz']
   //. ```
   function justs(maybes) {
-    return map (value) (filter (isJust) (maybes));
+    return map (prop ('value')) (filter (isJust) (maybes));
   }
   _.justs = {
     consts: {f: [Z.Filterable, Z.Functor]},
@@ -2557,7 +2552,7 @@
   _.lefts = {
     consts: {f: [Z.Filterable, Z.Functor]},
     types: [f ($Either (a) (b)), f (a)],
-    impl: B (map (value)) (filter (isLeft))
+    impl: B (map (prop ('value'))) (filter (isLeft))
   };
 
   //# rights :: (Filterable f, Functor f) => f (Either a b) -> f b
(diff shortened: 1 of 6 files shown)


Bradcomp:
The code looks good. 

I think I may prefer `lookup` over `value`, but not strongly. `value` just seems really ambiguous. We are in a little pickle because our current names aren't consistent:
get - verb: would suggest lookup as the closest analogue
prop - noun: would suggest value as the closest analogue

davidchambers:
Thank you for your feedback, @Bradcomp. :)

> I think I may prefer `lookup` over `value`, but not strongly. `value` just seems really ambiguous.

Does seeing the types of `value` and `values` together change your perspective?

```haskell
value  :: String -> StrMap a -> Maybe a
values ::           StrMap a -> Array a
```

I see this function as the singular of [`S.values`][1], so `value` seems natural to me. I agree that `value` is not descriptive, but neither is `values`. :rofl:

Although naming is an important aspect of API design, even nondescriptive names eventually become meaningful through association. [`S.ap`][2], anyone? :)

Does anyone wish to further support `lookup`? I am open to having my mind changed.


[1]: https://sanctuary.js.org/#values
[2]: https://sanctuary.js.org/#ap

PR #618, opened 2019-04-05
maybe, either: remove binary and ternary “encase” functions

davidchambers:
As the additions to the test suite demonstrate, the binary and ternary “encase” functions offer nothing more than what can be achieved via nested function composition.


davidchambers:
> This can be thought of as an optimization we missed after moving to simple currying.

Indeed. Simple currying provides numerous benefits, some of which are taking us time to discover. :)

PR #620, opened 2019-04-05
maybe, either: unify S.encase and S.encaseEither

davidchambers:
There's little value in having both `S.encase` and `S.encaseEither`: every use of the former could be rewritten in terms of the latter with help from [`S.eitherToMaybe`][1]. Thus we should remove `S.encase`.

This vacates the desirable name, so we should then rename `S.encaseEither`.

Finally, we should drop the `Error -> l` argument as it is unnecessary in the presence of [`S.mapLeft`][2].

In summary:

```haskell
--  Before
encase       ::                 (a -> b) -> a -> Maybe b
encaseEither :: (Error -> l) -> (a -> r) -> a -> Either l r

--  After
encase       ::                 (a -> b) -> a -> Either Error b
```

In conjunction with #618, this will reduce the number of “encase” functions from six to one. :)

[1]: https://sanctuary.js.org/#eitherToMaybe
[2]: https://sanctuary.js.org/#mapLeft


toolsh:
> I think this is a good change. `try/catch` is more suited to Either than Maybe anyway.

The change is useful.
However, typescript @types/sanctuary need related updates too.


> import S = require("sanctuary");
{}
> import which = require("which");
undefined
> var adb = S.encase(which.sync)("adb");
undefined
> S.type(adb);
{ namespace: Just ("sanctuary-either"),
  name: 'Either',
  version: 1 }
> adb.value
[eval].ts:5:5 - error TS2339: Property 'value' does not exist on type 'Maybe<string>'.
5 adb.value
      ~~~~~
undefined
>

PR #630, opened 2019-04-26
array: generalize S.takeLast and S.dropLast

davidchambers:
The implementations are not as efficient as they could be, but I want to avoid spreading complexity currently confined to `_takeDrop`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3003,59 +3003,69 @@
     impl: drop
   };
 
-  //# takeLast :: Integer -> Array a -> Maybe (Array a)
+  //# takeLast :: (Applicative f, Foldable f, Monoid (f a)) => Integer -> f a -> Maybe (f a)
   //.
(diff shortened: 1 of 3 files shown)


masaeedu:
If I'm right about the applicative constraint not being required in #627, then we probably don't need it here as well.

davidchambers:
You have raised a good point, @Bradcomp; I had not considered infinite structures. I would not *expect* these functions to work on infinite structures, so I do not find the behaviour surprising, but it is a shame that the type signatures do not accurately convey the constraints.

<https://github.com/sanctuary-js/sanctuary/pull/627#issuecomment-487275940> applies here too, @masaeedu, so I will merge this pull request in its current form. If it turns out that I have misunderstood the constraints, we can relax the constraints on `S.take`, `S.drop`, `S.takeLast`, and `S.dropLast` in another pull request. :)

PR #634, opened 2019-04-27
documentation: use “that” rather than “which” in defining clauses

davidchambers:
I discovered recently that I've been misusing _which_ for decades.

<https://www.grammarly.com/blog/which-vs-that/>:

> A defining clause […] gives information essential to the meaning of the sentence. _That_ is used in defining clauses.


svozza:
That article really brings out the descriptivist in me!

PR #662, opened 2020-01-10
either: generalize S.encase to permit a value of any type to be thrown

### LINKED ISSUE #646: Can we rework encase to allow any type to be thrown?
Something like this:

```diff
- encase :: (a -> b) -> a -> Either Error b
+ encase :: Throwing e (a -> b) -> a -> Either e b
```

Because I'm getting bitten by the fact that I expect a function to be safe after using `encase`, only to find out that it now throws a TypeError because the function I've encased throws a String.

davidchambers:
Closes #646

### DIFF
--- a/index.js
+++ b/index.js
@@ -480,6 +480,24 @@
   var p = $.BinaryTypeVariable ('p');
   var s = $.BinaryTypeVariable ('s');
 
+  //  Throwing :: Type -> Type -> Type -> Type
+  //
+  //  `Throwing e a b` is the type of functions from `a` to `b` that may
+  //  throw values of type `e`.
+  function Throwing(E) {
+    return function(A) {
+      return function(B) {
+        var T = $.Fn (A) (B);
+        T.format = function(outer, inner) {
+          return outer ('Throwing ' + show (E)) +
+                 outer (' ') + inner ('$1') (show (A)) +
+                 outer (' ') + inner ('$2') (show (B));
+        };
+        return T;
+      };
+    };
+  }
+
   //  TypeRep :: Type -> Type
   var TypeRep = $.UnaryType
     ('TypeRep')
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -480,6 +480,24 @@
...
+        T.format = function(outer, inner) {
+          return outer ('Throwing ' + show (E)) +
+                 outer (' ') + inner ('$1') (show (A)) +
+                 outer (' ') + inner ('$2') (show (B));
+        };
+        return T;
+      };
+    };
+  }
+
```
davidchambers: Verifying the correctness of this code is a little tricky because an exception thrown during the type checking of `(a -> b)` is caught by `encase`. :joy:

index.js
```diff
@@ -480,6 +480,24 @@
...
 
+  //  Throwing :: Type -> Type -> Type -> Type
+  //
+  //  `Throwing e a b` is the type of functions from `a` to `b` that may
+  //  throw values of type `e`.
+  function Throwing(E) {
+    return function(A) {
+      return function(B) {
+        var T = $.Fn (A) (B);
+        T.format = function(outer, inner) {
```
Avaq: Okay, so the implementation of this is just `Fn (A) (B)`, but with a different way to display the type in signatures.

index.js
```diff
@@ -2420,7 +2438,7 @@
   }
   _.encase = {
     consts: {},
-    types: [$.Fn (a) (b), a, $.Either ($.Error) (b)],
+    types: [Throwing (e) (a) (b), a, $.Either (e) (b)],
```
Avaq: ...so this change should affect the way things are displayed, and nothing more.

Avaq:
...And the rest of the changes are just tests. So what part of the code makes it so `encase` doesn't throw anymore? o.O

index.js
Avaq: Oh! I just found the answer to my own question! It's the change from `$.Error` to `e` in the return type.

PR #664, opened 2020-01-16
maybe, either: remove S.maybeToEither and S.eitherToMaybe

davidchambers:
Before:

```javascript
S.maybeToEither (x)
```

```javascript
S.eitherToMaybe
```

After:

```javascript
S.maybe (S.Left (x))
        (S.Right)
```

```javascript
S.either (S.K (S.Nothing))
         (S.Just)
```

I appreciate the concern @Avaq expressed in <https://github.com/sanctuary-js/sanctuary/pull/644#issuecomment-518285935>:

> My only concern is that in implementing the entire matrix of conversions between `a`, `Maybe a`, `Either a b`, and `Either b a`, we might be increasing API surface drastically for functions that can be expressed as simple compositions.

If others agree with this change, I would like to merge this pull request then rebase and merge #644. If you do not think these two functions should be removed, please say so. :)

/cc @futpib, @masaeedu


index.js
```diff
@@ -4275,7 +4229,9 @@
   //. Just ([1, 2, 3])
   //. ```
   function parseJson(pred) {
-    return B (filter (pred)) (B (eitherToMaybe) (encase (JSON.parse)));
+    return B (filter (pred))
+             (B (either (K (Nothing)) (Just))
+                (encase (JSON.parse)));
```
davidchambers: I find the new implementation just as clear as the existing one. `either (K (Nothing)) (Just)` *is* the natural transformation from either to maybe. :)

masaeedu:
Things like this can probably go into a hypothetical sanctuary lens library, which would make it easier to convert back and forth and fill in cells in the "conversion matrix" via ad-hoc composition.

PR #665, opened 2020-01-17
sanctuary-show@2.0.0

davidchambers:
<https://github.com/sanctuary-js/sanctuary-show/releases/tag/v2.0.0>

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "sanctuary-either": "1.2.0",
     "sanctuary-maybe": "1.2.0",
     "sanctuary-pair": "1.2.0",
-    "sanctuary-show": "1.0.0",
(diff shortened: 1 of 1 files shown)


Avaq:
You've marked this as a `minor` change, despite it being a major change in sanctuary-show itself. Is that because the function is not directly exposed to users and only used for representations?

davidchambers:
Strictly speaking this *is* a breaking change. Were one to be strict with versioning in a JavaScript setting, so many changes would be “breaking” that _breaking change_ would lose its meaning (there would be no minor or patch releases in the extreme case).

In practice, it is unlikely that anyone will need to update their code in response to this change.

Whether this change necessitates incrementing the major version number is moot, as I intend the next release to be v3.0.0. :)

PR #667, opened 2020-01-19
sanctuary-type-classes@12.1.0

### LINKED ISSUE #360: Add Intercalate
So, we have ```joinWith``` that joins the strings of the second argument separated by the first argument. Which only work on array of string. What if i need to work with different container? Can we abstract this? Yes, by constraints it with Foldable and Monoid.

The type signature would be:
```
intercalate :: (Foldable f, Monoid m) => m -> f m -> m
```

davidchambers:
Closes #360; closes #569

<https://github.com/sanctuary-js/sanctuary-type-classes/releases/tag/v12.1.0>

/cc @masaeedu, @syaiful6

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -56,6 +56,7 @@
       "files": ["index.js"],
       "globals": {"process": false},
       "rules": {
+        "max-len": ["error", {"code": 79, "ignoreUrls": true, "ignorePattern": "^ *//(# |  .* :: |[.] > |[.] // |[.] \\[.*\\]: )"}],
         "multiline-comment-style": ["off"]
       }
     }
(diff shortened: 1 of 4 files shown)


.eslintrc.json
```diff
@@ -56,6 +56,7 @@
       "files": ["index.js"],
       "globals": {"process": false},
       "rules": {
+        "max-len": ["error", {"code": 79, "ignoreUrls": true, "ignorePattern": "^ *//(# |  .* :: |[.] > |[.] // |[.] \\[.*\\]: )"}],
```
davidchambers: I modified the pattern to permit the following line:

> ```javascript
> //. [`Z.intercalate`]:          v:sanctuary-js/sanctuary-type-classes#intercalate
> ```

PR #679, opened 2020-02-27
Docs for installing through jsdelivr.

sugoidesune:


davidchambers:
Thank you for submitting this pull request, @sugoidesune! :clap:

index.js
```diff
@@ -333,6 +333,17 @@
 //. <script src="vendor/sanctuary-bundle.js"></script>
 //. ```
 //.
+//. The bundle can also be included via:
+//. ```html
+//. <script src="https://sanctuary.js.org/vendor/sanctuary-bundle.js"></script>
+//. ```
```
davidchambers: I suggest not advertising this option, as we should not encourage people to depend on something mutable.

index.js
```diff
@@ -333,6 +333,17 @@
...
+//.
+//. Installation of the bundle through jsdelivr:
+//. ```html
+//. <script src="https://cdn.jsdelivr.net/gh/sanctuary-js/sanctuary@3.0.0/dist/bundle.js"></script>
```
davidchambers: Perhaps we should use `X.Y.Z` as a placeholder, as we do above:

```suggestion
//. <script src="https://cdn.jsdelivr.net/gh/sanctuary-js/sanctuary@X.Y.Z/dist/bundle.js"></script>
```

In fact, this approach could supersede the `curl` approach.

index.js
```diff
@@ -341,6 +352,17 @@
...
+//. ```html
+//. <script src="https://cdn.jsdelivr.net/npm/sanctuary-maybe@2.1.0/index.js"></script>
+//. ```
+//.
+//. And accessed and aliased like so:
+//. ```javascript
+//. const Maybe = window.sanctuaryMaybe;
+//. // ...
+//. ```
+//.
```
davidchambers: Might this be more at home in the [sanctuary-maybe][1] readme?


[1]: https://github.com/sanctuary-js/sanctuary-maybe

sugoidesune: From a complete beginner point of view, I would like to keep at least one example of using an individual one. Especially since this is the only website for Sanctuary.
Personally when im new to a framework I like to see the quick start option, but also seeing that using individual modules for production is an option. Also it illustrates that you can use the types without the framework.

I think this would benefit adoption and beginners.

sugoidesune:
I dont know where the regex is for the js variables, that throw that ci error.

davidchambers:
You need to update [`no-unused-vars`][1]. :)


[1]: https://github.com/sanctuary-js/sanctuary/blob/v3.0.0/.eslintrc.json#L52

davidchambers:
Superseded by #688

PR #682, opened 2020-07-18
fantasy-land: add S.reduce_

davidchambers:
I've wanted this function many times during the past few years. @sanctuary-js/owners, are you in favour of this addition?

### DIFF
--- a/test/reduce.js
+++ b/test/reduce.js
@@ -7,7 +7,7 @@ const eq = require ('./internal/eq');
 
 test ('reduce', () => {
 
-  eq (S.show (S.reduce)) ('reduce :: Foldable f => (a -> b -> a) -> a -> f b -> a');
+  eq (S.show (S.reduce)) ('reduce :: Foldable f => (b -> a -> b) -> b -> f a -> b');
 
   eq (S.reduce (S.concat) ('x') ([])) ('x');
   eq (S.reduce (S.concat) ('x') (['A', 'B', 'C'])) ('xABC');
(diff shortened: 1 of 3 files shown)


test/reduce.js
```diff
@@ -7,7 +7,7 @@ const eq = require ('./internal/eq');
 
 test ('reduce', () => {
 
-  eq (S.show (S.reduce)) ('reduce :: Foldable f => (a -> b -> a) -> a -> f b -> a');
+  eq (S.show (S.reduce)) ('reduce :: Foldable f => (b -> a -> b) -> b -> f a -> b');
```
davidchambers, line 10: I updated this signature to match the signature used in the documentation.

Avaq:
It makes sense for reduce to be like that, I often have my reducer before I have the initial value, and I rarely (never?) have my initial value before I have a reducer. When I define reduce, I always make the arguments `f x xs`. Have you considered simply updating `reduce` instead of adding a variant?

Avaq:
Deleted my previous comment, as I had misunderstood the change.

gabejohnson:
My only concern is that this function would share the _signature_ of a right fold, but not the behavior. There's nothing _wrong_ about that (functions with different behaviors frequently share the same type signature), just something to consider.

davidchambers:
> My only concern is that this function would share the _signature_ of a right fold, but not the _behavior_.

Interesting observation, Gabe. I admit that I am not completely comfortable with `foldl` and `foldr` in Haskell. It is not clear to me *why* these functions are of different types:

```haskell
> :type foldl
foldl :: Foldable t => (b -> a -> b) -> b -> t a -> b

> :type foldr
foldr :: Foldable t => (a -> b -> b) -> b -> t a -> b

> foldl (\s c -> c:s) "." "abc"
"cba."

> foldr (\c s -> c:s) "." "abc"
"abc."
```

Thank you for considering this addition, folks. I may be the only person *excited* about this function, but as no one has voiced opposition, I will merge this pull request. :)

PR #683, opened 2020-07-26
fromLeft and fromRight

ryanbas21:
Forgot why I never opened this, but I can't see a reason to not push it forward. Let me know if there is anything you'd like me to add / change etc.

Summary of changes: Add functionality of `fromLeft` and `fromRight`

`fromLeft :: a -> Either a b -> a`
`fromRight :: b -> Either a b -> b`

### DIFF
--- a/index.js
+++ b/index.js
@@ -2079,7 +2079,7 @@
   //. a Just, the return value is the result of applying the function to
   //. the Just's value. Otherwise, the first argument is returned.
   //.
-  //. See also [`maybe_`](#maybe_).
+  //. See also [`maybe_`](#maybe_) and [`fromMaybe`](#fromMaybe).
   //.
   //. ```javascript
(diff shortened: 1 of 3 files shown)


davidchambers:
This pull request contains a subset of the changes proposed in #644. This pull request does not change the behaviour of `fromEither`, meaning that these changes could be included in the next minor release. This would result in two equivalent functions coexisting, which would be unusual given our avoidance of aliases. It seems like a reasonable course of action.

davidchambers:
@ryanbas21, as you may or may not be aware, #644 was open at the time you made your additions. That pull request has already been reviewed and discussed, and includes documentation. It makes sense to reuse @futpib's work rather than reproduce it, so I cherry-picked one of the commits and reverted the changes to `fromEither` included in that commit.

@futpib, I hope you are comfortable with me modifying one of your commits.

@ryanbas21, I hope you are comfortable with me commandeering your pull request.

PR #685, opened 2020-07-26
regexp: add S.replace and S.replace_

davidchambers:
This pull request adds the following two functions:

```haskell
replace  :: RegExp ->                          String  -> String -> String
replace_ :: RegExp -> (Array (Maybe String) -> String) -> String -> String
```

Design decisions:

  - Both functions take `pattern :: RegExp`, so there is no straightforward equivalent of `s1.replace (s2, s3)` for strings `s1`, `s2`, and `s3`. One *could* use `S.replace (S.regex ('') (S.regexEscape (s2))) (s3) (s1)`, but it would be simpler and clearer to use [`String#replace`][] directly in such cases.

    Using `pattern :: RegExp` exclusively provides two benefits:

      - __Simplicity.__ Supporting both `RegExp` and `String` would require splitting `replace` into two functions (which would pose a naming challenge) or using `Either RegExp String` (which would not be ergonomic).

      - __Clarity.__ It's unclear whether `R.replace ('o') ('x') ('foo')` should evaluate to `'fxo'` or to `'fxx'`, whereas the presence of the `g` flag in `/o/g` indicates that all occurrences of `o` should be replaced, and the absence of the `g` flag in `/o/` indicates that only the first occurrence of `o` should be replaced.

  - Like [`replace'`][], `S.replace_` ignores the “`offset`” and “`string`” arguments provided by [`String#replace`][] (when given a function). These are rarely useful.

  - Unlike [`replace'`][], `S.replace_` ignores the “`groups`” argument provided by [`String#replace`][] (when given a function) if the pattern contains any named capturing groups. I included a fix for this deficiency in [purescript/purescript-strings#126][].

  - Unlike [`replace'`][], `S.replace_` ignores the “`match`” argument provided by [`String#replace`][] (when given a function). This simplifies the function's type, and saves one from using `S.K (...)` when only the captured groups are important (as is commonly the case). If one *does* require access to the matched substring as a whole, one can use `/(...)/` to capture it. Admittedly, this approach is impractical if the RegExp object is defined in another module, but in such exceptional cases one can of course use [`String#replace`][] directly.

  - Unlike [`replace'`][], `S.replace_` acknowledges the existence of optional capturing groups:

    ```javascript
    > S.replace_ (S.regex ('') ('(Pure)?(Script)')) (S.show) ('JavaScript')
    'Java[Nothing, Just ("Script")]'
    ```

    Evaluating the equivalent PureScript expression results in an exception being thrown:

    ```purescript
    > replace' (unsafeRegex "(Pure)?(Script)" noFlags) (const show) "JavaScript"
    ! TypeError: undefined is not an object (evaluating 's.length')
    ```

    I have proposed the use of `Array (Maybe String)` in place of `Array String` in [purescript/purescript-strings#126][].

I have opened sanctuary-js/sanctuary-site#87 to provide several examples of `replace` and `replace_` in use.

@sanctuary-js/owners, what do you think of the proposed additions to the library?

---

@CrossEye, you may like to rethink [`R.replace`][], which currently exposes all the quirks of [`String#replace`][].

[`R.replace`]:                          https://ramdajs.com/docs/#replace
[`String#replace`]:                     https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace
[`replace`]:                            https://pursuit.purescript.org/packages/purescript-strings/4.0.1/docs/Data.String.Regex#v:replace
[`replace'`]:                           https://pursuit.purescript.org/packages/purescript-strings/4.0.1/docs/Data.String.Regex#v:replace'
[purescript/purescript-strings#126]:    https://github.com/purescript/purescript-strings/pull/126

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ jobs:
               nvm exec $1 npm test
             }
             case $CIRCLE_NODE_INDEX in
-              0) test_with_version 8 ;;
-              1) test_with_version 10 ;;
-              2) npm install && npm test ;;
+              0) test_with_version 10 ;;
+              1) npm install && npm test ;;
+              2) test_with_version 14 ;;
             esac
--- a/index.js
+++ b/index.js
@@ -452,6 +452,17 @@
     };
   }
 
+  //  invoke2 :: String -> a -> b -> c -> d
+  function invoke2(name) {
+    return function(x) {
+      return function(y) {
+        return function(target) {
+          return target[name] (x, y);
+        };
+      };
+    };
+  }
+
   //  toObject :: a -> Object
   function toObject(x) {
     return x == null ? Object.create (null) : Object (x);
@@ -1731,12 +1742,12 @@
   //. Curries the given ternary function.
   //.
   //. ```javascript
-  //. > const replaceString = S.curry3 ((what, replacement, string) =>
-  //. .   string.replace (what, replacement)
-  //. . )
+  //. > const defineProperty = S.curry3 (Object.defineProperty)
   //.
-  //. > replaceString ('banana') ('orange') ('banana icecream')
-  //. 'orange icecream'
+  //. > const o = defineProperty ({}) ('x') ({value: 42, writable: false})
+  //.
+  //. > o.x
+  //. 42
   //. ```
   function curry3(f) {
     return function(x) {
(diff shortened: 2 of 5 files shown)


.circleci/config.yml
```diff
@@ -28,7 +28,7 @@ jobs:
               nvm exec $1 npm test
             }
             case $CIRCLE_NODE_INDEX in
-              0) test_with_version 8 ;;
-              1) test_with_version 10 ;;
-              2) npm install && npm test ;;
+              0) test_with_version 10 ;;
+              1) npm install && npm test ;;
+              2) test_with_version 14 ;;
```
davidchambers, line 33: In order to use named capturing groups (in the test suite) I had to drop `8` from the set of Node.js versions.

index.js
```diff
@@ -1731,12 +1742,12 @@
   //. Curries the given ternary function.
   //.
   //. ```javascript
-  //. > const replaceString = S.curry3 ((what, replacement, string) =>
-  //. .   string.replace (what, replacement)
-  //. . )
+  //. > const defineProperty = S.curry3 (Object.defineProperty)
```
davidchambers, line 1745: It is not easy to find a built-in function that takes three arguments!

Avaq:
Regarding naming, maybe we can use

```hs
replace     :: RegExp -> (Array (Maybe String) -> String) -> String -> String
replaceWith :: RegExp ->                          String  -> String -> String
```

index.js
Avaq: You could use a variadic function, such as `Math.max`, right? Then you could create curried built-ins for any number of arguments.

Avaq:
> because `'$1-$2-$3'`

I did not realise that in the replacement string `$n` had special meaning.

davidchambers:
`replaceWith` would be an *excellent* name for a `RegExp -> String -> String -> String` function without support for [special replacement patterns][1]. :thinking:


[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_string_as_a_parameter

index.js
davidchambers: That's true. Binary versions of variadic functions are compelling, but the use cases for ternary versions are less apparent. I am happy to have found `Object.defineProperty` for this example. :)

Avaq:
I like the idea of a function `RegExp -> String -> String -> String`, that doesn't support special replacement patterns. It seems more sane to me, and less prone to creation of some weird attack vector where a user can supply `$0` as input to gain access the an extracted matching group. I think the usage of matching groups could be left solely for `RegExp -> (Array (Maybe String) -> String) -> String -> String`.

davidchambers:
Closing in favour of #686

Issue #687, opened 2020-08-04
ES5 => ES6


davidchambers:
Valued members of the Sanctuary community, I request your input.

I remain pleased with our decision not to require any build tools to publish this project. Our approach is simply to wait for a language feature to become universally supported before adopting it.

Sanctuary currently targets ES5. I :heart: one feature added to the language in ES6: `x => x`. I look forward to using arrow function expressions in the Sanctuary codebase, but only once this imposes no burden on those who depend on Sanctuary.

__Has the time come for Sanctuary to adopt arrow function expressions?__ Help us to decide by sharing your reaction!

  - :rocket: This would not be a problem for me as I already target ES6 (or later).
  - :thumbsup: ES5 compatibility is important to me but I am happy to transpile dependencies to ES5.
  - :thumbsdown: ES5 compatibility is important to me and I want to keep my deployment process simple.

davidchambers:
This took a little longer than expected. :see_no_evil:

PR #692, opened 2020-08-10
function: remove S.curry2, S.curry3, S.curry4, and S.curry5

davidchambers:
Sanctuary's `curry*` functions were added in late 2016 (#289). In the intervening years, arrow function expressions have become widely supported (#687). Arrow function expressions make manual currying painless, so these functions no longer provide significant value.

The primary motive for removing these functions is to reduce the library's surface area. The other reason is that currying a “foreign” function is rarely the only change one should make. Other possible changes include:

  - changing argument order;
  - transforming inputs (e.g. converting `Maybe String` to `Nullable String`);
  - transforming outputs (e.g. converting `Promise User` to `Future Error User`); and
  - catching exceptions.

Sanctuary's `curry*` functions discourage consideration of such potential API improvements.

/cc @laduke


index.js
```diff
@@ -424,6 +424,39 @@
...
+    return function(w) {
+      return function(x) {
+        return function(y) {
+          return function(z) {
+            return f (w, x, y, z);
+          };
+        };
+      };
+    };
+  }
```
Bradcomp, line 458: Are these functions used within the codebase? Why are we keeping them around?

davidchambers: These functions are used to curry `Z` functions. `curry4` is applied exactly once: `curry4 (Z.lift3)`.

PR #693, opened 2020-08-12
regexp: add S.replace

davidchambers:
Supersedes #686

This pull request adds the following function:

```haskell
replace :: (Array (Maybe String) -> String) -> RegExp -> String -> String
```

In <https://github.com/sanctuary-js/sanctuary/pull/686#discussion_r466296798>, @Avaq made a strong case for naming this more general function `replace`. @Avaq also noted that the specialized function, `replaceWith`, is trivially derivable from `replace`. For this reason I decided not to include `replaceWith` in this pull request; it could easily be added in the future if `replace (K ('...'))` proves irksome.

I have opened sanctuary-js/sanctuary-site#90 to provide several examples of `replace` in use.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4403,6 +4403,53 @@
     impl: matchAll
   };
 
+  //# replace :: (Array (Maybe String) -> String) -> RegExp -> String -> String
+  //.
+  //. Replaces occurrences of the given pattern within the given string
+  //. in accordance with the given replacement function, which receives an
+  //. array of captured values. Replaces all occurrences of the pattern if
+  //. its `g` flag is set; just the first occurrence otherwise.
+  //.
+  //. ```javascript
+  //. > S.replace (([$1]) => S.maybe ('') (S.toUpper) ($1)) (/(\w)/) ('foo')
+  //. 'Foo'
+  //.
+  //. > S.replace (([$1]) => S.maybe ('') (S.toUpper) ($1)) (/(\w)/g) ('foo')
+  //. 'FOO'
+  //.
+  //. > S.replace (S.show) (/(foo)(bar)?/) ('<>')
+  //. '<>'
+  //.
+  //. > S.replace (S.show) (/(foo)(bar)?/) ('<foo>')
+  //. '<[Just ("foo"), Nothing]>'
+  //.
+  //. > S.replace (S.show) (/(foo)(bar)?/) ('<foobar>')
+  //. '<[Just ("foo"), Just ("bar")]>'
+  //. ```
+  function replace(substitute) {
+    return function(pattern) {
+      return function(text) {
+        return text.replace (pattern, function() {
+          var groups = [];
+          var group, idx = 1;
+          //  eslint-disable-next-line no-plusplus
+          while (typeof (group = arguments[idx++]) !== 'number') {
+            groups.push (group == null ? Nothing : Just (group));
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -4413,6 +4413,53 @@
...
+  function replace(substitute) {
+    return function(pattern) {
+      return function(text) {
+        return text.replace (pattern, function() {
+          var groups = [];
+          var group, idx = 1;
+          //  eslint-disable-next-line no-plusplus
+          while (typeof (group = arguments[idx++]) !== 'number') {
+            groups.push (group == null ? Nothing : Just (group));
+          }
```
Avaq, line 4448: Why do you normally follow the no-plusplus rule, and how is this an exception?

index.js
```diff
@@ -4413,6 +4413,53 @@
...
+  //. '<[Just ("foo"), Nothing]>'
+  //.
+  //. > S.replace (S.show) (/(foo)(bar)?/) ('<foobar>')
+  //. '<[Just ("foo"), Just ("bar")]>'
+  //. ```
+  function replace(substitute) {
+    return function(pattern) {
+      return function(text) {
+        return text.replace (pattern, function() {
+          var groups = [];
```
Avaq, line 4443: Is it possible to know the length of this array upfront? Say, using `var groups = new Array(arguments.length)`. If so, then that would get you slightly better performance at relatively low implementation and maintenance cost. You would have to alter `groups.push` statements to index assignment statements instead, eg: `groups[idx] = (...)`.

index.js
davidchambers: I eschew the prefix increment operator (`++idx`) because it is a special case of `+=`, and I dislike special cases:

```console
$ python -c 'import this' | grep 'only one'
There should be one-- and preferably only one --obvious way to do it.
```

I use the postfix increment operator (`idx++`) only when it enables a clearer algorithm than is possible with `+=`. This is the case here. Admittedly, we could use `idx += 1` and change the initial value of `idx` from `1` to `0`, but it would then be less clear that the first element is skipped (and observant readers might wonder whether the first element is being skipped unintentionally).

index.js
davidchambers: > Is it possible to know the length of this array upfront?

No, due to the existence of named capturing groups:

```javascript
> '@sanctuary-js'.replace (/@([-\w]+)/, function() { return arguments.length; })
'4'

> '@sanctuary-js'.replace (/@(?<username>[-\w]+)/, function() { return arguments.length; })
'5'
```

Avaq: Ah, so these named groups are skipped? Is that what the `typeof` check below is doing?

davidchambers: That's right. I imagine `groups` is omitted when the pattern contains no named capturing groups to preserve backward compatibility. Providing the sensible empty value, `{}`, as an additional argument would break programs that assume that `Array.prototype.slice.call(arguments, 1, -2)` extracts (only) the captured values.

Avaq: Oh, wow. It took me a really long time to realize that what's happening there is: *skip one argument, and keep taking arguments until a number is encountered*. The numeric argument signifies the "end" of the capture group arguments. Wow. The signature of JavaScript's `replace` is crazier than I remember.

Avaq (approved):

PR #696, opened 2020-08-23
number: remove S.mean

davidchambers:
This is a useful function, but it feels too specialized to remain in the library. Do others agree?

/cc @jdudek


Avaq (approved):
I agree.

PR #697, opened 2020-08-23
array: add S.findMaybe

davidchambers:
With [`find`][1] one sometimes applies a transformation to determine whether an element is suitable only to apply the transformation a second time if a suitable element is found:

```javascript
> S.chain (S.parseInt (16))
.         (S.find (S.compose (S.isJust) (S.parseInt (16)))
.                 (['A', 'B', 'C']))
Just (10)
```

With `findMaybe` the transformation is provided just once:

```javascript
> S.findMaybe (S.parseInt (16)) (['A', 'B', 'C'])
Just (10)
```

As is often the case, the function's type is obvious but its name is not. What do you think of the name `findMaybe`? It follows the precedent set by [`map`][2] and [`mapMaybe`][3], although the correspondence is imprecise.

[1]: https://sanctuary.js.org/#find
[2]: https://sanctuary.js.org/#map
[3]: https://sanctuary.js.org/#mapMaybe

### DIFF
--- a/index.js
+++ b/index.js
@@ -3204,7 +3204,7 @@
   //. of the structure that satisfies the predicate; Nothing if there is no
   //. such element.
   //.
-  //. See also [`elem`](#elem).
+  //. See also [`findMaybe`](#findMaybe) and [`elem`](#elem).
   //.
   //. ```javascript
   //. > S.find (S.lt (0)) ([1, -2, 3, -4, 5])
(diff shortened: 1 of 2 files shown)


Avaq (approved):

davidchambers:
Thanks for the review, Aldwin. I will leave this pull request open for a few days to give others the opportunity to comment.

gabejohnson:
In PureScript, this function is called [findMap](https://pursuit.purescript.org/packages/purescript-foldable-traversable/4.1.1/docs/Data.Foldable#v:findMap).

There's another (more general) function [oneOfMap](https://pursuit.purescript.org/packages/purescript-foldable-traversable/4.1.1/docs/Data.Foldable#v:oneOfMap) which uses some `Plus f` instead of `Maybe`. That would probably be less useful to the average user though.

davidchambers:
Thanks, Gabe! `findMap` is a good name. Better than `findMaybe`, do you think? I am not attached to `findMaybe`.

gabejohnson:
I like `findMap` because it suggests that you're likely modifying the value you find. I don't have any real attachment though, just thought it useful to point toward some prior art.

davidchambers:
Superseded by #698

PR #699, opened 2020-08-29
array: add S.elem_

davidchambers:
I find myself flipping [`elem`][1] quite often.

Do you agree with this addition?

[1]: https://sanctuary.js.org/#elem


dicearr:
I also end up defining  `const within = S.flip (S.elem);` in almost every project. However, I wouldn't bother implementing it in Sanctuary as it can be easily defined in user-land. I think that the other `_` ending functions like `S.reduce_` are slightly more complicated for the user to implement.

PR #707, opened 2021-01-28
Rollup and Sanctuary

edlimerkaj:


index.js
```diff
@@ -334,6 +334,10 @@
 //. const $ = window.sanctuaryDef;
 //. // ...
 //. ```
+//. If you are using rollup in your project, where you have integrated Sanctuary, you might
+//. get errors. Please import builtins and globals from rollup-plugin-node-builtins and 
+//. rollup-plugin-node-globals respectively, in the rollup config file. These are needed to add 
+//. the utils folder, which is referenced in the sanctuary code.
```
davidchambers, line 340: This is great! Could you share the snippet from your Rollup config file? Perhaps we could include that here.

edlimerkaj: Yes, sure. Here is my rollup.config file for a simple project, where I've used Sanctuary and rollup:

import commonjs from 'rollup-plugin-commonjs';
import nodeResolve from 'rollup-plugin-node-resolve';
import builtins from 'rollup-plugin-node-builtins';
import globals from 'rollup-plugin-node-globals';

export default {
    input: './src/main.js',
    output: {
        file: './build/bundle.min.js',
        format: 'iife',
        name: 'bundle'
    },
    plugins: [
    nodeResolve({ preferBuiltins: false }), // or `true`
    commonjs(),
    globals(),
    builtins()]
}
;

davidchambers:
I'm sorry for dropping the ball, @edlimerkaj. Do you know whether the incompatibility still exists, more than two years later?

PR #712, opened 2021-05-16
circle: create test matrix without nvm

davidchambers:
This took a lot of trial and error, but the effort will be amortized as we apply this change to the other Sanctuary projects. :)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,34 +1,29 @@
 version: 2.1
 
+orbs:
+  node: circleci/node@4.3.0
+
 jobs:
-  build:
-    docker:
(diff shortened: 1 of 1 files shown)


svozza:
This looks a lot cleaner. You might also want to consider removing Node 10.x now that it is officially EOL.

Bradcomp (approved):

CrossEye:
That's very clean!

davidchambers:
> You might also want to consider removing Node 10.x now that it is officially EOL.

I consider this matrix a way to determine and document what we *do* support, rather than a promise of any kind. :)

Issue #645, opened 2019-08-02
Consider flipping the argument order of ap


Avaq:
I would argue that the current argument order does not favour composition with partial application. I would want the following two programs to be equivalent.

```js
S.lift3 (f) (ma) (mb) (mc)
```

```js
S.pipe ([
  S.map (f),
  S.ap (mb),
  S.ap (mc),
]) (ma)
```

However, they are not, because `S.ap` takes the function argument first, rather than last, necessitating a `flip` before `ap` is useful:

```js
S.pipe ([
  S.map (f),
  S.flip (S.ap) (mb),
  S.flip (S.ap) (mc),
]) (ma)
```

Furthermore, the observation that `ap` is the only Fantasy Land function in Sanctuary that does not follow the convention that its last argument is the `this` to the Fantasy Land method reinforces my view that the arguments should be flipped. This has also led me [to flip the arguments in the latest version of Fluture](https://gist.github.com/Avaq/ee2c6c819db4c37258e9a226e6380a38#arguments-to-some-functions-were-flipped).

Avaq:
I realized later that from the perspective of combinatory logic, flipping the current `ap` is a bad idea. It would no longer be equal to the [common `S` combinator](https://gist.github.com/Avaq/1f0636ec5c8d6aed2e45#file-combinators-md).

davidchambers:
On the subject of combinators, I now believe Sanctuary should provide all the common combinators, even though several are specializations of functions Sanctuary already provides.

If you are interested in pursuing this change, Aldwin, please create a branch named `avaq/ap`, flip the argument order, update the corresponding test cases, and finally create experimental branches for at least two projects that depend on Sanctuary (pointing to `avaq/ap`). I realize that this would be a significant amount of work, but we should not change a crucial function unless we are convinced that a change is warranted.

davidchambers:
Are you still in favour of this change, Aldwin?

Avaq:
The arguments in favour of and those against cancel each other out for me. I prefer to leave things as is, but only by a small margin.

davidchambers:
Consider it considered!

Avaq:
Actually why did we never consider just adding `ap_` or `flap` (:grin:) as I call it in my code-bases?

What makes `flap` a great name is that its primary purpose is to use it when `lift5` is not big enough, so in my code, it only ever appears in pipelines with 5 or more consecutive flaps.

davidchambers:
[...]

PR #713, opened 2021-07-02
Add ap_: A flipped variant of ap

Avaq:
See #645


Avaq:
I don't intend to finish this, as I have had no more need for it

PR #716, opened 2021-08-16
[WIP] add support for TypedArrays/arraBuffer/nodejs' Buffer

dotnetCarpenter:
Currently failing to implement tests for TypedArray.

In `assert.strictEqual (equals (actual) (expected), true);` the
`Setoid.test` test fails unconditionaly for a TypedArray (`Uint8Array`).

Need help ;)

Reproduce:

```js
const S = require ('./internal/sanctuary');
const eq = require ('./internal/eq');

eq (S.unchecked.drop (1) (new Uint8Array([10, 20, 30, 40, 50]))) (S.unchecked.Just (new Uint8Array([20, 30, 40, 50])));
```

PS. For now my goal is to support various Buffers for `unchecked` functions, since it seems like this should be opt-in and I don't know how to implement support in https://github.com/sanctuary-js/sanctuary-def yet.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2834,8 +2834,8 @@
       return function(xs) {
         if (n < 0) return Nothing;
 
-        //  Fast path for arrays.
-        if (Array.isArray (xs)) {
+        //  Fast path for objects supporting `slice`.
+        if (xs.slice) { // we want to support TypedArrays e.i. buffers
           return n <= xs.length ? Just (arrayCase (n, xs)) : Nothing;
         }
 
--- a/test/drop.js
+++ b/test/drop.js
@@ -28,4 +28,7 @@ test ('drop', () => {
 
   eq (S.drop (-1) (Cons (1) (Cons (2) (Cons (3) (Nil))))) (S.Nothing);
 
+  // eq (S.unchecked.drop (0) (new Uint8Array([10, 20, 30, 40, 50]))) (S.unchecked.Just (new Uint8Array([10, 20, 30, 40, 50])));
+  eq (S.unchecked.drop (1) (new Uint8Array([10, 20, 30, 40, 50]))) (S.unchecked.Just (new Uint8Array([20, 30, 40, 50])));
+
 });


davidchambers:
I don't think this makes sense.

Here is the type of `take`:

```haskell
take :: (Applicative f, Foldable f, Monoid (f a)) => Integer -> f a -> Maybe (f a)
```

We could specialize this by replacing `f` with `Array` and `a` with `String`:

```haskell
take :: Integer -> Array String -> Maybe (Array String)
```

Doing the same with `Uint8Array` results in a nonsensical type:

```haskell
take :: Integer -> Uint8Array String -> Maybe (Uint8Array String)
```

`Uint8Array` is a type rather than a type constructor: these arrays only ever contain 8-bit unsigned integers.

One could define a specialized function:

```haskell
takeUint8 :: Integer -> Uint8Array -> Maybe Uint8Array
```

dotnetCarpenter:
@davidchambers How would I go about implementing specialised functions for Sanctuary?

Would `TypedArray` encompass all of [Int8Array, Uint8Array, Uint8ClampedArray, Int16Array, Uint16Array, Int32Array, Uint32Array, Float32Array, Float64Array, BigInt64Array and BigUint64Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray#typedarray_objects)?

```js
takeTypedArray :: Integer -> TypedArray-> Maybe TypedArray
```

davidchambers:
The hypothetical `TypedArray` type is problematic because it would allow `takeTypedArray` to take a `Uint8Array` but return a `Maybe Float64Array`.

The [simplest][1] solution would be to define a specialized function for each type.


[1]: https://www.infoq.com/presentations/Simple-Made-Easy/

dotnetCarpenter:
So that would be 11 specialized functions, just for `take`?

```haskell
takeInt8Array :: Integer -> Int8Array -> Maybe Int8Array
```

```haskell
takeUint8Array :: NonNegativeInteger -> Uint8Array -> Maybe Uint8Array
```

```haskell
takeUint8ClampedArray :: NonNegativeInteger -> Uint8ClampedArray -> Maybe Uint8ClampedArray
```

```haskell
takeInt16Array :: Integer -> Int16Array -> Maybe Int16Array
```

```haskell
takeUint16Array :: NonNegativeInteger -> Uint16Array -> Maybe Uint16Array
```

```haskell
takeInt32Array :: Integer -> Int32Array -> Maybe Int32Array
```

```haskell
takeUint32Array :: NonNegativeInteger -> Uint32Array -> Maybe Uint32Array
```

```haskell
takeFloat32Array :: ValidNumber -> Float32Array -> Maybe Float32Array
```

```haskell
takeFloat64Array :: ValidNumber -> Float64Array -> Maybe Float64Array
```

Does Sanctuary support BigInt?

```haskell
takeBigInt64Array :: BigInt -> BigInt64Array -> Maybe BigInt64Array
```

```haskell
takeBigUint64Array :: NonNegativeBigInt -> BigUint64Array -> Maybe BigUint64Array
```

But actually, before going down this rabbit hole. I'm much more interested in support for [Nodejs' Buffer](https://nodejs.org/api/buffer.html) since I deal with a database that returns Buffers.

```haskell
takeBuffer :: NonNegativeInteger -> Buffer -> Maybe Buffer

dropBuffer :: NonNegativeInteger -> Buffer -> Maybe Buffer

takeLastBuffer :: NonNegativeInteger -> Buffer -> Maybe Buffer

dropLastBuffer :: NonNegativeInteger -> Buffer -> Maybe Buffer

takeWhileBuffer :: (a -> Boolean) -> Buffer -> Buffer

dropWhileBuffer :: (a -> Boolean) -> Buffer -> Buffer
```

Ref: [NonNegativeInteger](https://github.com/sanctuary-js/sanctuary-def#nonnegativeinteger--type)

dotnetCarpenter:
I tried to look through some code a few months back to find context for this. I did not find anything.

But as I recall, I got a Buffer from the database and wanted to use  `S.unchecked.drop` but the [`Array,isArray` check in `_takeDrop` would not let me](https://github.com/sanctuary-js/sanctuary/pull/716/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346L2838), so I used `.slice` instead without a `Maybe`.

Changing that one line, made it all work in Sanctuary but .... here I am.

Avaq:
If I'd want to `take` from a Buffer, I'd probably do something like:

```js
// firstFiveBytes :: Buffer -> Maybe Buffer
const firstFiveBytes = S.pipe ([
  Array.from,
  S.take (5),
  S.map (Buffer.from),
])
```

dotnetCarpenter:
@Avaq I'm imagining that working with Buffers without converting them to Array would be faster overall. But I should write some code to compare with.

There is only one line that need to change in Sanctuary in order to support Buffers, so it should be fairly easy to compare.

But it won't be today :)

PR #732, opened 2023-12-08
use ES6 syntax

davidchambers:
#687

### DIFF
--- a/index.js
+++ b/index.js
@@ -337,7 +337,7 @@
 //.
 //. ## API
 
-(function(f) {
+(f => {
 
   'use strict';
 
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -369,158 +369,98 @@
...
+    ({Nil, Cons} = List);
+    Sum = __doctest.require ('./test/internal/Sum');
+    S = Object.assign (
+      create ({
+        checkTypes: true,
+        env: env.concat ([List.Type ($.Unknown), Sum.Type]),
+      }),
+      {env}  // see S.env doctest
+    );
+    [Descending, Nil, Cons, Sum, S];
```
davidchambers, line 407: These identifiers would otherwise only be referenced in doctests. Referencing them here avoids lint errors.

index.js
```diff
@@ -3884,13 +3629,10 @@
...
   //. ```
-  function values(strMap) {
-    return Z.map (function(k) { return strMap[k]; }, Object.keys (strMap));
-  }
   _.values = {
     consts: {},
     types: [$.StrMap (a), $.Array (a)],
-    impl: values
+    impl: Object.values,
   };
```
davidchambers, line 3636: I simplified this implementation. ;)

index.js
Avaq: Maybe warrants a comment

Avaq (approved):

PR #739, opened 2023-12-20
remove test/internal/eq and test/internal/throws modules

davidchambers:
This pull request replaces all uses of our internal `eq` and `throws` functions with uses of [`assert.deepStrictEqual`][1] and [`assert.throws`][1] respectively.

[1]: https://nodejs.org/api/assert.html#assertdeepstrictequalactual-expected-message
[2]: https://nodejs.org/api/assert.html#assertthrowsfn-error-message


test/get.js
```diff
@@ -1,29 +1,31 @@
...
 
-  eq (S.get (S.is ($.Number)) ('x') ({x: 0, y: 42})) (S.Just (0));
-  eq (S.get (S.is ($.Number)) ('y') ({x: 0, y: 42})) (S.Just (42));
-  eq (S.get (S.is ($.Number)) ('z') ({x: 0, y: 42})) (S.Nothing);
-  eq (S.get (S.is ($.String)) ('z') ({x: 0, y: 42})) (S.Nothing);
-  eq (S.get (S.is ($.String)) ('x') ({x: 0, y: 42})) (S.Nothing);
-  eq (S.get (S.is ($.RegExp)) ('x') ({x: vm.runInNewContext ('/.*/')})) (S.Just (/.*/));
+  // <https://github.com/nodejs/node/issues/44462>
+  const pattern = vm.runInNewContext ('/.*/');
+  eq (S.get (S.is ($.RegExp)) ('x') ({x: pattern}), S.Just (pattern));
```
davidchambers, line 23: This change is necessary because [`assert.deepStrictEqual`][1] checks prototype identity (nodejs/node#44462). This assertion was testing three things at once: `S.is`, `S.get`, and our internal `eq` function. We never meant to assert that equivalent `RegExp` objects from different contexts are considered equal, as this has nothing to do with `S.get`.


[1]: https://nodejs.org/api/assert.html#assertdeepstrictequalactual-expected-message

Avaq (approved):

PR #740, opened 2023-12-21
convert test modules to ES modules

davidchambers:
This pull request converts all files in the __test__ directory from CommonJS modules to ECMAScript modules.

__test/internal/List__ and __test/internal/Sum__ are used by doctests in the main module, which remains a CommonJS module, so they must also remain CommonJS modules for now. I first renamed them:

```console
$ git mv test/internal/List.{js,cjs}
$ git mv test/internal/Sum.{js,cjs}
```

I then created wrapper modules for use within the test suite:

```console
$ touch test/internal/{List,Sum}.mjs
$ git add test/internal/{List,Sum}.mjs
```

The only other non-trivial change involved using [`module.createRequire`][1] to keep the `NODE_ENV` tests working.

[1]: https://nodejs.org/api/module.html#modulecreaterequirefilename


test/NODE_ENV.js
```diff
@@ -1,17 +1,26 @@
...
+
   const source = fs.readFileSync (path.join (__dirname, '..', 'index.js'), 'utf8');
 
+  const {version: $$version} = (
+    JSON.parse (
+      fs.readFileSync (
+        path.join (__dirname, '..', 'node_modules', 'sanctuary-def', 'package.json')
+      )
+    )
+  );
```
davidchambers, line 22: In the future we should be able to switch to an import declaration:

```javascript
import pkg from 'sanctuary-def/package.json' with {type: 'json'};
```

Avaq (approved):

PR #742, opened 2023-12-21
update `is` to only work with concrete types

davidchambers:
This change is motivated by our desire to support tree shaking. `create` is the most problematic function, but `is` also depends on the [environment][1].

This pull request makes `is` independent of the environment. This means that `S.is ($.Array ($.TypeVariable ('a')))` no longer works, but I imagine no one has ever used this functionality. [`$.test`][1] is still available for advanced use cases, and is almost as convenient given that one must import sanctuary-def to create a type variable anyway.

[1]: https://github.com/sanctuary-js/sanctuary/tree/v3.1.0#env
[2]: https://github.com/sanctuary-js/sanctuary-def/tree/v0.22.0#test

### DIFF
--- a/index.js
+++ b/index.js
@@ -526,7 +526,6 @@
       S[name] = def (name) (consts) (types) (impl);
     }
     S.env = env;
-    S.is = def ('is') ({}) ([$.Type, $.Any, $.Boolean]) ($.test (env));
     S.Maybe = Maybe;
     S.Nothing = Nothing;
     S.Either = Either;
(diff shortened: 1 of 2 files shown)


Avaq:
With the change we're envision for `env`, this may not even be necessary

davidchambers:
> With the change we're envisioning for `env`, this may not even be necessary

True. Even so, I'd like to simplify this function and rely on `$.test` in the rare situations in which type variables are needed.

davidchambers:
I'm with you, Aldwin. There's no compelling reason to remove this fringe functionality given our plans for `$.env` and `S.env`.

